PR #45, opened 2020-03-09
Streamline the "upload data" page

waxlamp:
This PR attempts to make the UX around selecting data for upload more intuitive and streamlined.

The process looks like this:
1. The user first selects a file or files to upload.
2. The interface automatically selects a reasonable name for the dataset, which the user can choose to edit.
3. The user can optionally enter a description of the dataset.

There is no longer a need to first select a place to upload the data files to, eliminating the necessary two-step process that was in place before.

Depends on #44

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -10,7 +10,7 @@ services:
       context: ./devops/girder
     command: "-d mongodb://mongodb:27017/girder --host 0.0.0.0"
     ports:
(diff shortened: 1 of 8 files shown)


manthey:
Eventually (not this PR), we want to allow picking a parent folder for the new folder rather than putting everything in the user's Public folder (though that would be the default).

PR #48, opened 2020-03-23
Parse uploaded files' filenames to determine dataset metadata

waxlamp:
This might still be missing some things that were requested, like a display of the time steps, Z values, etc., found during processing, and the opportunity to correct or deny such findings.

### DIFF
--- a/src/store/GirderAPI.ts
+++ b/src/store/GirderAPI.ts
@@ -269,7 +269,6 @@ export default class GirderAPI {
     // Keep the size of the cache under control.
     if (this.imageCache.size === 100) {
       const oldest = this.imageCache.entries().next().value[0];
-      console.log(oldest);
       this.imageCache.delete(oldest);
     }
 
--- a/src/store/GirderAPI.ts
+++ b/src/store/GirderAPI.ts
@@ -21,6 +21,7 @@ import {
   mergeHistograms,
   ITileHistogram
 } from "./images";
+import { getFilenameMetadata } from "@/utils/parsing";
 
 interface HTMLImageElementLocal extends HTMLImageElement {
   _waitForHistogram?: boolean;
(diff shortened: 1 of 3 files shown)


src/store/GirderAPI.ts
```diff
@@ -518,14 +526,23 @@ function parseTiles(items: IGirderItem[], tiles: ITileMeta[]) {
...
+    if (metadata.chan !== null && !channelInt.has(metadata.chan)) {
+      channelInt.set(metadata.chan, channelInt.size);
+    }
+
     tile.frames.forEach((frame, j) => {
-      const t = +frame.TheT;
-      const xy = +(frame.IndexXY || 0);
-      const z = +(frame.IndexZ !== undefined ? frame.IndexZ : frame.PositionZ);
-      const c = +frame.TheC;
+      const t = metadata.t || +frame.TheT;
```
manthey: I think we want these to be `metadata.t !== undefined ? metadata.t || +frame.TheT` so that zero behaves properly?

src/utils/parsing.ts
```diff
@@ -0,0 +1,74 @@
+function applyRegex(regex: RegExp, s: string): string | null {
+  const match = s.match(regex);
+  return match ? match[1] : null;
+}
+
+export const parseZ: (s: string) => string | null = applyRegex.bind(
+  null,
+  /Z(\d+)/
```
manthey: Based on the spec in the document, I think these  regexs might be
`/(^|[_-])(z,slice)(\d+)/i`, `/(^|[_-])(t,time,daty,hour,min,wk,d,h)(\d+)/i`, `/(^|[_-])(xy,x,y,stage,pos)(\d+)/i`, `/(^|[_-])(ch,channel,wave,c)(.+?)([_-\.]|$)/i`

src/views/dataset/NewDataset.vue
```diff
@@ -75,9 +76,14 @@ function findCommonPrefix(strings: string[]): string {
     return strings[0];
   }
 
+  // Extract the non-metadata prefix of each filename. Note that because of the
+  // way the regex is constructed, the first match group will never be `null`.
+  const re = /(.*?)(_channel|_XY|_Z|_time|$)/;
```
manthey: With the more complicated regex, this has to be more complicated, too.

src/store/GirderAPI.ts
waxlamp: See fada1e5.

src/utils/parsing.ts
waxlamp: See d65ef36.

src/views/dataset/NewDataset.vue
waxlamp: See d65ef36.

PR #51, opened 2020-05-01
Full image cache

manthey:


manthey:
@waxlamp This should always use the correct starting images.  You can probably just use the current tile and remove the old tile references, since the tile should always have a `fullImage` entry.

We should give it a better name (`wholeDynamicRangeImage` is too long, though).

manthey:
There is also a lingering typescript error.

PR #63, opened 2020-05-19
Lay out images in a roughly square layout

waxlamp:


src/store/GirderAPI.ts
```diff
@@ -624,13 +630,15 @@ function parseTiles(items: IGirderItem[], tiles: ITileMeta[]) {
...
+  // Lay out images in a grid that is as roughly square as possible.
+  //
+  // TODO: this approach assumes all images have the same size.
   lookup.forEach(images => {
-    const cwidth = images.reduce((acc, image) => acc + image.sizeX, 0);
-    const cheight = images.reduce(
-      (acc, image) => Math.max(acc, image.sizeY),
-      0
-    );
+    const rowLength = Math.floor(Math.sqrt(images.length));
```
manthey, line 637: Do you think `ceil` would be better here so that the alignment tends to prefer wider to taller?

waxlamp: For my test case (21 images) it does feel more pleasant this way.

See 358ced194c350f8452a9c206596e6a21f009323c and let me know what you think.

PR #64, opened 2020-05-19
Add checkboxes to toggle unrolling by XY, Z, and T

waxlamp:
This PR adds checkboxes to control whether to unroll the images contained along the XY, Z, and/or T dimensions.

You can use [small3.tar.gz](https://github.com/Kitware/UPennContrast/files/4653464/small3.tar.gz) to test this feature.

~Typechecking is currently failing, though the application does work. Before merging I will fix the typechecking, but I wanted to get a review before doing so (as it may be a bit involved).~ **UPDATE:** typechecking fixed now (2020-05-21 17:54)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2011,6 +2011,11 @@
         "@types/babel-types": "*"
       }
     },
+    "@types/bluebird": {
+      "version": "3.5.32",
+      "resolved": "https://registry.npmjs.org/@types/bluebird/-/bluebird-3.5.32.tgz",
+      "integrity": "sha512-dIOxFfI0C+jz89g6lQ+TqhGgPQ0MxSnh/E4xuC0blhFtyW269+mPG5QeLgbdwst/LvdP8o1y0o/Gz5EHXLec/g=="
+    },
     "@types/body-parser": {
       "version": "1.17.1",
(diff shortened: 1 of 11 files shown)


waxlamp:
The word "splay" is somewhat unpleasant. Is "unroll" or some other word preferable?

manthey:
TODO:

- [ ] Verify that the shown histogram is the combined histogram of all splayed images.  If not, it should be.

- [ ] Prevent an excessive number of promises / requests from being executed at once.
    The Bluebird promise library allows easy sequentialization based on a parallelism parameters; it might be useful.

- [ ] Disable splay control if there is only one value in the range.

manthey:
> The word "splay" is somewhat unpleasant. Is "unroll" or some other word preferable?

Maybe "spread" or "array"?  I'm thinking that the antonym of whatever word we pick should also be appropriate, and I don't think of what we currently have as "rolled".

Ultimately we'll want the XY control to be a three state: "collapse"/"single", "spread"/"tessellated"/"unrolled", "position".

manthey:
@waxlamp I've fixed the issue where full-range images weren't using the combined histogram to inform how they were generated.  As far as I'm concerned, this is good to share and merge.

PR #65, opened 2020-05-20
If the histogram is empty, don't clamp it.

### LINKED ISSUE #19: Reset to saved doesn't always work
I have two layers.  The primary Z-Value is set to 1 on a three z-value data set.  On the second layer, I shift to Z-Slice Offset 1, set a range, and save it.  I then switched to Z-Slice Offset 2; the histogram shows values from 1 to 1.  When I switch back to Z-Slice Offset 1, the histogram max and min are both clamped to the low value.  I expected "reset to saved" to return to the values I saved, but it doesn't.

manthey:
This resolves #19.

### DIFF
--- a/src/components/ContrastHistogram.vue
+++ b/src/components/ContrastHistogram.vue
@@ -226,6 +226,10 @@ export default class ContrastHistogram extends Vue {
       return;
     }
 
+    if (data.max - data.min <= 1) {
+      return;
+    }
+
     // ensure the values are within the bounds
     const copy = Object.assign({}, this.value);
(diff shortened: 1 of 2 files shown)


manthey:
You can test the bad behavior and the fix by setting a layer to have a z, xy, or t offset and the histogram to absolute mode.  For instance, move to z = 1, set a layer to have a z offset of -1, adjust your histogram.  Move to z = 0.  The histogram won't be defined (since there is no z -1 image).  Move back to z = 1.  Before, the histogram would be collapsed and only resetting it would fix it.

Issue #70, opened 2020-05-28
Unroll two axes into an aligned grid


manthey:
As of PR #64, you can unroll images.  We arrange all unrolled images as squarely as possible, which makes a compact display.  However, when we unroll two axes, it would be better if one axis was arrayed horizontally and one vertically, even if the display is less square.  For instance, if we unroll XY and Z, have the horizontal be XY 0, 1, 2, 3, ..., and the vertical be Z 0, 1, 2, 3, ....

arjunrajlab:
_Audit notes (2026-07):_ The 2-axis aligned-grid logic actually exists in the metadata layer — `parseTiles` sets `rowLength = unrollCount[firstAxis]` when >1 axis is unrolled (src/store/GirderAPI.ts:1456-1460) — but the DISPLAY ignores it: ImageViewer.vue recomputes `unrollW` via square-packing (1292-1299) and maps tiles with that (1072). Fix: have ImageViewer honor parseTiles' row length for the multi-axis case. (Worth a quick browser check first.)

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #71, opened 2020-05-29
If unrolling two or more axes, layout in a grid.

manthey:
If you unroll two or more axes, the first axis is unrolled horizontally and the rest vertically.

This may not be ideal, especially if you unroll three axes.  For three axes, it might be better to determine which of axes 1&2 x axis 3 and axis 1 x axes 2&3 are squarer.

Further, the axes are unrolled in the order they are reported by the file, which isn't guaranteed to be the order of the axes in the UI.  For instance, in normmedia_8well_col2_livecellgfp.nd2, there are 6 XY, 7 Z, and 4 T values.  Since T is reported first, it unrolls T horizontally and XY or Z vertically, but we list time third in the UI.

### DIFF
--- a/src/store/GirderAPI.ts
+++ b/src/store/GirderAPI.ts
@@ -397,7 +397,12 @@ export default class GirderAPI {
     return image;
   }
 
-  generateImages(images: IImage[], color: string, contrast: IContrast) {
+  generateImages(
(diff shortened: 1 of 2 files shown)


manthey:
This resolves #70.

PR #78, opened 2020-11-16
Initialize with girder plugin, serve frontend from docker

subdavis:
Follows the same template I've used on several other projects now.

* [ldc](https://github.com/Kitware/ldc) for easy setup and development
* Dockerfile with multi-stage build for client
* Girder plugin that moves girder's base path to `/girder`
* Switch npm to yarn (hence the large change count)

Right now the plugin doesn't do anything except move the frontend and add the vue spa


subdavis:
* Netlify should probably just be removed from this project.

subdavis:
* This breaks the fetching of `modes.json` in `yarn serve` because I didn't want to temporarily introduce extra config.  It should work when using the client from the docker image.
* Netlify should probably just be removed from this project.
* Likewise, dockerhub automated build should probably be configured.

devops/girder.Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# Delcare Environment Variables
+ENV GIRDER_MONGO_URI mongodb://mongo:27017/girder
+ENV GIRDER_ADMIN_USER admin
+ENV GIRDER_ADMIN_PASSWORD upenn
+ENV CELERY_BROKER_URL amqp://guest:guest@rabbit/
+ENV BROKER_CONNECTION_TIMEOUT 2
+
+# TODO: explain why this is necessary
+# Install Large Image from source instead of pip because ?
```
subdavis, line 24: I'd like to have a better comment here.

server/girder_plugin/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+        #     os.environ.get('WORKER_API_URL', 'http://girder:8080/api/v1'),
+        # )
+        # Setting().set(
+        #     'worker.broker',
+        #     os.environ.get('CELERY_BROKER_URL', 'amqp://guest:guest@rabbit/'),
+        # )
+        # Setting().set(
+        #     'worker.backend',
+        #     os.environ.get('CELERY_BROKER_URL', 'amqp://guest:guest@rabbit/'),
+        # )
```
subdavis, line 33: Worker's role in this project is somewhat unclear to me.  I guess it's just for opaque large image background jobs?  I can probably remove this comment block.

devops/girder.Dockerfile
manthey: This was necessary early on in the project because nd2 support was still in a branch or large_image.  We should be able to use it from pypi (but probably want the --pre flag, since I think we have benefit from some changes that haven't been officially released let).

server/girder_plugin/__init__.py
manthey: We don't need a remote worker for this project; all of our images should be natively readable by large image.

devops/.env
```diff
@@ -0,0 +1,5 @@
+# Environment config for docker compose
+
+COMPOSE_PROJECT_NAME=upenn_contrast
+GIRDER_ADMIN_USER=admin
+GIRDER_ADMIN_PASS=letmein
```
manthey, line 5: I'd recommend setting the password to the same as the fallback password (elsewhere it is upenn).  It doesn't matter which is the default.

subdavis: done

manthey:
Closing this for now.

PR #80, opened 2021-02-26
Add a screenshot button and dialog to download images

manthey:
This adds a "Screenshot" button and dialog.

There are some different options.  The area and resolution can be 
- screen - this is the actual pixels on the screen.  Currently, this can only do one layer at a time due to a limitation in the canvas screenshot library that is employed; this will be fixed in the future.
- viewport - the area of the screen, but rendered from the source file.  This can be higher resolution.  The output will either be the native image size or no larger than the specified value or 10,000 in the maximal dimension.
- full - the entire image rendered from the source file.  The output will either be the native image size or no larger than the specified value or 10,000 in the maximal dimension.
The viewport and full options aren't available when one or more axes is unrolled.  If this is desired, it will be more work.

The output can either be PNG or jpeg at a few set qualities.  It would be easy to give more quality options.  It would be easy to offer tiff output for viewport and full areas (but more work to do so for the screen).

This will eventually have support for saving a url to a specific view.  For now, that capability can sort of be obtained by duplicating the configuration.


manthey:
I'll merge this; code for urls for specific views will be in a separate PR.

PR #128, opened 2021-05-21
Precache histograms and full range images.

### LINKED ISSUE #95: Z-scrolling speedup: scrolling optimization and prioritization
Z-scrolling needs to be faster. Here is a priority list of what should be cached:
1. Contrast: top priority, but the current scheme of working with local data and then pulling the high res version after is perfectly fine, so this can be considered done, I think.
2. Z and time are the next highest priority ones that need teh snappy. Could cache either/both. For some applications, z probably more important, for others, time. Usually, this will be pretty clear from the data, because it will either have Z or time but not both. Perhaps we can choose which to cache first by whatever is has the most points. Like, if there are 50 Z slices and 4 time slices, then cache z first. If there are 2 Z slices and 49 time slices, then cache time first.
3. X-Y probably does not need to be cached, since people will usually go through that more slowly.

### LINKED ISSUE #124: Performance of Z max-merge: how much does it merge at once?
Switching to Z max-merge seems to take a while. I'm guessing it does the max-merge for all XY positions/times, and not just the one currently shown, right? If so, is it possible to switch it to doing the max-merge just for the current frame to make it a bit more performant? Could perhaps cache it to enable faster moving around after having "visited" a frame and merged it once?

manthey:
When the map is idle, precache histograms and full range images for the longest axis.

Closes #124.  Closes #95.

### DIFF
--- a/src/components/ImageViewer.vue
+++ b/src/components/ImageViewer.vue
@@ -35,7 +35,7 @@ import {
   zoomIdentity,
   zoomTransform
 } from "d3-zoom";
-import { IImageTile } from "../store/model";
+import { IImage, IImageTile } from "../store/model";
 
 function generateFilterURL(
   index: number,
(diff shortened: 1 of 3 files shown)


manthey:
@arjunrajlab Any objection to merging this?  Any needed changes can be on a new PR.

arjunrajlab:
Go for it!

PR #130, opened 2021-06-21
Change the default filename for image download

### LINKED ISSUE #120: Change default filename for image download.
Current filename is _dataset_-region.tif (e.g.), we should make it so it is _dataset_-_configuration_-_snapshot_-date.tif or something like that.

boucaud:
closes #120

### DIFF
--- a/src/components/Snapshots.vue
+++ b/src/components/Snapshots.vue
@@ -283,6 +283,7 @@
 import { Vue, Component, Prop, Watch } from "vue-property-decorator";
 import store from "@/store";
 import geojs from "geojs";
+import { formatDate } from "@/utils/date";
 
 @Component
 export default class Snapshots extends Vue {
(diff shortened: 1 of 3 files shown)


src/components/Snapshots.vue
```diff
@@ -369,6 +369,16 @@ export default class Snapshots extends Vue {
   getBasicDownloadParams() {
     let w = store.dataset!.width;
     let h = store.dataset!.height;
+
+    // Determine fileName
+    const datasetName = store.dataset?.name || "unknown";
+    const configurationName = store.configuration?.name || "unknown";
+    const snapshotName = this.newName || "unknown";
```
manthey, line 376: When there isn't a snapshot name, we might just want to leave this component out of the filename, rather than have the `unknown` present.  @arjunrajlab Do you have a preference?

arjunrajlab: Agreed, probably better not to have `unknown`.

manthey (approved):

Issue #87, opened 2021-05-14
When setting Z-slice to "Constant", use current Z value


arjunrajlab:
When setting Z-slice to "Constant", use current Z value from the slider. Right now, it just uses "0" as a default.

arjunrajlab:
Problem remains:
![image](https://user-images.githubusercontent.com/13998824/126043275-8980a735-d7ff-49fb-ad87-9a7c8543e48d.png)

PR #134, opened 2021-06-30
When setting a slice to "Constant", use the current displayed value

boucaud:
#87


manthey:
This works if you switch from "current" to "constant".  If you switch from "current" to "offset" and then to "constant", it still is set at 0.

boucaud:
> This works if you switch from "current" to "constant". If you switch from "current" to "offset" and then to "constant", it still is set at 0.

This should be fixed now.

manthey (approved):

PR #136, opened 2021-07-02
Make sliders show upper bounds and start at 1

### LINKED ISSUE #90: Label slider bounds
Sliders start from 0 (should start from 1), should also display the upper bound somewhere on the UI. It is very helpful for users, because it helps them orient themselves in their data. If we can show tick marks, great, but not necessary yet.

boucaud:
closes #90 

DeepinScreenshot_select-area_20210702144557

### DIFF
--- a/src/components/ValueSlider.vue
+++ b/src/components/ValueSlider.vue
@@ -1,24 +1,39 @@
 <template>
   <v-slider
     v-model="slider"
-    :max="max"
-    :min="min"
+    :max="max + offset"
+    :min="min + offset"
     hide-details
     :label="label"
     :disabled="min === max"
   >
     <template #append>
-      <v-text-field
-        v-model="slider"
-        class="mt-0 pt-0"
-        hide-details
-        single-line
-        type="number"
-        :min="min"
-        :max="max"
-        :disabled="min === max"
-        style="width: 60px"
-      ></v-text-field>
+      <v-container>
+        <v-row>
+          <v-col class="pa-0">
+            <v-text-field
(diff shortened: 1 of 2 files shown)


manthey:
This is behaving oddly for me.  If I click the up or down arrow in the number spinner, it seems to take two clicks to change the number instead of 1, and I don't see the even being emitted correctly (so the image doesn't change).

boucaud:
> This is behaving oddly for me. If I click the up or down arrow in the number spinner, it seems to take two clicks to change the number instead of 1, and I don't see the even being emitted correctly (so the image doesn't change).

This should be fixed now

manthey:
I really like the way this looks for small numbers, but for large numbers it doesn't quite work:
image

Perhaps printing the total as text below the baseline of the number entry field would be better?

boucaud:
> Perhaps printing the total as text below the baseline of the number entry field would be better?

DeepinScreenshot_select-area_20210715154613

Is that what you meant ? If so, does it look ok ?

arjunrajlab:
I like it! I think having it be "of 2801" instead of "/2801" might be a bit clearer, but having it below seems nice!

boucaud:
> I like it! I think having it be "of 2801" instead of "/2801" might be a bit clearer, but having it below seems nice!

I've pushed a change to have it look like this:

DeepinScreenshot_select-area_20210715163724

arjunrajlab:
That looks good to me!

manthey:
Excellent.

manthey (approved):

PR #144, opened 2021-07-16
Used a shared tile queue across layers.

manthey:
This should make the browser somewhat more responsive, as it will have fewer queued requests after panning or adjustments.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1901,6 +1901,13 @@
       "integrity": "sha512-f5j5b/Gf71L+dbqxIpQ4Z2WlmI/mPJ0fOkGGmFgtb6sAu97EPczzbS3/tJKxmcYDj55OX6ssqwDAWOHIYDRDGA==",
       "dev": true
     },
+    "@types/parse-json": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/@types/parse-json/-/parse-json-4.0.0.tgz",
+      "integrity": "sha512-//oorEZjL6sbPcKUaCdIGlIUeH26mgzimjBB77G6XRgnDl/L5wOnpyBGRe/Mmf5CVW3PwEBE1NjiMZ/ssFh4wA==",
+      "dev": true,
+      "optional": true
+    },
     "@types/q": {
       "version": "1.5.4",
       "resolved": "https://registry.npmjs.org/@types/q/-/q-1.5.4.tgz",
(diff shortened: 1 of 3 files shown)


arjunrajlab:
So… I played with it, and I do think it is perhaps a bit more responsive in some ways. It definitely seems to be much more robust once the cache is fully loaded, I think—before, if I scrolled in Z out of range, it would sometimes have to reload the cache, and I think that has been alleviated. I still think the cache loading will need to be quite a bit faster, although it does seem to be more intelligent about not starting over.

The progress bar is still a bit wonky in when it shows up, but seems perhaps a bit more reliable than before. Not critical at this stage.

I have noticed that the image seems to "build" more now, in the sense that it progressively layers each layer in a noticeable way. Ultimately, especially when scrolling through Z, it would be good to just show the composite and switch out the image buffer once the new image composite is completely built, but I'm assuming that will be handled later as per our discussions about the "blink" issue.

arjunrajlab:
Also, for whatever reason, I've noticed the Docker is using a lot of CPU, even when the browser window is closed. Perhaps a one time thing, not sure just yet.

arjunrajlab:
One thing I noticed that seems different is the following:

image

Not sure what the amd64 thing is about. The process that is eating up a lot of CPU appears to be related to some AMD64 emulation:
image

Not sure if that is different than before, but I don't remember noticing this previously.

arjunrajlab:
Hmm. I switched to the other branch and the AMD64 emulation thing is still going.

manthey:
Do you have one of the newer Macs with an Apple CPU (rather than Intel or AMD)?  If so, that might explain the AMD64 indicator.

arjunrajlab:
Indeed, that was on my laptop with an M1 chip in it (which is awesome BTW :)). I figured that would be the reason, but it's a little strange because the OS usually emulates Intel code directly through Rosetta2, so I'm surprised Docker runs the emulation. Anyway, yes, that's probably the issue because it doesn't seem to be doing that on my work computer, which is an Intel iMac. On my work iMac, I do see the following, though:
image
Anyway, probably not worth worrying about at this stage, just an FYI.

arjunrajlab:
Also, looking at the master branch now, I think the progressively-building-the-layer issue is the same on the old branch, so that's not an issue specific to this new code.

manthey:
I think this branch is an improvement over master, even if it doesn't fully address the problem.  I propose merging it and then having a further PR for the rest of the issue.

arjunrajlab:
Agree completely, definitely improved! Merging makes sense.

PR #145, opened 2021-07-22
WIP: Add annotations and tools 

boucaud:
Note: this is very much a work-in-progress

What works:
* Tool creation based on predefined template
* Tools are saved to girder
* Adding/removing tools to the configuration's toolset
* Using annotation creation tools
* Joining dot annotations to the closest blob annotation

Next steps (roughly in order):
- [ ] Save annotations and connections to girder
- [ ] Channel and XY/Z/Time locations for annotations
- [ ] Load saved annotations when needed, and display them on the relevant layers
- [ ] Connect to nearest for all annotation shapes, with channel restrictions
- [ ] Tool creation validation and better defaults
- [ ] Tidy up UI elements
- [ ] Clean up the branch. At this point annotation creation and toolsets should be fully functional and this could potentially be merged as a first step ?

Then, add support for snapping, selection, edition, the annotation browser.


arjunrajlab:
After reloading URL, the connections appear to not show:
image

arjunrajlab:
I feel like "Add tool" should by default add to the current toolset. It's a little awkward to create the tool and then add existing tool. Perhaps rename button to "Add tool to current toolset".

image

arjunrajlab:
"Current toolset" is sort of awkward. I would maybe just say "Annotation tools".
image

arjunrajlab:
Not sure what the functionality of the "Reset" button is. I would just remove it. I think most would just cancel and start again if needed.
image

arjunrajlab:
Blob creation requires a double click to end, even in freehand mode. If you freehand draw, I feel like it should just automatically close the blob upon mouse-up.

arjunrajlab:
"Connection restrictions" would be better named "Connect to nearest tag…". Also, "Restrict to layer" might be better as "Connect to annotation in layer…" with the default option being "Any" instead of "All".
image

arjunrajlab:
It would be nice to have tag autocomplete and maybe a tag browser if possible.
image

arjunrajlab:
Headings like "General Tool Properties" and "Annotation Description" are probably unnecessary. Also, the "Tool description" field takes up a lot of room, could probably shrink it down a lot.
image

PR #146, opened 2021-07-22
Fix layer slice indexes not updating sometimes, and starting from 0

boucaud:
#87

### DIFF
--- a/src/components/DisplayLayer.vue
+++ b/src/components/DisplayLayer.vue
@@ -83,6 +83,7 @@
         :max-value="maxXY"
         v-if="maxXY > 0"
         :displayed="displayXY"
+        :offset="1"
       />
       <display-slice
         :value="value.z"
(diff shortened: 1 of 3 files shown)


manthey:
offset isn't working right for me.  When I click on the text field and try to type, it keeps going to the maximum.

We probably don't want to constrain the offset to the current possible values, but I'm not sure what is the actual desired behavior.  As an example if I have 6 z slices and 2 layers, and 1 layer is set to current and the other to an offset of 2, what should be displayed when we navigate to z = 5 or 6?  Or should we prevent navigation to z = 5 or 6?  @arjunrajlab Do you have an opinion on what should be displayed?

There is presently a bug where we can throw an exception if layer 1 has the offset (I'll submit a fix for that shortly) and we navigate to where the offset is outside of the available values.

manthey (approved):

PR #148, opened 2021-08-20
Use background quads for smoother transitions.

manthey:
This has some tunable parameters to fetch more background quad images or use more or less texture memory.  Using more background textures results in some texture memory loading spikes that manifest as some initial hesitation in z scrolling, but produce nicer images.


manthey:
Make sure you rebuild the dockers and rerun npm install to test this branch.

PR #150, opened 2021-09-03
Improve DatasetInfo layout

### LINKED ISSUE #139: Non-intuitive behavior of "Remove" button in configuration view
image
The placement of the "REMOVE" button makes it seem like it would somehow remove a configuration, when in fact it is for removing the dataset itself. I think it might be better to place this button above the list of configurations so that it is more clearly associated with the Dataset, and change the button name to "REMOVE DATASET" so that it is clear exactly what it does.

boucaud:
Fixes #139 

current layout:

Screenshot from 2021-09-03 10-57-13

new layout:

Screenshot from 2021-09-03 11-12-58


arjunrajlab:
What is the difference between Tool name and annotation name? I guess annotation name sets the name of the annotation? Annotation name, if assigned to every annotation, is probably just the same as a tag, so I feel like we might not need it here. I think Annotation name could be a field better set in the annotation browser. Or could also be set by right clicking the annotation.

image

arjunrajlab:
I feel like Tool Name and Tool Description should come before the pop-up for Tool Type.

image

arjunrajlab:
There are two sections here, can they be laid out side to side instead of on top of each other? That would save a lot of space and would also make it easier to see that the pop-ups are actually pop-ups (as opposed to editable text fields) because right now the little triangle is far to the right.

image

arjunrajlab:
The width of many of these elements could be dramatically compressed, which would save a lot of space.
image

PR #153, opened 2021-09-27
Change tile frames defaults.

manthey:
This uses smaller but more textures.

This also adjusts a cache setting on the server to allow caching more tile frame textures.


manthey:
I'll merge this for now.  Subsequent work will (a) start caching textures as soon as an image set is uploaded, (b) show an indicator of when such textures are loaded and available.

PR #162, opened 2021-11-25
Implementation of the annotation girder plugin and browser

boucaud:
## Description

This adds a girder plugin to store and compute annotations and properties in the backend, as well as changes to the front-end to use the new API, browse and filter annotations and manage properties.

Note: this includes snapping features that will probably be moved out of this branch.
For now, To build the necessary wasm files, run `npx itk-js build itk`

### The annotation plugin
The girder plugin is located in `devops/girder/plugins/upenncontrast_annotation`.

It adds four new API endpoints: `/upenn_annotation`, `/annotation_connection`, `/annotation_property` and `/annotation_property_values`. Models and database collections are also created for each of these endpoints.

The endpoints among other things let users manage (create, read, update, delete) annotations, connections, properties and computed property values. There are also more advanced search functionalities as well as an endpoint to compute property values (see below). API descriptions are available at [http://localhost:8080/api/v1](http://localhost:8080/api/v1) with a running server.
 
The plugin should be loaded automatically, does not require additional operations to work.

There are however steps to follow in order to use the new property computation infrastructure (see below)

### The annotation browser

The annotation browser is a new interface component that lets users browse and filter annotations for the current dataset. It is accessible by clicking the top-right "Browse Annotation buttons".

DeepinScreenshot_select-area_20211125170053

It features four sections: general controls to manage annotation visualization, filters, properties, and an annotation list.

The filters control which annotations will be shown in the annotation list. By default, users can filter annotations by drawing ROIs in the main view, by specifying tags and shape, and by checking annotations in the list.
If properties are present and have been computed, users can also filter annotations with ranges on property values.

By default, no properties are present in the database and they have to be created manually. Users can create properties by clicking the '+' button in the relevant section. Properties need a name, a category, and the name of the docker image that implements their computation (see below).

Properties and property values do not currently automatically refresh all the time, and users should use the 'refresh button' to fetch new properties and values from the database if the displayed values are outdated.

#### Configuration steps

The new girder plugin can trigger the computation of annotation property values from docker images. To do so, multiple configuration steps have to be followed:

In the same node that hosts the girder server:
* Clone the example docker image repository here: `git clone git@github.com:boucaud/sample_intensity_worker.git`
* Move to this new directory: `cd sample_intensity_worker`
* Build the docker image, and specify the chosen image name and tag. This image name will need to be used later when creating properties in the annotation browser. Example: `docker build . -t boucaud/sample_intensity_worker:latest`

Specify the girder API url to girder_worker (@manthey can this be done with a configuration file instead ?):
* Navigate to [http://localhost:8080/#plugins/worker/config](http://localhost:8080/#plugins/worker/config)
* In the "Alternative Girder API URL, paste this URL: `http://girder:8080/api/v1/`
* Click "Save"

Finally, create a new property and compute:
* In the annotation browser, click the '+' button to create a new property.
* Enter the desired name (for the example, `radius_intensity`
* Enter the docker image name and tag (for the example, `boucaud/sample_intensity_worker:latest`)
* Choose the `layer` category
* Submit

The property should now be visible in the browser. You can configure which layer should be used for the calculations.

Clicking the "Computed" checkbox will start the computation for the current dataset.


boucaud:
@zjniu issues #170 #175 #176 #177 #178 #179 #180 #181 should be fixed in the latest version of this branch and the latest version of the worker.

Can you check that they are indeed fixed on your end ?

PR #195, opened 2022-02-25
Use the quad_info endpoint for tile_frames.

manthey:
Now, precaching the tile frames can be done by calling the tile_frames/quad_info endpoint with cache=schedule.  To match what is done in the javascript side, this endpoint needs to be called per channel with frameBase=<channel>, frameStride=<number of channels>, frameGroup=<IndexZ || 1>, frameGroupStride=<IndexZ / IndexC>.  This is in addition to the defaults used in this project defined in src/store/index.ts around line 745.

To make this more convenient, we could add (to large_image) a special value for frameBase which would schedule all tile_frames based on this heuristic.

### DIFF
--- a/src/components/ImageViewer.vue
+++ b/src/components/ImageViewer.vue
@@ -441,14 +441,14 @@ export default class ImageViewer extends Vue {
             if (!fullLayer.setFrameQuad) {
               setFrameQuad(someImage.tileinfo, fullLayer, {
                 ...baseQuadOptions,
-                progress: () => {
-                  if (this.cacheProgress < this.cacheProgressTotal) {
+                progress: status => {
+                  if (status.loadedCount === 0) {
+                    this.cacheProgressTotal += status.images.length;
+                  } else if (this.cacheProgress < this.cacheProgressTotal) {
                     this.cacheProgress += 1;
                   }
                 }
               });
-              this.cacheProgressTotal +=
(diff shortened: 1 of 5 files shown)


arjunrajlab:
Is everything there for us to test it out? I think incorporating the defaults into large_image would make the most sense, long term.

manthey:
Once https://github.com/girder/large_image/pull/792 is merged, you can hit the `GET` `/item/{itemId}/tiles/tile_frames/quad_info` endpoint to schedule all appropriate tile_frames computation and caching given an image's item ID.  Specifically, for this project, we will want to use the following parameters:
```
{
    'query': 'style=%7B%22min%22%3A%22min%22%2C%22max%22%3A%22max%22%2C%22palette%22%3A%5B%22%23000000%22%2C%22%23ffffff%22%5D%7D&cache=true',
    'maxTextureSize': 4096,
    'maxTextures': 32,
    'frameBase': 'c',
    'frameStride': 'c',
    'frameGroup': 'z',
    'frameGroupStride': 'auto',
    'cache': 'schedule'
```
The `query` parameter is the "auto brightness" used to make the preview cover as much dynamic range as possible.  The `maxTextures` and `maxTextureSize` are values we decided work well for this project.  The `frame...` parameters are the automation to generate separate textures by channel and then within that group z sequences together.  The `cache` parameter says to schedule caching if it isn't already done.

PR #202, opened 2022-03-18
Schedule quad_info computation on dataset upload

boucaud:


src/store/GirderAPI.ts
```diff
@@ -474,6 +474,29 @@ export default class GirderAPI {
...
+
+  scheduleTileFramesComputation(datasetId: string) {
+    return this.getImages(datasetId).then((items: IGirderItem[]) => {
+      return items.map((item: IGirderItem) => {
+        return this.client.get(
+          `/item/${item._id}/tiles/tile_frames/quad_info`,
+          {
+            params: {
+              query:
+                "style=%7B%22min%22%3A%22min%22%2C%22max%22%3A%22max%22%2C%22palette%22%3A%5B%22%23000000%22%2C%22%23ffffff%22%5D%7D&cache=true",
```
manthey, line 486: I don't know if we want to do
```
            "style=" +
            encodeURIComponent(
              JSON.stringify({
                min: "min",
                max: "max",
                palette: ["#000000", "#ffffff"]
              })
            ) +
            "&cache=true"
```
to match the code look in src/store/index.ts (or put them in common somewhere so it isn't repeated).

manthey:
I'm merging this.  We can revert if there was a reason not to do so.

PR #209, opened 2022-04-01
Add a getRegion convenience method.

manthey:


manthey:
@zjniu Try this again.  If you instantiate the class with the datasetId, then you don't need to pass it to the getRegion method (but you can).

boucaud (approved):

PR #223, opened 2022-05-11
Increase the tile queue tracking parameter.

manthey:
The default is large enough for a single layer, but we have lots of layers.  Making it bigger uses more browser memory but prevents trying to discard tiles needlessly.  Ideally, we'd modify geojs so when layers share a tile queue the tracking parameter would be a per-lyater value not a fixed value.

There is even a note in geojs saying if the track value is too low, it will cause needless computation.

I may automate this in geojs instead of the merging this PR.


manthey:
@boucaud This was the actual issue with images with many tiles having poor performance -- the call to `toString()` wasn't really the culprit.

Issue #240, opened 2022-07-22
Toggle selection


finetjul:
On 2022/07/01, @arjunrajlab suggested to: 
 - on a new selection, toggle the current selection with the new selection. e.g. if `currentSelection={A, B, C}` and `newSelection={B, D, E}`, then new `currentSelection={A, C, D, E}`.

finetjul:
On 2022/07/06, @luciemac suggested to add a "selection mode" button to control the behavior of the new selection:
 - NEW: the new selection clears the previous selection
 - ADD: the new selection is added to the previous selection
 - SUBSTRACT: the new selection is removed from the previous selection

luciemac:
@finetjul guess we can close this issue?

finetjul:
Yes #242 has implemented it, thanks.

Issue #236, opened 2022-07-18
Channel toggle very buggy.


zjniu:
Toggling channels on or off is currently very buggy. There are many cases when toggling off or on a channel has no effect. An example of how to reproduce such an issue is the following. Suppose we have channels A and B.

1. First turn off all channels.
2. Turn on channel A.
3. Turn on channel B.
4. Turn off channel B.
5. Turn off channel A.

Now, if we try to turn on channel A, nothing happens. Note that this is only one example and there are other ways to trigger this behavior.

arjunrajlab:
I think this mostly works now, according to @zjniu. Apparently if you toggle really fast it can still be an issue, but I think it's fine for now.

PR #242, opened 2022-07-22
Add select tool

### LINKED ISSUE #233: Point selection tool selects multiple points in close proximity.
When using the point selection tool, clicking either of two points in close proximity selects both points. As an example, in the image below, when clicking on either of the two leftmost points (0 and 1 in the annotation browser), both get selected. I assume this is because the current implementation just searches for points within a fixed radius around the point clicked by the user. Perhaps it would be useful to limit the selection to simply the closest point? I verified that using the lasso tool on just one of the point avoids this issue.
image

luciemac:
Closes #233, #236,  #240, #245

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -678,11 +678,15 @@ export default class AnnotationViewer extends Vue {
         );
       });
 
-    // Update annotation store (make sure annotations are selected in the Annotation Browser)
-    this.annotationStore.appendSelected(selectedAnnotations);
+    // Update annotation store
+    // TODO: should add / Remove / toggle instead of only appending
+
+    // this.annotationStore.selectAnnotations(selectedAnnotations);
+    this.annotationStore.toggleSelected(selectedAnnotations);
+    // this.annotationStore.unselectAnnotations(selectedAnnotations);
 
     // Remove the selection annotation from layer (do not show the annotation used to select)
-    this.removeSelectionAnnotationFromLayer(selectAnnotation);
+    this.annotationLayer.removeAnnotation(selectAnnotation);
   }
 
   private async addAnnotationFromGeoJsAnnotation(annotation: any) {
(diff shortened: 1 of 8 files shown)


src/components/AnnotationViewer.vue
```diff
@@ -503,6 +534,161 @@ export default class AnnotationViewer extends Vue {
...
+      // If the selection annotation type is "Polygon"
+      // Check if the tested annotation (independently from its type)
+      // is in the defined polygon
+      return annotation.coordinates.reduce((isIn: boolean, point: any) => {
+        const isPointIn = geojs.util.pointInPolygon(
+          point,
+          selectionAnnotationCoordinates
+        );
+        return isIn || isPointIn;
+      }, false);
```
finetjul, line 601: it looks like you do not need to browse them all but stop as soon as you found one in. => `annotation.coordinates.some()`

src/components/AnnotationViewer.vue
```diff
@@ -503,6 +534,161 @@ export default class AnnotationViewer extends Vue {
...
+      .filter((annotation: IAnnotation) => {
+        return (
+          (selectLocation.XY === annotation.location.XY ||
+            this.store.unrollXY) &&
+          (selectLocation.Z === annotation.location.Z || this.store.unrollZ) &&
+          (selectLocation.Time === annotation.location.Time ||
+            this.store.unrollT)
+        );
+      })
+      .filter((annotation: IAnnotation) => {
```
finetjul, line 632: why split in 2 `filter()` calls instead of one ?
```
this.annotations.filter((annotation: IAnnotation) => {
 return sameLocation(selectLocation, annotation, this.store.unrollXY, this.store.unrollZ, this.store.unrollT) &&
            this.shouldSelectAnnotation(type, coordinates, annotation, unitsPerPixel);
 }
```

src/components/AnnotationViewer.vue
luciemac: We do need to browse them all. 
The idea here, is to get annotations within a manually defined polygon.
We have to go over every annotations and check if it is in the polygon.
We don't want to stop as soon as we found one. We want all of them.

finetjul: If you need ALL to be true (in the polygon), it should have been `isIn && isPointIn` .

In such case, `annotation.coordinates.every()` would be what you are looking for.

luciemac:
Closes #240 #245 
It should also fix #236

@zjniu If you want to test

PR #264, opened 2022-09-07
Improve the test environment.

manthey:
Specifically, the Girder plugin setup.py file was missing some requirements.  The tox file should specify using find-links to fetch large_image wheels.  The Dockerfile now more conveniently starts in the directory where the plugin is located.  Lastly, there is a test_env.sh script that can be run in the docker to allow girder tests to properly find mongo (unfortunately, girder tests expect mongo to be on localhost, which isn't the case in a docker-compose environment).

### DIFF
--- a/devops/girder/Dockerfile
+++ b/devops/girder/Dockerfile
@@ -1,7 +1,12 @@
 FROM girder/girder:latest-py3
 
 RUN apt-get update && \
-    apt-get install --yes --no-install-recommends fuse
+    apt-get install --yes --no-install-recommends \
+    fuse \
+    iptables \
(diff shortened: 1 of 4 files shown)


manthey:
I'll note that the Dockerfile uses the master branch of large_image, while tox is using the latest published version.

manthey:
@luciemac I can't remember if you tried this out.  We should merge it in if there are no issues with it.

luciemac:
@manthey, I tried it, but could not make it work on my machine.

arjunrajlab:
Bump on this PR.

PR #266, opened 2022-09-14
Bug fix: right panel overlap

bruyeret-kitware:
Bug found: opening both right panels together ("Snapshot" and "Browse annotations") creates a UI bug
Fixed it by allowing only a single right panel to be open at a time
Also added `/assetstore` and `/db` to `.gitignore`

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,11 +7,11 @@
       >
       <bread-crumbs />
       <v-spacer />
-      <v-btn @click.stop="snapshotPanel = !snapshotPanel" id="snapshotButton"
+      <v-btn @click.stop="updateRightPanel('snapshotPanel')" id="snapshotButton"
         >Snapshot</v-btn
       >
       <user-menu />
-      <v-btn @click.stop="annotationPanel = !annotationPanel"
+      <v-btn @click.stop="updateRightPanel('annotationPanel')"
         >Browse Annotations</v-btn
       >
       <server-status />
@@ -83,6 +83,8 @@ export default class App extends Vue {
 
   annotationPanel = false;
 
+  lastRightPanelUpdate: string | undefined;
+
   fetchConfig() {
     // Fetch the list of available tool templates
     // It consists of a json file containing a list of items, each item describing
@@ -108,6 +110,18 @@ export default class App extends Vue {
 
   goHome() {
     this.$router.push({ name: "root" });
+  }
+
+  updateRightPanel(panel: string) {
+    this.$data[panel] = !this.$data[panel];
+    // The last panel updated has to be closed if it is not the currently updated panel
+    if (
+      this.lastRightPanelUpdate !== undefined &&
+      this.lastRightPanelUpdate !== panel
+    ) {
+      this.$data[this.lastRightPanelUpdate] = false;
+    }
+    this.lastRightPanelUpdate = panel;
   }
 }
 </script>
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,11 +7,11 @@
       >
       <bread-crumbs />
       <v-spacer />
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -109,6 +111,18 @@ export default class App extends Vue {
   goHome() {
     this.$router.push({ name: "root" });
   }
+
+  updateRightPanel(panel: string) {
```
finetjul, line 115: seems more like a `toggleRightPanel()` than `updateRightPanel()`

src/App.vue
```diff
@@ -109,6 +111,18 @@ export default class App extends Vue {
...
+    this.$data[panel] = !this.$data[panel];
+    // The last panel updated has to be closed if it is not the currently updated panel
+    if (
+      this.lastRightPanelUpdate !== undefined &&
+      this.lastRightPanelUpdate !== panel
+    ) {
+      this.$data[this.lastRightPanelUpdate] = false;
+    }
+    this.lastRightPanelUpdate = panel;
```
finetjul, line 124: `this.lastRightPanelUpdate` -> `this.lastModifiedRightPanel` ?

src/App.vue
```diff
@@ -109,6 +111,18 @@ export default class App extends Vue {
   goHome() {
     this.$router.push({ name: "root" });
   }
+
+  updateRightPanel(panel: string) {
+    this.$data[panel] = !this.$data[panel];
+    // The last panel updated has to be closed if it is not the currently updated panel
+    if (
+      this.lastRightPanelUpdate !== undefined &&
```
finetjul, line 119: this.lastRightPanelUpdate != null &&

src/App.vue
```diff
@@ -112,16 +112,16 @@ export default class App extends Vue {
...
   }
 
-  updateRightPanel(panel: string) {
+  toggleRightPanel(panel: string) {
     this.$data[panel] = !this.$data[panel];
     // The last panel updated has to be closed if it is not the currently updated panel
     if (
-      this.lastRightPanelUpdate !== undefined &&
-      this.lastRightPanelUpdate !== panel
+      this.lastModifiedRightPanel !== null &&
```
finetjul, line 119: `this.lastModifiedRightPanel != null` to compare with null AND undefined

manthey: It is often considered better practice to be explicit about comparing null and undefined (e.g., `foo !== null && foo !== undefined`.  The 2021 spec for ES adds the new nullish operator (`??`) to confuse things further.

bruyeret-kitware: The declaration of `lastModifiedRightPanel` is the following: `lastModifiedRightPanel: string | null = null;`
It means that it shouln't be possible for `lastModifiedRightPanel` to be `null`
Should I edit anyway to compare against both `null` and `undefined`?

finetjul: > It means that it shouln't be possible for lastModifiedRightPanel to be `null`

You meant "to be `undefined`", right ?

> It is often considered better practice to be explicit about comparing null and undefined

I personally find it too verbose, error prone (if developer forgets that the variable can become undefined) and potentially harmful when it comes to performance.

Now, in a typescript environment things are a bit different. I lack experience here.

I guess it is all a matter of making a choice (setup linter rules) and enforcing it all over the project. I can see both `!=`/`==` vs `!==`/`===` being used in the code. Your call.

And back to this very specific issue, because you enforced the type to be `string | null`, I guess that gives you enough safety to write `!== null`, sorry about this fuzz.

bruyeret-kitware: > You meant "to be `undefined`", right ?

Yes, sorry for the mistake!

I'll keep it as it is right now then, thank you!

Issue #261, opened 2022-09-02
Turning on and off annotations is slow.


zjniu:
When there is a high number of annotations, turning on and off annotations takes a long time. See https://github.com/Kitware/UPennContrast/issues/228#issuecomment-1206866452 for a worked example for recreating.

finetjul:
Some ideas discussed so far:
 - reproduce the issue
 - identify the reason why it is slow/crash (@manthey said geo.js should handle hundred thousands annotations)
 - potentially replace polygons with centroids for display when zooming out
 - do not display annotations when zooming out when the number of annotations is very large
 - create a heatmap when zooming out

arjunrajlab:
Further update: appears that turning annotations ON is relatively fast, but turning off seems to take a long time.

bruyeret-kitware:
I did a small bug fix on master to speed up annotation turn off, do you see a change?

arjunrajlab:
I think that worked! Thank you!

PR #267, opened 2022-09-15
Fix some issues with layers not always turning on or off.

manthey:
Before, you could occasionally switch to single layer mode and have two or zero channels showing.

This was caused by calling a function to toggling the layer that checked if the target event was an input element instead of just doing it.

### DIFF
--- a/src/components/DisplayLayer.vue
+++ b/src/components/DisplayLayer.vue
@@ -157,7 +157,7 @@ export default class DisplayLayer extends Vue {
     if (this.visible === value) {
       return;
     }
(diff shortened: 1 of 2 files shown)


luciemac:
LGTM

luciemac:
@manthey 
Do you think it can be related to issue #261 ?

manthey:
> @manthey Do you think it can be related to issue #261 ?

I don't think so -- this is actually a problem where we were toggling the control to indicate if a layer was on or off, but not toggling the actual state.

PR #270, opened 2022-09-20
Improve connections

luciemac:
@zjniu 
Here is the PR. If you can try it, and let me know if something is not working the way it should.
There is a new route: _annotation_connection/connectTo_ (from localhost: http://localhost:8080/api/v1/annotation_connection/connectTo)

Body content should be a json defined as below:
{
	"annotationsIds": [
		"id1",
		"id2",
		"idN"
	],
	"tags": ["tag1", "tag2"],
	"channelId": "0"
}

with (similar to what is done on front side):
- annotationsIds: a list of annotations ids you want to find connection with other annotations [1]
- tags: a list of tags to identify annotations that [1] should connect to
- channelId (optional  - because on front side you can select "Any")

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/connections.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/connections.py
@@ -7,8 +7,11 @@ from bson.objectid import ObjectId
 from girder.models.folder import Folder
 
 from .annotation import Annotation
+from ..helpers.connections import annotationToAnnotationDistance
 
 import jsonschema
+import math
+import numpy as np
 
 class ConnectionSchema:
     tagsSchema = {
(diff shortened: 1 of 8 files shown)


devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/connections.py
```diff
@@ -92,10 +95,81 @@ def validate(self, document):
 
   def create(self, creator, connection):
     self.setUserAccess(connection, user=creator, level=AccessType.ADMIN, save=False)
+    a = self.save(connection)
```
finetjul, line 98: needed ?

finetjul (commented):
Some files are missing EOL

luciemac:
@finetjul Ready for an other pass if you want

finetjul:
LGTM

PR #271, opened 2022-09-20
Added tooltips

bruyeret-kitware:
Option and hotkey (T) to enable/disable tooltips.

Three additive options to decide which tooltips to display :
- All
- Selected
- Hovered

Hovering event detection works only after the first tool deselection and when no tool is selected.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-app id="inspire">
+  <v-app v-mousetrap="mousetrapBindings" id="inspire">
     <v-app-bar class="elevation-1" app clipped-right>
       <v-app-bar-nav-icon @click.stop="drawer = !drawer" />
       <v-toolbar-title @click="goHome" class="logo"
@@ -61,6 +61,7 @@ import AnnotationBrowser from "@/components/AnnotationBrowser/AnnotationBrowser.
 import BreadCrumbs from "./layout/BreadCrumbs.vue";
 import vMousetrap from "./utils/v-mousetrap";
 import { Vue, Component, Prop } from "vue-property-decorator";
+import store from "@/store";
 import toolsStore from "@/store/tool";
 
 Vue.use(vMousetrap);
@@ -76,6 +77,7 @@ Vue.use(vMousetrap);
   }
 })
 export default class App extends Vue {
+  readonly store = store;
   readonly toolsStore = toolsStore;
   drawer = false;
   snapshotPanel = false;
@@ -84,6 +86,15 @@ export default class App extends Vue {
   annotationPanel = false;
 
   lastModifiedRightPanel: string | null = null;
+
+  mousetrapBindings = [
+    {
+      bind: "t",
+      handler: () => {
+        this.store.setDrawTooltips(!this.store.drawTooltips);
+      }
+    }
+  ];
 
   fetchConfig() {
     // Fetch the list of available tool templates
--- a/src/components/AnnotationBrowser/AnnotationToggles.vue
+++ b/src/components/AnnotationBrowser/AnnotationToggles.vue
@@ -34,6 +34,23 @@
           ></v-checkbox>
         </v-list-item>
         <v-list-item>
+          <v-checkbox
+            hide-details
+            dense
+            v-model="drawTooltips"
+            label="Show Tooltips"
+          ></v-checkbox>
+        </v-list-item>
+        <v-list-item>
+          <v-checkbox
+            hide-details
+            :disabled="!drawTooltips"
+            dense
+            v-model="tooltipsOnSelected"
+            label="Show Only Tooltips Selected Annotations"
+          ></v-checkbox>
+        </v-list-item>
+        <v-list-item>
           <v-select
             v-model="annotationSelectionType"
             :items="annotationsSelectionTypeItems"
@@ -95,6 +112,22 @@ export default class AnnotationToggles extends Vue {
 
   set drawAnnotations(value: boolean) {
     this.store.setDrawAnnotations(value);
+  }
+
+  get drawTooltips() {
+    return this.store.drawTooltips;
+  }
+
+  set drawTooltips(value: boolean) {
+    this.store.setDrawTooltips(value);
+  }
+
+  get tooltipsOnSelected() {
+    return this.store.tooltipsOnSelected;
+  }
+
+  set tooltipsOnSelected(value: boolean) {
+    this.store.setTooltipsOnSelected(value);
   }
 
   get filteredDraw() {
--- a/src/components/AnnotationBrowser/AnnotationToggles.vue
+++ b/src/components/AnnotationBrowser/AnnotationToggles.vue
@@ -38,7 +38,7 @@
             hide-details
             dense
             v-model="drawTooltips"
-            label="Show Tooltips"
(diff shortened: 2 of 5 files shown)


luciemac:
Not sure we should merge the "hovering" option as it does not really work.
What do you think @finetjul?

src/components/AnnotationBrowser/AnnotationToggles.vue
```diff
@@ -33,6 +33,41 @@
...
+            label="Show Hovered Annotation Tooltip"
+          ></v-checkbox>
+        </v-list-item>
+        <v-list-item>
+          <v-checkbox
+            hide-details
+            :disabled="!drawTooltips"
+            dense
+            v-model="tooltipOnSelected"
+            label="Show Selected Annotations Tooltips"
```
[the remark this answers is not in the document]
luciemac: The same way you were doing it with the checkboxes, the list should be disabled if `!drawTooltips`

src/components/AnnotationViewer.vue
```diff
@@ -292,6 +311,82 @@ export default class AnnotationViewer extends Vue {
...
     this.annotationLayer.draw();
+    this.drawTooltips();
+  }
+
+  drawTooltips() {
+    const oldFeatures = this.textLayer.features();
+
+    if (this.shouldDrawTooltips) {
+      // Avoid checking the store for each annotation
+      const all = this.tooltipOnAll;
```
luciemac, line 322: I don't really understand the necessity of these variables, can you explain?

src/components/AnnotationViewer.vue
```diff
@@ -292,6 +311,82 @@ export default class AnnotationViewer extends Vue {
...
+      const selected = this.tooltipOnSelected;
+      const hovered = this.tooltipOnHovered;
+
+      const displayedAnnotations = this.annotationLayer
+        .annotations()
+        .map((a: any) => a.options("storedAnnotation"))
```
luciemac, line 328: Please use a better variable name: a -> annotation ?

src/components/AnnotationViewer.vue
```diff
@@ -292,6 +311,82 @@ export default class AnnotationViewer extends Vue {
...
+        .filter((a: any) => {
```
luciemac, line 329: Do we agree that in this case, "a" is what you called a "stored annotation" i.e. an annotation from the store?
If so, I guess instead of any, "a" type should be IAnnotation?

src/components/AnnotationViewer.vue
```diff
@@ -292,6 +311,82 @@ export default class AnnotationViewer extends Vue {
...
+        textBaseline: "middle",
+        color: "black",
+        textStrokeColor: "#FFF8",
+        textStrokeWidth: 2
+      };
+      this.textLayer
+        .createFeature("text")
+        .data(displayedAnnotations)
+        .position((annotation: IAnnotation) => {
+          return simpleCentroid(annotation.coordinates);
```
luciemac, line 354: Depending on the annotationType, you don't need to compute the centroid everytime.
For pointAnnotation, you can directly set position with `annotation.coordinates[0]`

src/components/AnnotationViewer.vue
```diff
@@ -292,6 +311,82 @@ export default class AnnotationViewer extends Vue {
...
+        })
+        .style({
+          text: (annotation: IAnnotation) => {
+            return "[ " + annotation.tags.join(", ") + " ]";
+          },
+          offset: { x: 0, y: 6 },
+          ...baseStyle
+        });
```
luciemac, line 382: Can't you add both texts at the same time?

src/components/AnnotationViewer.vue
```diff
@@ -292,6 +311,82 @@ export default class AnnotationViewer extends Vue {
...
+      // Centroid is computed once per new line (optimization needed?)
+      // TODO: More performance with renderThreshold https://opengeoscience.github.io/geojs/apidocs/geo.textFeature.html#.styleSpec
+      const baseStyle = {
+        fontSize: "12px",
+        fontFamily: "sans-serif",
+        textAlign: "center",
+        textBaseline: "middle",
+        color: "black",
+        textStrokeColor: "#FFF8",
+        textStrokeWidth: 2
```
luciemac, line 348: Is the text shown on top of a white/transparent rectangle? 
Because, in my opinion, it is possible that we won't be able to read information as the text font color is black, and we are working with black/white images..

src/App.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-app id="inspire">
+  <v-app v-mousetrap="mousetrapBindings" id="inspire">
```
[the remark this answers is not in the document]
bruyeret-kitware: Mousetrap makes it possible to do complicated shortcuts, you can see the possibilities [here](https://craig.is/killing/mice).
It was also already in the project.
I find it to be clearer than using a switch statement and a state, which would be reinventing the wheel.

src/components/AnnotationViewer.vue
bruyeret-kitware: They are not necessary, but because we have to use them in the loop over all annotations, I think it is more efficient to fetch them from the store once and then use them as local variables.
The performance impact is not measured though.

src/components/AnnotationViewer.vue
bruyeret-kitware: I used the same naming as in the function `drawAnnotations` above, should I rename in this function too?

luciemac: not necessarily

src/components/AnnotationViewer.vue
bruyeret-kitware: It returns an `IAnnotation` or `undefined` if there is no "storedAnnotation" option, I'll change it for `IAnnotation | undefined`.

src/components/AnnotationViewer.vue
bruyeret-kitware: It is black with a half-transparent white stroke for readability.
I didn't find any option to add a background color for text.

src/components/AnnotationViewer.vue
bruyeret-kitware: It would be redundant as this is what is done by simpleCentroid

src/components/AnnotationViewer.vue
bruyeret-kitware: No, that is impossible for now, see comment line 338 `// One text feature per line as in https://opengeoscience.github.io/geojs/tutorials/text/`.

finetjul:
> Not sure we should merge the "hovering" option as it does not really work. What do you think @finetjul?

Ok to remove the "hovering" from this PR, we can discuss it on Friday with Arjun.

src/components/AnnotationViewer.vue
bruyeret-kitware: Yes, my bad, I was thinking it was checking.
Do you think it is worth it to check for pointAnnotation?
We could put this check in simpleCentroid as it would factorize the code and make it clearer.

luciemac: yes, it avoids doing some computations when it is not needed.
I don't know if it would be clearer, because do we need to compute a centroid for a point, I don't know.
I think, both solution are ok.

luciemac:
LGTM

PR #283, opened 2022-09-28
Added multi-image tiff support

bruyeret-kitware:
Added stride options per channel when needed

### DIFF
--- a/src/views/dataset/MultiSourceConfiguration.vue
+++ b/src/views/dataset/MultiSourceConfiguration.vue
@@ -29,13 +29,15 @@
             </template>
           </v-combobox>
         </v-col>
-        <v-col>
+        <v-col v-if="!areStridesSetFromFile && maxFramesPerItem > 1">
           <v-text-field
-            :rules="[input => input === '' || parseInt(input) > 0]"
+            prefix="Stride"
+            :rules="[
+              input => input === null || input === '' || parseInt(input) > 0
+            ]"
             :disabled="
-              (assignments[assignment] &&
-                assignments[assignment].value.source === 'file') ||
-                maxFramesPerItem <= 1
+              assignments[assignment] &&
+                assignments[assignment].value.source === 'file'
             "
             v-model="strides[assignment]"
             hide-details
@@ -46,10 +48,16 @@
         <v-col>
           <v-btn
             :disabled="
-              !assignments[assignment] ||
-                assignments[assignment].value.source === 'file'
+              (!assignments[assignment] &&
+                (typeof strides[assignment] !== 'string' ||
+                  strides[assignment] === '')) ||
+                (assignments[assignment] &&
+                  assignments[assignment].value.source === 'file')
             "
-            @click="assignments[assignment] = null"
+            @click="
+              assignments[assignment] = null;
+              strides[assignment] = null;
+            "
             >Clear</v-btn
           >
         </v-col>
@@ -155,6 +163,7 @@ export default class NewDataset extends Vue {
     T: null,
     C: null
   };
+  areStridesSetFromFile = false;
 
   collectedMetadata: {
     metadata: {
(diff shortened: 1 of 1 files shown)


src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -135,6 +157,13 @@ export default class NewDataset extends Vue {
     T: null,
     C: null
   };
+  strides: { [dimension: string]: number | string | null } = {
+    XY: null,
```
luciemac, line 161: Why do you need to support 'string' type when it used to support only number?

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -135,6 +157,13 @@ export default class NewDataset extends Vue {
...
+    Z: null,
+    T: null,
+    C: null
+  };
+  areStridesSetFromFile = false;
```
luciemac, line 166: missing a type definition?

src/views/dataset/MultiSourceConfiguration.vue
bruyeret-kitware: The input for the stride is a `v-text-field` and its v-model uses a string instead of a number.

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -222,14 +248,19 @@ export default class NewDataset extends Vue {
...
-        this.strides = {
-          C: tile.IndexStride?.IndexC || null,
-          T: tile.IndexStride?.IndexT || null,
-          XY: tile.IndexStride?.IndexZ || null,
-          Z: tile.IndexStride?.IndexZ || null
-        };
+      const frames: number = tile.frames?.length || 1;
+      this.maxFramesPerItem = Math.max(this.maxFramesPerItem, frames);
+      if (tile.IndexStride) {
+        this.assignmentNames.forEach(key => {
```
luciemac, line 257: key -> dimension?

src/views/dataset/MultiSourceConfiguration.vue
bruyeret-kitware: It could be explicitly defined as `boolean` but setting it to `false` implicitly defines it as `boolean`.
Should I edit the declaration?

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -275,6 +306,18 @@ export default class NewDataset extends Vue {
       ? this.collectedMetadata.filesInfo
       : null;
 
+    const framesAsAxes: { [dim: string]: number } = {};
+    Object.entries(this.strides).forEach(([key, value]) => {
```
luciemac, line 310: key -> dimension ?
value -> stride?

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -275,6 +306,18 @@ export default class NewDataset extends Vue {
...
+      if (value) {
+        if (typeof value === "string") {
```
luciemac, line 312: I think this conversion should be done before being stored, what do you think?

src/views/dataset/MultiSourceConfiguration.vue
luciemac: I think you can specify the input type of the v-text-field https://vuetifyjs.com/en/api/v-text-field/#props-type

src/views/dataset/MultiSourceConfiguration.vue
bruyeret-kitware: It would be better so that we only need to support `null` and `number`
We need a setter for that because `v-text-field` edits `strides`
But I didn't manage to create a setter for `strides`, because it is an object

src/views/dataset/MultiSourceConfiguration.vue
luciemac: I think it is better to define types!

luciemac:
LGTM

Issue #129, opened 2021-06-14
Add unroll for layers


arjunrajlab:
Add an "unroll" feature for layers:
![image](https://user-images.githubusercontent.com/13998824/121904451-31d58f80-ccf7-11eb-97f7-6ba5f8158cc4.png)

manthey:
Per channel or per layer?  If per layer, I assume each layer still retain its specified coloring, channel, offsets, etc.

arjunrajlab:
Good question. I think per layer, so it would retain coloring, channel, offsets, etc.

arjunrajlab:
Updated my original issue to say layer instead of channel.

finetjul:
@zjniu can you please give it a try ?

PR #288, opened 2022-10-06
Unroll layers.

manthey:
This also split a large function into several functions in the ImageViewer's draw routine.  Multiple annotation layers are created (one per unrolled map).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5370,6 +5370,16 @@
       "integrity": "sha512-jDctJ/IVQbZoJykoeHbhXpOlNBqGNcwXJKJog42E5HDPUwQTSdjCHdihjj0DlnheQ7blbT6dHOafNAiS8ooQKA==",
       "dev": true
     },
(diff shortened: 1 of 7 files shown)


manthey:
Resolves #129.

arjunrajlab:
Couple artifacts I noticed: one is that with unroll, just one channel has the smooth scrubbing (in this case red), while the others do not:

https://user-images.githubusercontent.com/13998824/194652989-4b8fa613-d764-4b6d-ac6d-65f4c8c4ba67.mov

After going back to "multiple" the issue persists, along with some strange artifact:

https://user-images.githubusercontent.com/13998824/194653055-dd3928d5-1258-4153-9d8d-584f257212e1.mov

After refreshing the page, it was fine again.

manthey:
Yes.  This is actually the bug I was hunting down.  I'll have a fix today.

arjunrajlab:
Fix looks good!

PR #289, opened 2022-10-07
Interface thoughts - first features

bruyeret-kitware:
- OVERALL FIXES: 1 and 2
- FILE NAVIGATOR: 3 to 7 and some elements of 8
- UPLOADING DATA: 11


src/views/dataset/DatasetInfo.vue
```diff
@@ -285,21 +285,21 @@ export default class DatasetInfo extends Vue {
       return;
     }
 
-    const date = new Date();
     try {
+      const dataset = this.store.dataset;
+      if (dataset === null) {
+        throw new Error("dataset was null when it shouldn't be");
```
luciemac, line 291: I guess you should put a simpler error message: "Dataset not set"

bruyeret-kitware: That some code I only moved around
I'll change both error message for "Dataset not set" and "Config not set"

src/App.vue
```diff
@@ -1,26 +1,44 @@
...
-      <v-btn @click.stop="toggleRightPanel('snapshotPanel')" id="snapshotButton"
-        >Snapshot</v-btn
-      >
-      <user-menu />
-      <v-btn @click.stop="toggleRightPanel('annotationPanel')"
-        >Browse Annotations</v-btn
+      <v-btn
+        color="primary"
+        class="ml-4"
+        :to="{ name: 'newdataset' }"
```
luciemac, line 12: newdataset -> newDataset?

luciemac:
LGTM then

src/App.vue
bruyeret-kitware: This is the name of the route I took from src/views/dataset/index.ts

```
path: "/new",
name: "newdataset",
component: NewDataset,
```

```
path: "/import",
name: "importdataset",
component: ImportDataset,
```

Should I refactor "newdataset" and "importdataset" ?

luciemac: Ohhhh I see.
No you don't need to change anything!

PR #300, opened 2022-10-18
Annotation Snapping feature

bruyeret:
Annotation snapping from Adrien's branch
The code has only been adapted to work with current code
Features:
- ellipseToDot and blobToDot (same algorithm)
- blobToBlob
ITK workers fail sometimes, it needs to be investigated


arjunrajlab:
Having some issues:
Snap blob to blob goes to a point:
image
image

Also have found that ellipse to point and blob to point both don't give any annotations.

arjunrajlab:
Actually, snap ellipse to point works, but it is not the intended behavior. I think the ellipse to point should take the maximum intensity within a certain radius for the point. Like, you click on a place, then it makes a point annotation at the maximum intensity within that radius. So it's click a point, then gives a point. I should say that I thought it was not working at first because it didn't work for the first several clicks.

arjunrajlab:
Okay, so the snap blob to blob actually does work. I was mistakenly setting it to "point" as annotation type :). My fault. But the functionality is there. Sometimes it seems to not build an annotation, not sure why.

arjunrajlab:
It also seems a little bit slow compared to previously, but who knows.

arjunrajlab:
OK, so I think I know why it works sometimes and not others (blob to blob). If I make the initial blob with a polygon that has a large distance from end to end that it must automatically close, then it doesn't work. If I "close" the blob by ending my drawing fairly close to the beginning of my drawing, then it seems to work. Can we change the default to work even if the ends of the drawing are far apart?

bruyeret:
The problem of the BlobToBlob not creating an annotation comes from ITK which doesn't find a contour.
I'm not sure yet why it doesn't, but it could be something "unfixable".
Do you remember the tool behaving like this sometimes?

For the ellipse to point, how should the user specify the radius?

arjunrajlab:
Wow, that is really weird about ITK not finding a contour! Can the front end maybe "fill it in" before sending to ITK? If it is not easily fixed, then whatever, we can wait until later to figure out what to do.

arjunrajlab:
For ellipse to point, I think it would actually be best to have the radius scale to the interface, sort of like how the circle for the point annotation works now (if you zoom in, the point annotation remains the same size on the interface). That way, if you zoom in, it allows you to be more fine grained. Does that seem easy? If not, then I think a radius of 2 pixels is probably fine for most purposes.

src/components/AnnotationViewer.vue
```diff
@@ -830,6 +810,73 @@ export default class AnnotationViewer extends Vue {
...
+  }
+
+  async addAnnotationFromSnapping(annotation: any) {
+    if (!annotation || this.maps.length !== 1) {
+      return;
+    }
+    const mapentry = this.maps[0];
+    const coordinates = annotation.coordinates();
+    this.annotationLayer.removeAnnotation(annotation);
+    const tool = this.selectedTool;
```
luciemac, line 838: Is it necessary?

src/components/AnnotationViewer.vue
```diff
@@ -830,6 +810,73 @@ export default class AnnotationViewer extends Vue {
...
+    if (!tool) {
+      return;
+    }
+    const location = tool.values.annotation.coordinateAssignments;
+    if (!location) {
+      logError("Invalid snapping tool, annotation was not configured properly");
+      return;
+    }
+    const layer = location.layer;
```
luciemac, line 847: FYI, you can also have:
const {layer} = location;

src/components/AnnotationViewer.vue
```diff
@@ -839,7 +886,87 @@ export default class AnnotationViewer extends Vue {
...
+  updateCursorAnnotation(evt?: any) {
+    if (
+      !this.selectedTool ||
+      !this.cursorAnnotation ||
+      !this.selectedToolRadius
+    ) {
+      return false;
+    }
+    const map = this.maps[0].map;
+    const basePositionGCS = evt?.mapgcs ? evt.mapgcs : this.lastCursorPosition;
```
luciemac, line 906: Check if this.maps[0] is defined first

src/components/AnnotationViewer.vue
```diff
@@ -858,6 +985,14 @@ export default class AnnotationViewer extends Vue {
         const annotation = this.selectedTool.values.annotation;
         this.annotationLayer.mode(annotation?.shape);
         break;
+      case "snap":
+        if (this.selectedTool.values.snapTo.value === "circleToDot") {
```
luciemac, line 989: Isn't it an enum or something like that to ensure "circleToDot" typo?

src/utils/itk.ts
```diff
@@ -0,0 +1,216 @@
...
+          reject();
+        } else {
+          const result = JSON.parse(outputs[0].data);
+          webWorker.terminate();
+          resolve(result);
+        }
+      })
+      .catch((error: any) => {
+        logError(`Failed to run ITK pipeline:
+${error.message}
```
luciemac, line 77: indentation

src/components/AnnotationViewer.vue
bruyeret: No, I'll remove it

src/components/AnnotationViewer.vue
bruyeret: Thanks!

src/components/AnnotationViewer.vue
bruyeret: It corresponds to a value written in `public\config\templates.json` (line 46) and I can't think of a way of doing things better
In the rest of the code source these values are also checked against simple strings

src/utils/itk.ts
bruyeret: This is a literal string, if I add indentation, it will be included in the string
Do you mean that the error message also needs indentation? or I should reformat it?

luciemac:
LGTM

finetjul:
> @luciemac the code is ready for a review but the CI fails because it misses wasm files generated by the `build-workers-***` command. I think that putting the built files in the repo is not a good idea (for example if the source files are modified but the built files are not pushed). Can we add the `build-workers-***` command to the CI?

Je pense qu'il faut pousser les fichiers wasm dans le repo. WDYT @luciemac  ?

luciemac:
@finetjul indeed, it will be necessary to push .wasm files as long as we are not sure how complicated it would be to rebuilt them.

bruyeret:
@luciemac ready to merge if you are

luciemac:
yes LGTM

PR #301, opened 2022-10-18
More UI changes

bruyeret:
- Small fixes, refactor and UI changes (15, 16)
- Replace stride option with a new variable (14)
- Added Z Max Merge switch to each channel (18)

### DIFF
--- a/src/components/DisplayLayer.vue
+++ b/src/components/DisplayLayer.vue
@@ -187,7 +187,7 @@ export default class DisplayLayer extends Vue {
     if (this.isZMaxMerge === value) {
       return;
     }
-    let newVal = {};
+    let newVal;
     if (value) {
       newVal = {
         type: "max-merge",
--- a/src/components/DisplayLayer.vue
+++ b/src/components/DisplayLayer.vue
@@ -7,6 +7,23 @@
         @click.native.stop
         @mousedown.native.stop
         @mouseup.native.stop
+        v-mousetrap="[
+          {
+            bind: zMaxMergeBinding(index),
+            handler: () => (isZMaxMerge = !isZMaxMerge)
+          }
+        ]"
+        class="toggleButton"
+        v-model="isZMaxMerge"
+        :title="`Toggle Z Max Merge (Hotkey ${zMaxMergeBinding(index)})`"
+        :displayed="displayZ"
+        dense
+        hide-details
+      />
+      <v-switch
+        @click.native.stop
+        @mousedown.native.stop
+        @mouseup.native.stop
         class="toggleButton"
         v-model="visible"
         :title="`Toggle Visibility (Hotkey ${index + 1})`"
@@ -111,8 +128,8 @@
 </template>
 
 <script lang="ts">
-import { Vue, Component, Prop, Emit } from "vue-property-decorator";
-import { IDisplayLayer, IContrast } from "../store/model";
+import { Vue, Component, Prop, Emit, Watch } from "vue-property-decorator";
+import { IDisplayLayer, IContrast, IDisplaySlice } from "../store/model";
 import DisplaySlice from "./DisplaySlice.vue";
 import ContrastHistogram from "./ContrastHistogram.vue";
 import store from "../store";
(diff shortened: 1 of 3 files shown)


arjunrajlab:
I am having issues when uploading e.g. the DDX58_AXL_EGFR_well2.nd2 file.

image

It seems to find some extra "Frames per image variable". Also, it is somehow trying to include time, even though that doesn't exist. Overall, not sure what changed, it was actually working quite well before.

arjunrajlab:
The z max merge slider works great, though!

arjunrajlab:
OK, so the DDX58…… dataset works now. But this dataset doesn't:
https://www.dropbox.com/sh/ota13lyrlzorul8/AADF9GzR4FeGikoelspk7dNIa?dl=0

image

image

Gets to the above screen, but seems to be recognizing too many variables and won't let you click "Submit".

arjunrajlab:
Verified that multiple datasets work now with import, thank you @bruyeret!! I think it is ready to merge if okay with you @finetjul

arjunrajlab:
Minor point: Can you put a little text saying "Z max-merge" and "Channel on/off" next to those sliders? Just so that people know what they do. Thanks! @bruyeret

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -102,29 +88,36 @@ export default class NewDataset extends Vue {
     return this.$route.params.id;
   }
 
+  // Create a short string describing the array with some elements from the array
+  // For example: ["foo", "bar", "foobar", "barfoo"] => "foo, bar, foobar..."
+  arrayToExampleValues(arr: string[]) {
+    let values = arr.slice(0, 3).join(", ");
+    if (arr.length > 3) {
+      values = String.prototype.concat(values, "...");
```
finetjul, line 96: -> `values += '...'`

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -102,29 +88,36 @@ export default class NewDataset extends Vue {
...
+    }
+    return values;
```
finetjul, line 98: I think it won't work if your items are VERY long.

Your limit could be instead a number of characters:
``` 
function truncate(str, n){
  return (str.length > n) ? str.slice(0, n-1) + '&hellip;' : str;
};
```

Or even better, [handle it at CSS level](https://stackoverflow.com/a/9792134).

src/components/DisplayLayer.vue
```diff
@@ -153,6 +170,46 @@ export default class DisplayLayer extends Vue {
...
+
+  get isZMaxMerge() {
+    return this.zSlice.type === "max-merge";
+  }
+
+  set isZMaxMerge(value: boolean) {
+    if (this.isZMaxMerge === value) {
+      return;
+    }
+    let newVal = {};
```
finetjul, line 190: no need to initialize with an empty object. `newVal` will be overwritten anyway

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -348,10 +338,37 @@ export default class NewDataset extends Vue {
...
+  areDimensionsSetToDefault() {
+    return Object.keys(this.dimensionNames).every(
+      dim =>
+        this.getDefaultAssignmentItem(dim)?.value ===
+        this.assignments[dim]?.value
+    );
+  }
+
+  canSubmit() {
+    const filledAssignments = Object.entries(this.assignments).filter(
```
finetjul, line 368: `Object.entries` -> `Object.values`

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -348,10 +338,37 @@ export default class NewDataset extends Vue {
...
+      ([_, assignment]) => assignment !== null
+    ).length;
```
finetjul, line 370: technically it is more a `.reduce()` than a `.filter()`

finetjul (changes requested):

src/views/dataset/MultiSourceConfiguration.vue
bruyeret: I prefer doing it at JS level to avoid having a big chunk of text almost unused in the DOM.

bruyeret:
For the "Z max-merge" and "Channel on/off" labels, it could be a bit cluttered if they are next to the switches.
I could put these labels as tooltips, what do you think @arjunrajlab ?

src/views/dataset/MultiSourceConfiguration.vue
finetjul: By doing it as JS level, you are not very flexible when it comes to dynamic layout. What if user zooms the ui (CTRL+'+') or if the UI is visible from a mobile device. If tomorrow you change the "language" or the font size, you will also have to modify the limit in the number of characters.

arjunrajlab:
> For the "Z max-merge" and "Channel on/off" labels, it could be a bit cluttered if they are next to the switches. I could put these labels as tooltips, what do you think @arjunrajlab ?

Yeah, let's try with the text next to the switches… just makes it more discoverable. If it's too weird, we can go to tooltips.

arjunrajlab:
I like those labels a lot!

src/components/DisplayLayer.vue
```diff
@@ -153,6 +178,46 @@ export default class DisplayLayer extends Vue {
...
+    }
+    let newVal;
+    if (value) {
+      newVal = {
+        type: "max-merge",
+        value: null
+      };
+    } else {
+      newVal = this.alternativeZSlice;
+    }
```
luciemac, line 206:
```
const zNewValue = value
?  {
        type: "max-merge",
        value: null
      }
 : this.alternativeZSlice;
```

bruyeret: Good point, I didn't see it that way !

luciemac:
LGTM

PR #302, opened 2022-10-26
Feature: add tags to selected annotations

bruyeret:
Open a dialog when wlicking "TAG SELECTED" in the action menu of the browse annotation panel.
Remove the embedding of IAnnotation in geojs annotations ("storedAnnotation" option) which caused bugs when updating the IAnnotation.

image

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -334,12 +334,19 @@ export default class AnnotationViewer extends Vue {
   }
 
   drawTooltips() {
-    const oldFeatures = this.textLayer.features();
+    this.textLayer.features().forEach((feature: any) => {
+      this.textLayer.deleteFeature(feature);
+    });
 
     if (this.showTooltips) {
-      const annotationsToTooltip = this.annotationLayer
+      const displayedAnnotations: IAnnotation[] = this.annotationLayer
         .annotations()
-        .map((annotation: any) => annotation.options("storedAnnotation"))
+        .map((geojsAnnotation: any) => {
+          return this.annotations.find(
+            iAnnotation =>
+              iAnnotation.id === geojsAnnotation.options("girderId")
+          );
+        })
         .filter((annotation: IAnnotation | undefined) => {
           return (
             annotation &&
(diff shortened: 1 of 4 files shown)


arjunrajlab:
Just tested, works well, looks good to me for merging!

luciemac:
LGTM

src/store/annotation.ts
```diff
@@ -282,6 +282,41 @@ export class Annotations extends VuexModule {
     this.setSelected([]);
   }
 
+  @Action
+  public async tagAnnotationIds({annotationIds, tags}: {annotationIds: string[], tags: string[]}) {
+    sync.setSaving(true);
+    await Promise.all(
+      annotationIds
+        .map(annotationId => this.annotations.findIndex((annotation: IAnnotation) => annotation.id === annotationId))
```
luciemac, line 290: Don't you think it could be useful to have a helper method to get annotations from annotationsIds ?

bruyeret: In this case I need the index to use it later but that's something that could be helpful in other places in the code
But I think it should be done in another pull request with eventually other reformatting

src/components/AnnotationViewer.vue
```diff
@@ -334,12 +334,19 @@ export default class AnnotationViewer extends Vue {
...
+    this.textLayer.features().forEach((feature: any) => {
+      this.textLayer.deleteFeature(feature);
+    });
 
     if (this.showTooltips) {
-      const annotationsToTooltip = this.annotationLayer
+      const displayedAnnotations: IAnnotation[] = this.annotationLayer
         .annotations()
-        .map((annotation: any) => annotation.options("storedAnnotation"))
+        .map((geojsAnnotation: any) => {
```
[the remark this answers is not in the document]
bruyeret: It is a [geojs annotation](https://opengeoscience.github.io/geojs/apidocs/geo.annotation.html) but I didn't find a ts type for it. This `.map()` finds the corresponding IAnnotation in the store and the filter after it receives a `IAnnotation | undefined`.

PR #306, opened 2022-10-31
Put layer labels only on the first row

bruyeret:
Labels of switches "Channel on/off" and "Z max-merge" moved to first row only.


luciemac:
LGTM

luciemac (approved):

PR #309, opened 2022-11-04
Fix an issue when comparing histogram image lists.

### LINKED ISSUE #307: Issue with time offset causing a crash
There is an option, useful for time-lapse image visualization, for an offset in the time:

image

image

Upon initially loading this dataset and visualizing with the offset layer, when you get to T=4 (so the red layer here would be T=5, which doesn't exist), the program crashes. If you reload the page, it works, though. I suspect it is trying to reference, on initial load, a frame which doesn't exist (T=5). 

Here's what it should look like, which does work properly for all T < max(T):
image

@manthey @finetjul

manthey:
Specifically, Vue sometimes makes the image list an observable and then it doesn't compare with the non-observable list of the same items. Instead, compare the entries in the list.

This fixes #307.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -955,7 +955,11 @@ export class Main extends VuexModule {
         }
         return null;
       };
-      if (images !== layer._histogram.images) {
+      if (
+        !layer._histogram.images ||
+        images.length !== layer._histogram.images.length ||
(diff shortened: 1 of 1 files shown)


arjunrajlab:
Worked for me, fixed bug, thanks!

Issue #303, opened 2022-10-28
Annotation tool creation UI streamlining


arjunrajlab:
Thoughts for UI for streamlining the annotation tool creation:

https://www.figma.com/file/pQthSFEYk61A1Qq6HCxbCT/Annotation-creation-UI%2FUX?node-id=0%3A1

arjunrajlab:
I just updated the Figma slightly, let me know if you have any questions @bruyeret!

arjunrajlab:
For the "Advanced" options, I think that it would be better to have a dropdown instead of a checkbox. Like, one of these:

<img width="47" alt="image" src="https://user-images.githubusercontent.com/13998824/201667659-d0538e9a-e492-476e-89c3-1894932c8a65.png">

Also, I think even if we cannot do the two list thing, the current UI is not very discoverable or intuitive. Let me think what might be easiest to implement but still discoverable.

arjunrajlab:
Also, here, it should show the actual job-date tag next to the checkbox so that the user sees what the tag will be:
<img width="483" alt="image" src="https://user-images.githubusercontent.com/13998824/201674036-54ffbbc4-7889-44c9-a336-852a2b54cbd7.png">

bruyeret:
The UI is defined in a JSON file, and in order to add some logic to the UI (add and format current date), we have to create special UI elements (like the "Annotation Configuration" or the "Connect to nearest annotation with tag").
It can be done but it will take a lot more time than adding an example tag like: `Add job date tag (e.g. "my-image job 1985-10-26 21:00:00 UTC-5")`
What do you prefer?

arjunrajlab:
Ah, okay, I see. Let's just put in the example tag then. :) More UI feedback in a bit later today, sorry, baby has me busy :).

bruyeret:
Ok that's done!
Congratulations on the baby! Don't worry for the UI feedback, I still have issues to work on :D

arjunrajlab:
OK, here's my UI feedback:
https://www.figma.com/file/pQthSFEYk61A1Qq6HCxbCT/Annotation-creation-UI%2FUX?node-id=0%3A1
<img width="288" alt="image" src="https://user-images.githubusercontent.com/13998824/202412721-ae19aed8-ec5f-4ca8-95ad-221a50696243.png">

I think this significantly streamlines the UI and is more what I was envisioning. Let me know if you have questions!

PR #313, opened 2022-11-10
Annotation tool creation UI streamlining

bruyeret:
UI streamlining of [issue 303](https://github.com/Kitware/UPennContrast/issues/303).
All the features from [the document](https://www.figma.com/file/pQthSFEYk61A1Qq6HCxbCT/Annotation-creation-UI%2FUX?node-id=0%3A1) have been implemented with a difference:
The pair of lists was too long to implement, instead there is a two step creation :
- choose a tool type
image

- configure the tool
image

Tool description, connect to nearest and Z and Time assignment have not been hidden/removed.
Instead, an advanced mode has been added which reveals these options and can be discovered by the user who may also ignore these options.


luciemac (approved):

luciemac:
LGTM

bruyeret:
The UI is almost identical to what you proposed @arjunrajlab 
What do you think about it?
Let me know if I can ping Lucie for a review

arjunrajlab:
Looks great! Only thing is that the "snap-to" annotations don't seem to be working anymore. One thing: for the automatic naming, perhaps if it is a manual/snap-to tool, the name would be e.g. "nucleus YFP blob (manual)" or "nucleus YFP blob (snap-to)" or something like that.

arjunrajlab:
This looks good to me! Issues:
1. The "snap-to" annotations don't appear to be working for me anymore.
2. This interface is a little bit odd… it's a text field that you have to click on, but there is no indication that is a dropdown menu. Then it shows a pop-up menu to the right. It's okay, just quite non-standard looking to me.
image
3. It would be good if the automatic naming incorporated "snap-to" or "manual". E.g.: "nucleus YFP blob (manual)" or "nucleus YFP blob (snap-to)".

bruyeret:
I think that all the mentionned issues have been fixed, do you prefer a this selection UI?
I didn't see that it was possible to create headers and dividers in list selection, the code is also a bit clearer.

arjunrajlab:
This update looks good to me! I tested it out and it seems to work great. Thanks!

luciemac:
LGTM

Issue #304, opened 2022-10-30
Connections Not Displaying


zjniu:
Connect to nearest functionality on manual annotations is currently broken. Either no connections are being made, or they are not displayed properly.

arjunrajlab:
Can confirm that this is an issue, see Slack for screenshots. I wonder if @bruyeret maybe it's on the back end and needs to pull the latest girder/large_image docker images? We have found that we need to run docker with --no-cache a couple times to get it to fully update.

PR #314, opened 2022-11-15
Remove z coordinate from IGeoJSPoint

bruyeret:
Z coordinate of IGeoJSPoint was used inconsistently, being often undefined and causing bugs.
Closes [#304](https://github.com/Kitware/UPennContrast/issues/304)

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
@@ -3,7 +3,8 @@ import numpy as np
 
 def pointToPointDistance(coord1, coord2):
     """Get the distance between two points using points coordinates.
-    Points coordinates are represented by {x: x_coord, y: y_coord, z: z_coord}.
+    If both points have z coordinate, get the XYZ distance, else get the XY distance
+    Points coordinates are represented by {x: x_coord, y: y_coord, z?: z_coord}.
 
     Args:
         coord1 (dict): Coordinates of point 1
@@ -12,18 +13,20 @@ def pointToPointDistance(coord1, coord2):
     Returns:
         Number: Squared distance between point1 and point2
     """
-    return math.sqrt(
-      math.pow(coord1["x"] - coord2["x"], 2) + math.pow(coord1["y"] - coord2["y"], 2)
-    )
+    squareDist = math.pow(coord1["x"] - coord2["x"], 2) + math.pow(coord1["y"] - coord2["y"], 2)
+    if ("z" in coord1 and "z" in coord2):
+      squareDist += math.pow(coord1["z"] - coord2["z"], 2)
+    return math.sqrt(squareDist)
   
 
 def simpleCentroid(listCoordinates):
     """Compute the simple centroid of a polygon using its list of coordinates.
     The centroid corresponds to the barycenter of the polygon.
+    Compute centroid for z coordinate iff all points have z
 
     Args:
         listCoordinates (dict[]): List of point coordinates.
-          Point coordinates are represented by {x: x_coord, y: y_coord, z: z_coord}
+          Point coordinates are represented by {x: x_coord, y: y_coord, z?: z_coord}
 
     Returns:
         dict: Coordinates of the centroid
@@ -32,10 +35,18 @@ def simpleCentroid(listCoordinates):
     x = np.sum([coord["x"] for coord in listCoordinates]) / nbCoordinates
     y = np.sum([coord["y"] for coord in listCoordinates]) / nbCoordinates
 
-    return {
-      "x": x,
-      "y": y
-    }
+    if all(["z" in coord for coord in listCoordinates]):
+      z = np.sum([coord["z"] for coord in listCoordinates]) / nbCoordinates
+      return {
+        "x": x,
+        "y": y,
+        "z": z
+      }
+    else:
+      return {
+        "x": x,
+        "y": y
+      }
 
 def isAPoint(annotation):
   return annotation["shape"] == 'point'
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
@@ -13,7 +13,7 @@ def pointToPointDistance(coord1, coord2):
         Number: Squared distance between point1 and point2
     """
     return math.sqrt(
-      math.pow(coord1["x"] - coord2["x"], 2) + math.pow(coord1["y"] - coord2["y"], 2) + math.pow(coord1["z"] - coord2["z"], 2)
+      math.pow(coord1["x"] - coord2["x"], 2) + math.pow(coord1["y"] - coord2["y"], 2)
     )
   
 
@@ -31,12 +31,10 @@ def simpleCentroid(listCoordinates):
     nbCoordinates = len(listCoordinates)
     x = np.sum([coord["x"] for coord in listCoordinates]) / nbCoordinates
     y = np.sum([coord["y"] for coord in listCoordinates]) / nbCoordinates
-    z = np.sum([coord["z"] for coord in listCoordinates]) / nbCoordinates
 
     return {
       "x": x,
-      "y": y,
-      "z": z
+      "y": y
     }
 
 def isAPoint(annotation):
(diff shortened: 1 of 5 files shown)


devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
```diff
@@ -13,7 +13,7 @@ def pointToPointDistance(coord1, coord2):
         Number: Squared distance between point1 and point2
```
finetjul, line 13: the documentation says that a z coord is expected, code must be consistent with documentation

finetjul (commented):
Please write "Closes #304" instead of `[issue 304](https://github.com/Kitware/UPennContrast/issues/304)`

I do not think removing `z` is a good idea. Some alternatives: 
 - make sure it is never `undefined`
 - if undefined, consider it to be 0
 - check in functions if you have xy or xyz

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
bruyeret: In the functions simpleCentroid and pointToPointDistance, the z coordinate should be ignored, as it has no meaning, I forgot to update the doc.
It was declared because GeoJS sometimes output a z coordinate (and sometimes not, see [GeoJS doc](https://opengeoscience.github.io/geojs/apidocs/geo.html#.geoPosition)).
But in our case it corresponds to nothing, not even the z coordinate of the tile or the channel.
I think it is error prone to keep a variable that shouldn't be used in the code. This is what caused #304.
Whe could eventually mention it in the interface declaration :
```
export interface IGeoJSPoint {
  x: number;
  y: number;
  z?: number; // Artifact from GeoJS, unused
}
```

finetjul: > Please write "Closes #304" instead of `[issue 304](https://github.com/Kitware/UPennContrast/issues/304)`
> 
> I do not think removing `z` is a good idea. Some alternatives:
> 
> * make sure it is never `undefined`
> * if undefined, consider it to be 0
> * check in functions if you have xy or xyz

@manthey WDYT ?

luciemac: Actually we store in the database a 'Location' entry, for each annotation, defines as {"XY": n, "Z": m}.
When looking for the _closestAnnotation_, we filter annotations on their locations, so a "Z" value is already taken into account. So I don't think it is an issue to remove it from distance computation

arjunrajlab:
Actually, I wonder if we should leave Z in the computation, perhaps just with delta Z = 0. The reason is that in the future, we may want to compute distances in 3D. If we leave it in, then perhaps that functionality would be easier to implement. But not a big deal, can remove if it makes it significantly better/easier to maintain.

bruyeret:
Do you know if it would be a Z per annotation corresponding to the slice, or if an annotation (for example a blob) could be scattered on the Z axis or inbetween two Z slices?

bruyeret:
I think we could keep room for it without making things too complex.
We don't know exactly how it will be used later so we don't know its unit.
- make z optional for now, and send undefined z to girder as we don't use it yet (otherwise it is sometimes 0)
- `simpleCentroid` doesn't depend on the unit of z, we can take it into account when defined
- `pointToPointDistance` would rely on z being homogeneous to x and y, so let's rename it `pointToPointXYDistance`
WDYT @finetjul?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
```diff
@@ -12,32 +13,40 @@ def pointToPointDistance(coord1, coord2):
...
+    squareDist = math.pow(coord1["x"] - coord2["x"], 2) + math.pow(coord1["y"] - coord2["y"], 2)
+    if ("z" in coord1 and "z" in coord2):
+      squareDist += math.pow(coord1["z"] - coord2["z"], 2)
+    return math.sqrt(squareDist)
   
 
 def simpleCentroid(listCoordinates):
     """Compute the simple centroid of a polygon using its list of coordinates.
     The centroid corresponds to the barycenter of the polygon.
+    Compute centroid for z coordinate iff all points have z
```
finetjul, line 25: iff -> if

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
```diff
@@ -12,32 +13,40 @@ def pointToPointDistance(coord1, coord2):
...
+      return {
+        "x": x,
+        "y": y,
+        "z": z
+      }
+    else:
+      return {
+        "x": x,
+        "y": y
+      }
```
finetjul, line 49: I find the following more readable (factorized):
```
centroid = {"x":x, "y": y}
if (all(...)):
  z = ...
  centroid.z = z
return centroid;
```

src/utils/annotation.ts
```diff
@@ -91,17 +91,27 @@ export function simpleCentroid(coordinates: IGeoJSPoint[]): IGeoJSPoint {
   if (coordinates.length === 1) {
     return coordinates[0];
   }
-  const sums: IGeoJSPoint = { x: 0, y: 0, z: 0 };
+  const sums = { x: 0, y: 0, z: 0 };
+  const allZ = !coordinates.some(coord => coord.z === undefined);
```
finetjul, line 95: `.some` -> `.every` ?

finetjul: `allZ` -> `hasZ`

src/utils/annotation.ts
```diff
@@ -91,17 +91,27 @@ export function simpleCentroid(coordinates: IGeoJSPoint[]): IGeoJSPoint {
...
   coordinates.forEach(({ x, y, z }) => {
     sums.x += x;
     sums.y += y;
-    sums.z += z;
+    if (z !== undefined) {
```
finetjul, line 99: test instead `allZ` ?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
bruyeret: "iff" means "if and only if", should I replace it with "if and only if"?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
```diff
@@ -12,32 +13,40 @@ def pointToPointDistance(coord1, coord2):
...
     x = np.sum([coord["x"] for coord in listCoordinates]) / nbCoordinates
     y = np.sum([coord["y"] for coord in listCoordinates]) / nbCoordinates
-    z = np.sum([coord["z"] for coord in listCoordinates]) / nbCoordinates
 
-    return {
-      "x": x,
-      "y": y,
-      "z": z
-    }
+    if all(["z" in coord for coord in listCoordinates]):
```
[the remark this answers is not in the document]
bruyeret: It could lead to errors at runtime (error 500)

src/utils/annotation.ts
bruyeret: It would work and that's what I initially did, but typescript then tells me for z that `Object is possibly 'undefined'.ts (2532)`
I don't know how to tell typescript that it is not undefined

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
finetjul: good to know.

src/utils/annotation.ts
finetjul: I believe there are special comments that force typescript to ignore some cases, no ?

bruyeret: Indeed, I found `// @ts-ignore` comment to ignore a line

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/connections.py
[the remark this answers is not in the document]
finetjul: let's leave it as such, it will be fast when there is no Z.

finetjul:
LGTM

bruyeret:
It can be merged if you think it's ok @luciemac

luciemac:
LGTM

Issue #310, opened 2022-11-04
Annotations not showing with time offset


arjunrajlab:
In the case where we have two layers showing two different times, one current time, one next time:
<img width="534" alt="image" src="https://user-images.githubusercontent.com/13998824/200030569-19a65955-4af0-4c21-9590-c2bf64a07e73.png">
Annotations don't show up for the offset layer (red). That is, at T=1, annotations on T=2 do not show up in red.
Here are two screenshots showing the issue, one for T=1, other for T=2.
<img width="2376" alt="image" src="https://user-images.githubusercontent.com/13998824/200030937-75cb117d-a662-43d7-9974-1a9b0a081aae.png">
<img width="2362" alt="image" src="https://user-images.githubusercontent.com/13998824/200030991-92976688-d9ae-4ee8-b907-1fbf02b5be36.png">
In the second screenshot above, there are green annotations on T=2 that should be showing up as red annotations when T=1.
I'm guessing it's just that the layer display doesn't know to capture annotations from the next frame.

arjunrajlab:
I think the challenge might be to pull the annotations from not just the current frame, but from all the relevant frames indicated by the offsets, etc.

arjunrajlab:
I'm still getting the original bug here on master…

arjunrajlab:
Confirmed that this is an issue with the display. Annotations are being attached to the correct frame coordinate, but the annotations are not displaying when the layer has an offset.

PR #315, opened 2022-11-16
Clean the debounce of getLayerHistogram

bruyeret:
Reformat getLayerHistogram to make it clearer.
Fix an infinite recursion.

Closes [#310](https://github.com/Kitware/UPennContrast/issues/310)

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -932,34 +932,50 @@ export class Main extends VuexModule {
       );
       if (!layer._histogram) {
         layer._histogram = {
-          promise: Promise.resolve(),
-          last: true,
-          next: null,
-          images: null
+          promise: Promise.resolve(null),
+          lastHistogram: null,
+          lastImages: null,
+          nextImages: null,
+          lock: false
         };
       }
 
       // debounce histogram calls
       let nextHistogram = () => {
-        if (layer._histogram.next) {
-          const images = layer._histogram.next;
-          layer._histogram.last = null;
-          layer._histogram.promise = this.api.getLayerHistogram(images);
-          layer._histogram.next = null;
-          layer._histogram.images = images;
-          layer._histogram.promise.then((value: any) => {
-            layer._histogram.last = value;
-            return null;
+        if (
+          layer._histogram &&
+          !layer._histogram.lock &&
+          layer._histogram.nextImages !== null
+        ) {
+          const histogramObj = layer._histogram;
+          const images = layer._histogram.nextImages;
+          histogramObj.lock = true;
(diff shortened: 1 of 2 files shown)


src/store/index.ts
```diff
@@ -932,34 +932,50 @@ export class Main extends VuexModule {
...
+            histogramObj.lastHistogram = value;
+          });
+          histogramObj.promise.catch(() => {
+            histogramObj.lastHistogram = null;
+          });
+          histogramObj.promise.finally(() => {
+            histogramObj.lastImages = images;
+            histogramObj.nextImages = null;
+            histogramObj.lock = false;
+            nextHistogram();
```
luciemac, line 964: You can write
```
 histogramObj.promise
    .then(...)
    .catch(...);
    .finally(...);
```

src/store/index.ts
```diff
@@ -932,34 +932,50 @@ export class Main extends VuexModule {
...
-        layer._histogram.next = images;
-        if (layer._histogram.last) {
-          nextHistogram();
-        }
+
+      const imagesToId = (images: IImage[] | null) => {
+        return (
+          images?.map(i => `${i.item._id}#${i.frameIndex}`).join(",") || "null"
+        );
+      };
```
luciemac, line 974: `const imagesToId = (images: IImage[] | null) => images?.map(i =>`${i.item._id}#${i.frameIndex}`).join(",") || null;`

src/store/index.ts
```diff
@@ -932,34 +932,50 @@ export class Main extends VuexModule {
...
       };
-      if (images !== layer._histogram.images) {
-        layer._histogram.next = images;
-        if (layer._histogram.last) {
-          nextHistogram();
-        }
+
+      const imagesToId = (images: IImage[] | null) => {
+        return (
+          images?.map(i => `${i.item._id}#${i.frameIndex}`).join(",") || "null"
```
luciemac, line 972: why do you return a string "null" and not just "null"?

bruyeret: This is made so that `imagesToId` always return a string.

luciemac: what about an empty string?

bruyeret: An empty string could be returned with `null` or `[]` as arguments

bruyeret:
This PR fixes issue #310 but the bug was already fixed by PR #309 which fixed issue #307.
This PR should be merged anyway because it refactors and clarifies this part of the code.
I kept the logic of PR #309.

luciemac:
LGTM

PR #318, opened 2022-11-23
Replace annotation channel with layerId

### LINKED ISSUE #310: Annotations not showing with time offset

bruyeret:
This is a breaking change as the server side annotations don't have the same model anymore.
Annotations had a property "channel" but the user input a layer.
Replace the property "channel" with a "layerId" as a channel can be retrieved from the layer ID.

As the user can change the channel of a layer, he can then change the channel of an annotation.

When the user deletes a layer, it doesn't delete the annotations linked to the layer but they can't be seen them anymore.
Should the linked annotations be deleted too?
Maybe open a dialogue to ask the user wether he wants to replace the layer of all annotations to be deleted, or delete them

Closes #310

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
@@ -63,12 +63,12 @@ class AnnotationSchema:
             },
             'coordinates': coordsSchema,
             'tags': tagsSchema,
-            'channel': {'type': 'integer'},
+            'layerId': {'type': 'string'},
             'location': locationSchema,
             'shape': shapeSchema,
             'datasetId': {'type': 'string', 'minLength': 1},
         },
-        'required': ['coordinates', 'tags', 'channel', 'location', 'shape', 'datasetId']
+        'required': ['coordinates', 'tags', 'layerId', 'location', 'shape', 'datasetId']
     }
 
 class Annotation(AccessControlledModel):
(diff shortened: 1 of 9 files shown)


arjunrajlab:
Hmm. I'm not an expert, but this strikes me as not a very good way to implement things. It makes the data model dependent on the view. Annotations really should be associated with a particular "frame" in the data: XY, T, Z, Channel.

There will be a lot of problems because people can change layers all the time. I'm also not even sure this solves the original issue very well, because as we go through T, the annotation should switch layers depending on what time is shown. 

@manthey @finetjul thoughts?

finetjul:
I agree, annotation should be associated to a channel, not a "layer".
The tool is however associated to a "layer".
When creating an annotation from a tool, we save into the annotation the time and channel of the layer of the tool.
For display purpose, we consider the annotations of each layer and display them with the layer color.

arjunrajlab:
I was anticipating that the solution to the problem is that for the display of a layer, when it picks what annotations to show, it must pull annotations from the frame coordinate plus the offset. Just like how the image it pulls to display also must be the frame coordinate plus the offset. But I am no expert. Perhaps @bruyeret if you can show me where this part is in the code, I could be of some assistance.

bruyeret:
I misunderstood the way the problem should be solved.
I thought that the annotations where per layer, because the user specifies a layer when creating a tool.
The layer is in fact only used to get the [XY, T, Z, Channel] data (except when specifying a fixed value), and the annotation can be displayed by any layer having the same [XY, T, Z, Channel].
Thank you @arjunrajlab, I think that I know how to fix the bug but I will reach out to you if needed!
I will close this pull request and start over.

PR #327, opened 2022-12-02
Fix layerShouldDisplayAnnotation

bruyeret:
It didn't take into account different types of IDisplayLayer & unrolling
The logic should now be fixed
Fixes [the bug mentionned ](https://github.com/Kitware/UPennContrast/pull/319#issuecomment-1335176242) in PR #319


luciemac:
LGTM

PR #331, opened 2022-12-12
Add compositing

### LINKED ISSUE #244: Tile compositing not working
@manthey not sure if the file reader was set to do this, but it seems that currently, the software doesn't do compositing. For instance, here is an ND2 file with a bunch of tiles, but the software doesn't composite it together into a big stitched image. Is that a  feature that is easy to enable?

https://www.dropbox.com/sh/aagt5bf24t0o7zm/AACPG12_mndS0pqMXgqG4aada?dl=0

bruyeret:
Add compositing option when importing compatible a single nd2 file.
Change the UI.

Closes  #244

### DIFF
--- a/src/utils/parsing.ts
+++ b/src/utils/parsing.ts
@@ -134,23 +134,23 @@ export function splitFilenames(filenames: string[]): string[][] {
     });
 }
 
-const enumType: { [index: string]: string } = {
-  t: "t",
-  time: "t",
-  timepoint: "t",
-  s: "xy",
-  stage: "xy",
-  position: "xy",
-  xy: "xy",
-  XY: "xy",
-  z: "z",
-  Z: "z",
-  slice: "z",
-  chann: "chan",
-  channel: "chan"
+const enumType: { [index: string]: TDimensions } = {
+  t: "T",
+  time: "T",
+  timepoint: "T",
+  s: "XY",
+  stage: "XY",
+  position: "XY",
+  xy: "XY",
+  XY: "XY",
+  z: "Z",
+  Z: "Z",
+  slice: "Z",
+  chann: "C",
+  channel: "C"
 };
 
-export function getStringType(inputString: string): string | undefined {
+export function getStringType(inputString: string): TDimensions | undefined {
   return enumType[inputString];
 }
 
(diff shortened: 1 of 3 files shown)


arjunrajlab:
I think it would be better to put the compositing checkbox just to the right of the Positions (XY) UI as indicated below.
Also, I would change the text from "Enable compositing" to "Composite positions into single image".

image

arjunrajlab:
I have noticed an issue with the file "normmedia_8well_col2_livecellgfp.nd2". If I upload it, it gets stuck on the "Variables" and "Assignments" screen and clicking "Submit" doesn't work. This problem appears to arise both when "Enable compositing" is checked or unchecked.

src/utils/parsing.ts
```diff
@@ -173,85 +173,70 @@ export function getFields(filenames: string[][]) {
...
-            }
-          }
-        });
-      }
+    values.sort((a, b) => a.localeCompare(b));
+
+    // Guess an ID (XY, Z, T, C) for this variable
+    let typename: TDimensions | undefined = undefined;
+    if (numberOfElement <= 1) {
+      return;
```
luciemac, line 210: Don't you need to return a value here ?

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -389,54 +433,125 @@ export default class NewDataset extends Vue {
...
+          );
+          break;
+      }
+    }
+    if (!channels) {
+      channels = ["Default"];
+    }
+
+    // Find all possible (XY, Z, T, C)
+    const sources: any[] = [];
```
luciemac, line 505: can't you add a type instead of "any"?

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -389,54 +433,125 @@ export default class NewDataset extends Vue {
...
+    const tilesMetadata = this.tilesMetadata;
+    if (!tilesMetadata) {
+      return;
+    }
+    for (let itemIdx = 0; itemIdx < this.girderItems.length; ++itemIdx) {
+      const item = this.girderItems[itemIdx];
+      const framesLength = tilesMetadata[itemIdx].frames.length;
+      for (let frameIdx = 0; frameIdx < framesLength; ++frameIdx) {
+        const source = {
```
luciemac, line 514: Is it really needed to create a variable here?

bruyeret:
> I have noticed an issue with the file "normmedia_8well_col2_livecellgfp.nd2". If I upload it, it gets stuck on the "Variables" and "Assignments" screen and clicking "Submit" doesn't work. This problem appears to arise both when "Enable compositing" is checked or unchecked.

I noticed that the submit button was slow, but it works for me, I have to wait 7/8 seconds after clicking
Does waiting work for you @arjunrajlab?

I found where the problem is
Uploading a multi-source file of 33536B takes 7.65s; 7059B takes 917ms
Uploading the multi-source file takes more time now because there is a "source" entry for each frame and it makes the file grow
The first file weights only 33kB but takes 7.65s to upload, which is very very slow (less than 5kB/s)
Do you have an idea about why @manthey?

src/utils/parsing.ts
bruyeret: No, this is in a `forEach` loop

arjunrajlab:
Yes, I did notice that the interface did eventually go through. But when it did, it still seemed to have an error on the next page. Also, this slowdown was not there before, so it must have been something introduced with this new code.

bruyeret:
What is the error you get on the next page? I don't get any error
Yes the slow down has been introduced with this new code
As I said, this is beacause the size of the multi-source file has grown and its upload is very slow
I'm waiting for David's opinion on the topic

bruyeret:
Yes I agree, I had a quick talk with David about that and he told me that 
> It will be slower to load a frame (it now has to load chunks from the file from lots of spots)

I don't know how we could make things faster, I'll discuss that with David

manthey:
Speed on composited images (and only composited image) will be slower on first load.  They should eventually be cached and just as snappy as other files.  To get, for instance, a thumbnail, we can't just read one section of a file, we have to read a bunch of sections to composite it.

arjunrajlab:
@manthey It's actually slow even WITHOUT compositing, though, which is puzzling.

arjunrajlab:
I think it would be fine if it is slower for composited files, but I think it should be the same for non-composited files, right?

bruyeret:
David will try to fix the problem on girder's side (it is something with file caching)
If it is too long or too hardto fix, I'll make it slow only for composited files

arjunrajlab:
Sounds good, thanks!

arjunrajlab:
Tried the file normmedia_8well_col2_livecellgfp.nd2 and still had problems. It was faster to process, but still quite slow to upload even if you do not check "composite". Once loaded, navigating the file (e.g. moving the XY slider) was much less responsive than before.

arjunrajlab:
On the same file, composite doesn't seem to work…

arjunrajlab:
Like, it doesn't show any image.

bruyeret:
I haven't bump David's changes into this branch yet, did you do it?

bruyeret:
I'll do it tomorrow, sorry for that!

luciemac:
@bruyeret can you add / update tests ?

arjunrajlab:
Also, the updates from @manthey did lead to some speed up, but didn't solve the underlying problem of the order(s) of magnitude slowdown.

arjunrajlab:
@manthey OK, I updated large_image to the latest version (19.*). I am still getting an issue with this file:
normmedia_8well_col2_livecellgfp
It seems to parse correctly, but the image just never shows up (see screenshot). Are you able to get this file to work for you?

image

arjunrajlab:
I tested it and I think it is good!

arjunrajlab:
Does this completely get rid of the JSON parsing if there is no compositing?

arjunrajlab:
I tried checking "composite positions" for PMAIP1_NGFR_WNT5A.nd2" and no image showed up. But without checking that box, it was fine.

arjunrajlab:
I added PMAIP1_NGFR_WNT5A.nd2 to the Dropbox: https://www.dropbox.com/sh/89axp9vwtt81o4v/AADHKe4Hon64InYWtnY9hRjFa?dl=0

arjunrajlab:
OK, I think we can merge now!

luciemac:
LGTM

PR #340, opened 2023-01-16
Add hotkey to toggle all layers' visibility

bruyeret:
Remove an old hack to toggle all zMaxMerge

Closes 99


src/store/index.ts
```diff
@@ -680,6 +681,35 @@ export class Main extends VuexModule {
...
+    if (!layers || !this.dataset) {
+      return;
+    }
+    const currentZMaxMerge = layers.every(
+      layer => layer.z.type === "max-merge"
+    );
+    const newLayerZ: IDisplaySlice = currentZMaxMerge
+      ? { type: "current", value: null }
+      : { type: "max-merge", value: null };
+    layers.every(layer => (layer.z = newLayerZ));
```
luciemac, line 696: Why don't you use a forEach loop here?
In my opinion, every method exists to check if every element of an array pass a test

bruyeret: Yes sorry, I copy pasted something and it went wrong

luciemac:
LGTM

PR #344, opened 2023-01-23
Better scale sparse histograms on sparse images

manthey:
If the second most common value in a histogram is rare compared to the most common value, increase the vertical scale.


arjunrajlab:
Works for me!

PR #345, opened 2023-01-23
Fix some top bar formatting.

manthey:
Specifically, long filenames wrapped poorly, and the application tile would get truncated.


arjunrajlab:
Tested, looks good to me, thanks!

bruyeret (approved):
Thanks, it is looking better
LGTM

PR #333, opened 2022-12-21
Property computation and Analyze panel refactor

bruyeret:
See issue #328 for more information on the content of this PR.

Closes #328

PR #346, opened 2023-01-27
Add import/export feature

### LINKED ISSUE #255: Import and export of annotations
Handy feature would be the import and export of annotations and their associated properties. We are thinking the annotations could be basically serialized as JSON and exported and imported in that way. The properties would be the ones currently displayed in the list on the annotation browser. UI would be buttons at the bottom of the annotation browser area.

### LINKED ISSUE #285: Add an endpoint to upload many annotations to server at once
Can we have an endpoint that the worker can access to send multiple annotations to the database all at the same time?
@luciemac @zjniu

### LINKED ISSUE #329: New Annotation Display Mechanism Causes Freezing
After the annotation display mechanism was changed in 66be8a71e159d4b61adea2451253190b4539f448, old datasets created before this commit experiences UI freezing for most actions (e.g., turning on/off channels, changing XY, clicking on NimbuImage to return to the home page). Newer datasets created using the latest commit db34c4df8d740838b2935f593ff4030077611362 does not seem to have the same issue. However, as a temporary fix, it is possible to get rid of freezing when reverting the AnnotationViewer.vue file back one commit to 404236a8b13dbdd8dc68d2449e27b4461719ab31.

### LINKED ISSUE #355: Improve performance when changing annotation style
when hovering an annotation on the gui for example

bruyeret:
The import / export buttons are currently in the annotation browser but can be moved in the future (but after PR #333 to avoid conflicts)

Fix some performance issues: rework annotation viewer, use Maps and Sets
Fix an untracked bug which made annotation connections between annotations of different datasets possible
Change style of annotations instead of removing and adding them
Add some endpoints for bulk POST and DELETE
Clean connections and property values orphans when deleting annotations
Add downloadToClient helper function

Annotation property values exported before/after PR #333 but computed with a worker after/before PR #333 can't be imported
This problem shouldn't be a problem for final users if the update of the workers is done at the same time as the update of the frontend
A workaround is to not export or import values and recompute them

Closes #255 
Closes #329 
Closes #285
Closes #355

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
@@ -20,6 +20,8 @@ class Annotation(Resource):
         self.route('POST', (), self.create)
         self.route('PUT', (':id',), self.update)
         self.route('POST', ('compute',), self.compute)
+        self.route('POST', ('multiple',), self.createMultiple)
+        self.route('DELETE', ('multiple',), self.deleteMultiple)
 
     # TODO: anytime a dataset is mentioned, load the dataset and check for existence and that the user has access to it
     # TODO: creation date, update date, creatorId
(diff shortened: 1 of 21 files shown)


luciemac:
LGTM

arjunrajlab:
Looks good to me. It would be good to have the option for a filename/location for the file. Also, when importing, can we have an option to append or overwrite?

luciemac:
LGTM

bruyeret:
Updated the description of the PR, a lot of features/improvements/fixes are brought by this PR.
Sorry for the size of the PR.

@manthey One of the performance issue came from `showLabels` being `true` by default in the annotation layer.
This caused the layer to call `CanvasRenderingContext2D.fillText()` a lot of times without displaying text.
Shouldn't this flag be set to `false` as no text is displayed by default anyway and this hurts the performance?

arjunrajlab:
Looks good to me!

luciemac (approved):

PR #362, opened 2023-03-22
Fix auto page change in annotation list

### LINKED ISSUE #359: Weird behavior in annotation list when hovering.
Noticed some unexpected behavior when hovering over annotations in the annotation browser.

1. When annotations are not being sorted or when sorting by ids in ascending order, the annotation list behaves correctly. However, if sorting by ids in descending order, hovering over the annotations causes the list to flip to indices 4 to 0 as shown in the following video.

[Screencast from 03-10-2023 10:36:06 PM.webm](https://user-images.githubusercontent.com/87096829/224462963-5a93506b-453b-409d-9037-64f94110449c.webm)

2. When annotations are not being sorted, the annotation list behaves correctly. However, if sorting by property value in ascending or descending order, hovering over the annotations causes the list to flicker nonstop as shown in the following video.

[Screencast from 03-10-2023 10:11:57 PM.webm](https://user-images.githubusercontent.com/87096829/224463035-1412ce0c-a5e7-4e4a-a62c-96b1813cbb53.webm)

bruyeret:
When sorting the items in the annotation list, the page computed for an annotation was wrong.
The ideal thing would be to get the filtered and sorted items from the v-data-table but the component doesn't expose such thing or any function which could make it easy to compute the page of an item.
The solution chosen is to get the component using $refs and compute the filtered/sorted list is it is done in the component.

Closes #359


arjunrajlab:
Just tested now, looks good!

src/components/AnnotationBrowser/AnnotationList.vue
```diff
@@ -301,20 +302,58 @@ export default class AnnotationList extends Vue {
     return this.tableOptions.itemsPerPage;
   }
 
+  get vDataTable() {
+    const vDataTableParent = this.$refs.dataTable as any;
+    if (!vDataTableParent) {
+      return null;
+    }
+    const vDataTable = vDataTableParent?.$children?.[0];
```
luciemac, line 310: You already check vDataTableParent existence

src/components/AnnotationBrowser/AnnotationList.vue
```diff
@@ -301,20 +302,58 @@ export default class AnnotationList extends Vue {
...
+    if (!vDataTable) {
+      return null;
+    }
+    return vDataTable;
```
luciemac, line 314: Can it be undefined?
Otherwise, I think you can have only one return statement

bruyeret: It can be null or a vDataTable element, I changed it to a one liner

luciemac:
LGTM

PR #367, opened 2023-03-24
Darken annotations stroke

### LINKED ISSUE #365: Stroke color change for annotations to be less than just the color itself
When heavily contrasted, the point annotations can "disappear" because the stroke color is the color of the layer, and the fill has an opacity of 0.5. Suggested fix is to change the stroke color slightly darker than the layer color.

image
image

bruyeret:
Closes #365


luciemac:
LGTM

PR #370, opened 2023-04-06
Fix an issue debouncing histogram calls.

### LINKED ISSUE #356: Performance issue when scrubbing XY coordinates
There still appears to be some lagginess when scrubbing the XY slider for DDX58_AXL_EGFR_well2.nd2. This problem has been there for a few months now and was not resolved by the latest import/export PR.

manthey:
If you switch frames quickly, multiple histogram calls are made to the server.  There was a bug that if there was a pending histogram call, the next histogram call might be skipped.

Fixes #356.

### DIFF
--- a/src/components/ValueSlider.vue
+++ b/src/components/ValueSlider.vue
@@ -62,7 +62,7 @@ export default class ValueSlider extends Vue {
 
   set slider(value: number) {
     const numberValue = typeof value === "number" ? value : parseInt(value);
-    if (numberValue == this.internalValue) {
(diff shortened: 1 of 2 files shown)


manthey:
Curiously, the browser is caching less than expected -- most queries (including histograms) have ETags, but Chrome only uses the disk cache some times and is actually requesting the full response at other times.  It'd be worth looking into that as a separate issue.

arjunrajlab:
Works much more smoothly now! I think we can merge. I did notice some lagginess upon first load, but I should test on another machine.

PR #372, opened 2023-04-20
Refactor configurations

### LINKED ISSUE #341: Rework configurations
Rework configurations:
- A configuration contains the following data: properties, tools, layers (annotations are per dataset)
- Each dataset has a default configuration when uploaded
- The user can create new configurations from a dataset
- The user can import configurations when the number of channels is the same
- The user can rename configurations
- The user can remove a configuration from a dataset
- When a configuration doesn't belong to any dataset it is unreachable and then definitly deleted and the properties, tools and layers are removed from the database (ask for confirmation before the last "remove from dataset" or "delete dataset")

bruyeret:
Update collections / configurations using #341 

Closes #341


src/components/CustomFileManager.vue
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  @Watch("isLoggedIn")
+  async fetchLocation() {
+    const publicFolder = await this.store.api.getUserPublicFolder();
+    if (publicFolder) {
+      this.defaultLocation = publicFolder;
+    } else {
+      this.defaultLocation = this.store.girderUser;
+    }
```
luciemac, line 169: this.defaultLocation = publicFolder ? publicFolder : this.store.girderUser; ??

src/layout/BreadCrumbs.vue
```diff
@@ -20,37 +20,110 @@
...
+
+  mounted() {
+    this.refreshItems();
+  }
+
+  @Watch("$route")
+  async refreshItems() {
+    let index = 0;
+    for (const { title, to, recordPromise } of [
+      {
```
luciemac, line 94: Can you create a variable to make it more readable ?

src/store/GirderAPI.ts
```diff
@@ -175,6 +94,26 @@ export default class GirderAPI {
...
+    const resourceObj: { folder: string[]; item: string[] } = {
+      folder: [],
+      item: []
+    };
+    for (const resource of resources) {
+      const type = resource._modelType;
+      if (type === "folder" || type === "item") {
+        resourceObj[type].push(resource._id);
+      }
+    }
```
luciemac, line 107:
```
resourceObj = resources.filter((resource) => {
  const type = resource._modelType;
  return type === "folder" || type === "item";
});
```
??

src/store/GirderAPI.ts
```diff
@@ -175,6 +94,26 @@ export default class GirderAPI {
     return result.data.length > 0 ? result.data[0] : null;
   }
 
+  async move(resources: IGirderSelectAble[], folderId: string) {
```
luciemac, line 97: Remove async?

src/store/GirderAPI.ts
```diff
@@ -580,41 +528,84 @@ function asDataset(folder: IGirderFolder): IDataset {
...
+): IDatasetConfigurationBase {
+  const config: IDatasetConfigurationBase = {
+    compatibility: getDatasetCompatibility(dataset),
+    layers: getDefaultLayers(dataset),
+    tools: [],
+    propertyIds: [],
+    snapshots: []
   };
-  return tool;
+  return config;
```
luciemac, line 566: return {
    compatibility: getDatasetCompatibility(dataset),
    layers: getDefaultLayers(dataset),
    tools: [],
    propertyIds: [],
    snapshots: []
  };

src/views/configuration/DuplicateImportConfiguration.vue
```diff
@@ -0,0 +1,111 @@
...
+  async submit(configurations: IDatasetConfiguration[]) {
+    const dataset = this.dataset;
+    const parentFolder = this.path;
+    if (!dataset || !parentFolder) {
+      return;
+    }
+
+    // Duplicate each configuration and create a view for it
+    const promises: Promise<any>[] = [];
+    for (const configuration of configurations) {
```
luciemac, line 89: is it more efficient than using a .map ?

src/views/configuration/ImportConfiguration.vue
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  async submit(configurations: IDatasetConfiguration[]) {
+    if (!this.store.dataset) {
+      return;
+    }
+
+    // Create a view for each configuration
+    const promises: Promise<any>[] = [];
+    for (const configuration of configurations) {
```
luciemac, line 35: same here

src/store/AnnotationsAPI.ts
```diff
@@ -69,14 +69,11 @@ export default class AnnotationsAPI {
...
+  async getAnnotationsForDatasetId(datasetId: string): Promise<IAnnotation[]> {
     const annotations: IAnnotation[] = [];
-    const pages = await fetchAllPages(
-      this.client,
-      "upenn_annotation",
-      id,
-      100000
-    );
+    const pages = await fetchAllPages(this.client, "upenn_annotation", {
+      params: { datasetId, limit: 100000, sort: "_id" }
```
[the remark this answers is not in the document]
bruyeret: The limit value is the limit of documents per api call
There already is a default limit of 1000 in `fetchAllPages`
It is only at 1000 to avoid fetching to much data at the same time
The limit is adapted by the caller depending on the things that are fetched
Here, for annotations, a limit of 100000 is fine but it depends only on the size of the documents fetched

src/store/GirderAPI.ts
bruyeret: `resourceObj` is an object containing two arrays of IDs, I can't build it using a filter
I could use a `forEach` instead of the `for ... of`, do you think it is worth it?

luciemac: Oh I see ! you can leave it this way I think

src/store/AnnotationsAPI.ts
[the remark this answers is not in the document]
bruyeret: Oh ok, maybe I should instead increase the default limit to 100000 and not specify anything when calling `fetchAllPages`

arjunrajlab:
Here, it would be nice to have the "Dataset: ******" be a dropdown showing all the available Datasets in the Collection.

image

That way, one could quickly access all Datasets in the Collection.

PR #384, opened 2023-05-11
Zoom histogram

### LINKED ISSUE #366: Can we build a "zoom" feature for the histograms to make them easier to adjust?
Can we make the histogram zoom in with a scroll wheel? Make the resolution to 1024 instead of 512. Maybe make the resolution 4096. Scroll wheel is fine.

bruyeret:
Add zoom to histograms and fix the clamping of absolute values in the histogram

Closes #366

### DIFF
--- a/src/components/ContrastHistogram.vue
+++ b/src/components/ContrastHistogram.vue
@@ -16,37 +16,37 @@
       :id="`input-${_uid}`"
     />
     <div class="wrapper">
-      <svg :width="width" :height="height">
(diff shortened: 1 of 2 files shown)


arjunrajlab:
I tested Zoom histogram and max-merge annotations, and both look great to me, thanks! If Lucie says good, then we can merge.

arjunrajlab:
I haven't tested the endpoint yet.

luciemac:
LGTM

PR #385, opened 2023-05-12
Display annotations of all images in max-merge

### LINKED ISSUE #383: When you do max-merge, can we show the max-merge of the annotations?
It would be good, when max-merged, to show the max-merge of the annotations. For instance, if I have spots across a range of different Z values, then if I merge Z, I'd like to see all the spots.

bruyeret:
Closes #383


luciemac:
LGTM

PR #388, opened 2023-05-16
Group layers

### LINKED ISSUE #382: Can we build a "meta-layer" that is a composite of multiple other layers?
It would be good to have a meta-layer. I think this can be most easily achieved as a grouping of existing layers. In a way it's just a toggle group. The most important difference is that if you do "Unroll", it should show the meta-layer instead of the individual layers.

bruyeret:
Use drag & drop to move layers and group them together
Grouping layers together will show them together in unroll mode

Bonus feature:
The order of the layers in the layer list defines the order in which they will be displayed in unroll layers mode

Closes #382


arjunrajlab:
Works great!! So nice! One minor point: There is a "Create group" region, but it's a little unclear what to do. I think if we could change the text to "Drag layer here to create group" that would be perfect. Thanks @bruyeret!

arjunrajlab:
Looks great to me! Ready to merge from my end!

luciemac:
LGTM

PR #390, opened 2023-05-22
Add pixel value on hover

### LINKED ISSUE #381: Pixel value on hover on image
It would be nice to have a pixel value show up at the bottom of the screen for the pixel value for all the layers. I think the format should be: [ Brightfield = 1039, Cy3 = 4893, DAPI = 3999]. If the channel is not visible, then don't show the value.

I think most people will want the real value and not the approximate value held in the UI. So it will have to fetch from server. I don't think it's important that it be real time, so if it's a small fraction of a second, that's fine. I think we can hide/freeze the value until we move to a new area and wait a bit, then show the new value. 

The option to show the pixel value can be kept in the "Settings" tab.

image

bruyeret:
Closes #381


arjunrajlab:
Looks great, thank you!

luciemac:
LGTM

PR #398, opened 2023-05-24
Move buttons between annotation list and actions

### LINKED ISSUE #393: Move buttons around in the Annotation tab; automatically expand the list panel
I think it might be good to rearrange the buttons here:

image

I think we could move the buttons in the "Annotation" panel (like Export to JSON, etc.) to the "Actions" subpanel. I think we could move the "Delete selected" and "Tag selected" to the Annotation panel because they are commonly used.

bruyeret:
Also:
- fix selection of annotations in the list
- fix an annotation list getter calling `sort` (in-place) on a store variable

Closes #393


arjunrajlab:
Looks ready to merge to me!

luciemac:
LGTM

PR #399, opened 2023-05-24
Move property checkboxes to the annotations pannel

### LINKED ISSUE #394: Move property checkboxes to annotation menu
Currently, if you want to show a particular property in the annotation list, you have to select it from the "analyze" tab:

image

That's a bit counterintuitive, since usually people want to choose from the available properties to display from the annotation tab.

I think it would be better to have a list of all properties in the Annotation panel. I suggest making a new panel here called "Property list":

image

That would basically have all the computed properties listed and checkboxes for both "List property value" (which would be like "In list") and "Use property value as filter" (which would be like "As filter").

I think it would also be useful to organize it by tags. Perhaps we could list, for each property, also the tags as "chips" so that we know that it's applied to all [cell] and [RNA] or whatever tags. But that is not as important.

bruyeret:
Closes #394

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationBrowser.vue
+++ b/src/components/AnnotationBrowser/AnnotationBrowser.vue
@@ -7,6 +7,7 @@
       <v-expansion-panels hover multiple v-model="expanded">
         <annotation-filters></annotation-filters>
         <annotation-actions></annotation-actions>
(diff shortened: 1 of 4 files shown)


arjunrajlab:
Works great!

luciemac:
LGTM

PR #401, opened 2023-05-25
Fix reset of osm layers when changing dataset

### LINKED ISSUE #396: Flicker in collections
We have noticed a slight visual glitch in the collections view. Here is a set of datasets that can reproduce the issue:

https://www.dropbox.com/scl/fo/65yduxqxkv8n3bet0la5b/h?dl=0&rlkey=6c6rbcjih336ujdk2mvo6o4x5

The problem is that when you switch to different datasets, when I navigate in XY, it flickers back to the first dataset in the collection briefly. Here's a video showing the issue:

https://www.dropbox.com/s/jo7s1rsck8r53m8/20230523_Nimbus_Flicker.mov?dl=0

bruyeret:
Close #396

### DIFF
--- a/src/components/ImageViewer.vue
+++ b/src/components/ImageViewer.vue
@@ -168,6 +168,8 @@ export default class ImageViewer extends Vue {
 
   private ready = { layers: [] };
 
+  private resetMapsOnDraw = false;
+
   maps: IMapEntry[] = [];
   tileWidth: number = 0;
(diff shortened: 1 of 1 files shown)


luciemac:
LGTM

arjunrajlab:
Sorry, I forgot to say: looks good to me as well. Tested and the problem seems better.

PR #402, opened 2023-05-25
Add remove connections tool

### LINKED ISSUE #400: Add a "Remove connections" tool
It would be nice to have a Remove Connections tool to disconnect annotations from each other. I think having the same tag selection structure as the Connect Selected tool would be good:

image

That way, if you want to select what you would in particular like to Disconnect from and to, it is possible.

bruyeret:
Close #400

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/connections.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/connections.py
@@ -21,6 +21,7 @@ class AnnotationConnection(Resource):
         self.route('PUT', (':id',), self.update)
         self.route('POST',('connectTo',) , self.connectToNearest)
         self.route('POST', ('multiple',), self.multipleCreate)
+        self.route('DELETE', ('multiple',), self.deleteMultiple)
 
     # TODO: anytime a dataset is mentioned, load the dataset and check for existence and that the user has access to it
     # TODO: load both childe and parent annotations, and check that they share existence, access and location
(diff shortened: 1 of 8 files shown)


luciemac:
LGTM

arjunrajlab:
I tried it, but I get these green circles everywhere and it doesn't seem to actually remove any connections:

image

bruyeret:
Sorry, I didn't mention I added an endpoint to girder, it needs to be rebuilt

arjunrajlab:
Mostly works! The one issue is that if I don't specify the child tag, then it says "none" instead of "any"… and doesn't appear to work. I would expect it to remove all possible connections. In other words, perhaps it should act like a "*" instead of like a none. Same is true if I don't specify the parent tag.

image

bruyeret:
This is expected:
- Inclusive mode => annotation match if it contains some of the specified tags
- Exclusive mode => annotation match if it contains all of the specified tags

If no tag is specified, this is equivalent to:
- Invlusive mode => annotation can't contain a specified tag => no annotation will match
- Exclusive mode => annotation will contain all specified tag => any annotation will match

It is handeled as a special case only for the UI (show None or All)
No matter the number of tags, the logic is always the same:
```js
if (tagsInclusive) {
    // There is a tag of "tags" in the annotation tags
    // Filter on: annotation has tag1 OR tag2 OR tag3...
    output = output.filter(annotation =>
      tags.some((tag: string) => annotation.tags.includes(tag))
    );
  } else {
    // All tags of "tags" are in the annotation tags
    // Filter on: annotation has tag1 AND tag2 AND tag3...
    output = output.filter(annotation =>
      tags.every((tag: string) => annotation.tags.includes(tag))
    );
  }
```

arjunrajlab:
Ah, okay, now I understand. I think it's still perhaps a little confusing to the end user, but let's see. Perhaps we can just add a little help text to orient the user, like "Click here to switch between all or none" or something like that. But yes, let's merge, looks great!

PR #414, opened 2023-05-30
Fix annotation list checkboxes for selection

### LINKED ISSUE #413: Select all button not working
In the following, if I check the top button, it doesn't seem to select all in the list, see image below. Also, in general, selection and the list seems to be a bit buggy. If you select with the select tool, it doesn't mark the checkboxes anymore.

image

bruyeret:
Close #413

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationList.vue
+++ b/src/components/AnnotationBrowser/AnnotationList.vue
@@ -52,6 +52,8 @@
         :items="filteredItems"
         :headers="headers"
         show-select
+        item-key="annotation.id"
+        v-model="selectedItems"
         :page="page"
         @update:items-per-page="itemsPerPage = $event"
         @update:group-by="groupBy = $event"
@@ -72,8 +74,8 @@
               <td :class="tableItemClass">
                 <v-checkbox
                   hide-details
-                  multiple
                   title
+                  :value="item.isSelected"
                   @click.stop="() => toggleAnnotationSelection(item.annotation)"
                 />
               </td>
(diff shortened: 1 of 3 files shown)


luciemac:
LGTM

arjunrajlab:
Looks great! Only thing: currently, when I click the "select all" box, it only selects from the visible part of the list (like annotations 1-10). Is it possible to select all of them?

bruyeret:
I can add some code to do this but this is the default behavior of the data table
As this is how most data tables work, users may expect this behavior
I could instead add a "Use filter as selection" in the annotation actions:
image

What do you think?

arjunrajlab:
I asked around in lab, and I think they would like to have the behavior as I describe: if you click "select all" it selects all from the entire list, not just what is displayed.

Also, it seems that if the list is very long (thousands), then if you show "all" in the list, the interface becomes very sluggish. Perhaps we can eliminate the "all" option and just have the options for 10, 50, and 200. So 200 would be the max allowed.

bruyeret:
The last commit should fix the select all and items per page options issue @arjunrajlab!

src/components/AnnotationBrowser/AnnotationList.vue
```diff
@@ -227,16 +254,18 @@ export default class AnnotationList extends Vue {
...
+        return "-";
+      }
 
-    if (!Object.keys(values).includes(propertyId)) {
-      return "-";
-    }
-    return values[propertyId];
+      if (!Object.keys(values).includes(propertyId)) {
+        return "-";
+      }
```
luciemac, line 266: Factorize return : 
````
      if (!values || !Object.keys(values).includes(propertyId)) {
        return "-";
      }
````

File: src/components/AnnotationBrowser/AnnotationList.vue [hunk 2]
```diff
@@ -246,6 +275,24 @@ export default class AnnotationList extends Vue {
...
   }
 
+  get selectAllIndeterminate() {
+    const nSelected = this.selectedItems.length;
+    return nSelected > 0 && nSelected < this.filteredItems.length;
+  }
+
+  get selectAllValue() {
+    const nSelected = this.selectedItems.length;
+    return nSelected === this.filteredItems.length;
```
luciemac, line 285: Not sure `nSelected` is needed here. I think I'll go with a simple: `this.selectedItems.length === this.filteredItems.length;`

PR #423, opened 2023-06-07
Fix connectTo when there is no connection

bruyeret:
(A quick fix)
If there is no annotation to connect to, the server used to return a 500 error
Also reformat the code a little bit


luciemac:
LGTM

bruyeret:
This PR can be merged @luciemac 
This is a quick fix without a related issue

PR #424, opened 2023-06-07
Fix clicking an annotation change page

### LINKED ISSUE #418: Clicking on annotation does not take you to that position in the list anymore
I'm not sure when this bug happened, but it seems to have been here for a while, I think.

bruyeret:
Also reformat the computation of hovered annotation

Close #418


arjunrajlab:
Just tested, looks great!

luciemac:
LGTM

PR #425, opened 2023-06-07
Add (command/ctrl)+del shortcut

### LINKED ISSUE #404: Keyboard shortcut for Delete Selected (Delete key)
Could we add a keyboard shortcut for delete selected (Command/Control-delete key)? It may be better to have command-delete instead of plain delete to avoid inadvertent deletion.

bruyeret:
On Mac: command+del
On Windows and Linux: ctrl+del

Close #404

### DIFF
--- a/src/components/ImageViewer.vue
+++ b/src/components/ImageViewer.vue
@@ -82,6 +82,7 @@
 // in cosole debugging, you can access the map via
 //  $('.geojs-map').data('data-geojs-map')
 import { Vue, Component, Watch } from "vue-property-decorator";
+import annotationStore from "@/store/annotation";
 import store from "@/store";
 import geojs from "geojs";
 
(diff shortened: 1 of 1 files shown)


arjunrajlab:
On linux, can we make it control backspace instead of control delete?

arjunrajlab:
Also, on Mac, same thing. Right now, it's targeting the wrong "delete", so it doesn't work on my laptop (which only has backspace, but not delete).

arjunrajlab:
I think backspace would work let’s try that. Sent from my iPhoneOn Jun 7, 2023, at 12:09 PM, Thibault Bruyère ***@***.***> wrote:﻿
I am using a builtin feature of a library to have command on mac and ctrl on others
It is very hard to detect the user's operating system
Does it work if the binding is command+backspace or ctrl+backspace?

bruyeret:
Ok it's done! I just pushed a commit but it is not visible yet in this PR

arjunrajlab:
Works now! Looks good to me!

PR #426, opened 2023-06-07
Make worker menu stand out

### LINKED ISSUE #416: Can we make the popup window stand out a little more?
The popout window is good, but tends to blend into the overall background, which has led to some inadvertent clicks on "unroll". Can we make it stand out a bit more, perhaps with a drop shadow or outline or something?

image

bruyeret:
Add a border to the menu

Close #416


arjunrajlab:
Looks good to me!

luciemac:
LGTM

PR #430, opened 2023-06-08
Avoid fetching annotation twice

### LINKED ISSUE #419: Exporting large number of annotations to CSV crashes entire UI
We have found that when exporting >100K annotations to CSV, the operation works and produced the CSV, but then the entire UI shuts down (like, the terminal running the server actually quits). I saw in a previous issue that perhaps this is related to the fact that the CSV is generated by the front end, but not sure.

bruyeret:
May Close #419

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -527,10 +527,16 @@ export class Main extends VuexModule {
       datasetView.lastViewed = Date.now();
       this.setDatasetViewImpl(datasetView);
       const promises: Promise<any>[] = [
-        this.api.updateDatasetView(datasetView),
-        this.setSelectedDataset(datasetView.datasetId),
-        this.setSelectedConfiguration(datasetView.configurationId)
+        this.api.updateDatasetView(datasetView)
       ];
+      if (this.dataset?.id !== datasetView.datasetId) {
+        promises.push(this.setSelectedDataset(datasetView.datasetId));
+      }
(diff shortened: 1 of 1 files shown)


bruyeret:
Ready to merge if it looks ok to you @luciemac

src/store/index.ts
```diff
@@ -527,10 +527,16 @@ export class Main extends VuexModule {
       datasetView.lastViewed = Date.now();
       this.setDatasetViewImpl(datasetView);
       const promises: Promise<any>[] = [
-        this.api.updateDatasetView(datasetView),
-        this.setSelectedDataset(datasetView.datasetId),
-        this.setSelectedConfiguration(datasetView.configurationId)
+        this.api.updateDatasetView(datasetView)
       ];
+      if (this.dataset?.id !== datasetView.datasetId) {
```
luciemac, line 532: Can't `dataset` be null ?

luciemac:
then LGTM

src/store/index.ts
bruyeret: It can be null
In that case, `this.dataset?.id` it is different from `datasetView.datasetId` and the dataset must be reloaded

bruyeret: No `datasetView` shouldn't be `null`, at least not from the signature of `getDatasetView`
There could be an error if the dataset view for the corresponding ID doesn't exist, but the API makes the asumption that the ID is valid

PR #438, opened 2023-06-12
Filter UI

### LINKED ISSUE #429: Change "Filter" UI a little
In talking with people in the lab, the filter panel is super useful. I think we should probably just have it not even be a panel but just a header that is always there at the top. Also, I think we can simplify the UI a lot and make it both more compact and easier to understand. Here's my mockup below. The functionality would be exactly the same, but the options are hopefully a bit more clear.

image

bruyeret:
Change filter UI

Change and factorize tag exclusive/inclusive logic

Old logic:
- Exlusive: each filter tag is in input tags list
- Inclusive: at least one filter tag is in input tags list

New logic:
- Exclusive: filter tags list is equal to input tags list
- Inclusive: input tags list is included in filter tags list

Note that: Old logic exclusive = New logic inclusive

Close #429

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationFilters.vue
+++ b/src/components/AnnotationBrowser/AnnotationFilters.vue
@@ -1,33 +1,16 @@
 <template>
-  <v-expansion-panel dense>
-    <v-expansion-panel-header>
-      Filters <v-spacer></v-spacer>
-    </v-expansion-panel-header>
-    <v-expansion-panel-content>
-      <v-container>
-        <v-row>
-          <v-col class="pa-0">
-            <tag-filter-editor v-model="tagFilter"></tag-filter-editor>
-          </v-col>
-        </v-row>
-        <v-row>
-          <v-col class="pa-0">
-            <shape-filter-editor v-model="shapeFilter"></shape-filter-editor>
-          </v-col>
-        </v-row>
-        <v-row v-for="propertyId in propertyIds" :key="propertyId">
-          <property-filter-histogram
-            :propertyId="propertyId"
-          ></property-filter-histogram>
-        </v-row>
-        <v-row>
-          <v-col>
-            <roi-filters></roi-filters>
-          </v-col>
-        </v-row>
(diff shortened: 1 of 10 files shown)


arjunrajlab:
Also, when the filter is there, the connection between the list and the annotations (clicking either on the list or the annotation) seems broken.

arjunrajlab:
Actually, the connection is still good, I just had to refresh. I'll write if I can reproduce the issue.

bruyeret:
There is very little room to fit every filter, so I need a mechanism to make the elements wrap when there is not enough room.
For example, when a tag is very long, the tag filter should be on its own line:
image

When selecting "Point" in the dropdown, nothing changes for me, this is only when adding "points" to the list of tags that the button moves down.
There is another reason for this wrapping: the ROI filter which should show the different regions of interest and the property filter. You can see how it looks when you have a lot of filters which take space:
image


Do you want me to reduce the space between filters to allow for bigger tags before the element wraps? Or a different wrapping behavior?

arjunrajlab:
Ah, I see. I think the wrapping behavior is fine. Could we just remove this big space here in the screenshot below? I think that would help save space.

image

arjunrajlab:
Like, the word "Point" only takes up a small portion of the horizontal real estate taken up by the dropdown, which seems inefficient, given that Point is the longest option of the dropdown.

luciemac:
LGTM

arjunrajlab:
Also this dropdown has a lot of empty space:

image

bruyeret:
Is it better now @arjunrajlab?

arjunrajlab:
Looks great, ready to merge!

bruyeret:
@luciemac Sorry I messed up with my branches, the last force push removes a commit which shouldn't be in this PR

arjunrajlab:
Tested, looks good to me!

luciemac:
LGTM

arjunrajlab:
Let's merge!

PR #439, opened 2023-06-13
Make dataset updates less frequent

### LINKED ISSUE #432: Snapshots not working propertly
When I click on a snapshot in the list in Snapshots, it doesn't seem to "move" to that snapshot.

bruyeret:
Fixes some bugs which made the viewer sometimes reset

Close #432

### DIFF
--- a/src/components/Snapshots.vue
+++ b/src/components/Snapshots.vue
@@ -146,7 +146,7 @@
         :items-per-page="5"
         item-key="key"
         class="accent-1"
-        @click:row="item => loadSnapshot(item.name)"
+        @click:row="loadSnapshot"
       >
         <!-- Search bar -->
         <template v-slot:top>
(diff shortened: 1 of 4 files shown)


luciemac:
LGTM

arjunrajlab:
So Snapshots still has a minor bug… if I put a snapshot called "Snap1" on XY=1 and "Snap2" on XY=2, then when I click either one, it doesn't take me to the right XY. It just moves the box, but keeps it at the current XY.

arjunrajlab:
@bruyeret :point_up:

luciemac:
LGTM

arjunrajlab:
Just tested, the bug I had noted is fixed. However, snapshots do not seem to store the contrast settings: it should recall the contrast settings for every Snapshot. That is how it used to work, so perhaps the code for handling that is still in there?

bruyeret:
I made a new issue from your comment @arjunrajlab

PR #442, opened 2023-06-13
Layer idx

### LINKED ISSUE #421: Refer to layers by LayerID (in tools and annotations) rather than layer index
Referring to layer index has the problem that if you change the layer order, it will screw up references from tools, properties, etc.

bruyeret:
**Replace layer indices with layer ids**

Workers API changed, each reference to a layer is now made using an ID
Old tools used to use layer indices and now use layer IDs

@arjunrajlab @zjniu
This PR is a draft as the worker API has changed, so that the workers can be changed to match the new API
Another important thing, in `devops\girder\annotation_client\annotation_client\annotations.py`, at line 194:
```python
"channelId": connectTo['layer']
```
This line doesn't work as there is no hard relation between layers and channels.
It works even less now that `connectTo['layer']` is an ID (a string) and not an index (a number).

Should be merged after #438

Close #421


luciemac:
LGTM

PR #444, opened 2023-06-16
Snapshots update

### LINKED ISSUE #440: Minor UI fix for Snapshot modal dialog
In the modal dialog for saving the snapshot, it would be good if it enabled typing in the "name" field right away, and if you hit enter, then it triggers the "Create" button.

image

### LINKED ISSUE #441: Make initial snapshot size a little smaller
When you first click "Snapshot", the default rectangle is the size of the whole frame, which is a bit hard to see, especially if you are zoomed in. It might be nice to make the default starting rectangle somewhat inset from the current viewport, perhaps by 20%. That way, it is clear to the user what the rectangle is and what it is for.

### LINKED ISSUE #443: Contrast setting of snapshots
See [this comment](https://github.com/Kitware/UPennContrast/pull/439#issuecomment-1590815424)

bruyeret:
Close different issues related to Snapshots.

For #443, when a layer of the snapshot is not in the current layers or when the channel have changed, a modal asks the user which action he wants to take. Is it clear enough how it is currently done @arjunrajlab ?

Close #440 
Close #441 
Close #443

### DIFF
--- a/src/components/Snapshots.vue
+++ b/src/components/Snapshots.vue
@@ -1,278 +1,310 @@
 <template>
-  <v-card v-if="store.configuration">
-    <v-card-title class="headline">
-      Snapshot
-    </v-card-title>
+  <div>
+    <v-card v-if="store.configuration">
+      <v-card-title class="headline">
+        Snapshot
+      </v-card-title>
 
-    <v-dialog v-model="imageTooBigDialog">
-      <v-alert class="ma-0" type="error">
-        <div class="title">
-          Image can't be downloaded
-        </div>
-        <div class="ma-2">
-          Image size can't exceed {{ maxPixels }} pixels.<br />
-          When downloading raw channels, subsampling is not allowed.<br />
-          Downloading layers allows subsampling image to {{ maxPixels }} pixels
-          if the image is too big.
-        </div>
-        <div class="d-flex">
-          <v-spacer />
-          <v-btn @click="imageTooBigDialog = false">OK</v-btn>
+      <v-dialog v-model="imageTooBigDialog">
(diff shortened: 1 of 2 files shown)


src/components/Snapshots.vue
```diff
@@ -1,268 +1,310 @@
...
+          </v-card-text>
+          <v-card-actions class="d-flex justify-end">
+            <v-btn @click="overwriteConfigurationLayers" color="red">
+              Overwrite configuration layers
+            </v-btn>
+            <v-btn @click="changeDatasetViewContrasts" color="secondary">
+              Try to apply contrasts anyway
+            </v-btn>
+            <v-btn @click="layersOverwritePanel = false" color="primary">
+              Don't change layers
```
luciemac, line 301: I think, you should use 'do not' instead of 'don't' here

src/components/Snapshots.vue
```diff
@@ -366,6 +415,15 @@ export default class Snapshots extends Vue {
   exportChannel: "all" | number = "all";
   format: string = "png";
 
+  layersOverwritePanel: boolean = false;
+  overwrittingLayers: IDisplayLayer[] = [];
+
+  mounted() {
+    this.isSaveSnapshotValid = true;
+    // const formElem = this.$refs.saveSnapshotForm;
```
luciemac, line 423: to remove ?

arjunrajlab:
Looks great! One thing: the Snapshot should keep track of which layers are on vs. off. Sometimes users want to show a layer in one snapshot but not another.

luciemac:
LGTM

arjunrajlab:
Looks good to me, too!

PR #460, opened 2023-07-06
Make guesses of variable immutable only with nd2

### LINKED ISSUE #448: File parsing with a single multi-page TIFF doesn't allow for changing the variable
Take for instance loading a single multi-page TIFF. Here is an example:

MS_1.tiff

https://www.dropbox.com/scl/fi/1ght7mp7582vn97w6ppew/MS2_1.tif?dl=0&rlkey=fi4944g03hum38ioe7m2x7av6

It's a time series, but when it loads, it interprets it as a z series. Which is the right default behavior, but there is no option to "clear" the variable and reassign it to something else, like T.

image

bruyeret:
Also make better typescript

Close #448

### DIFF
--- a/src/views/dataset/MultiSourceConfiguration.vue
+++ b/src/views/dataset/MultiSourceConfiguration.vue
@@ -107,18 +107,36 @@ interface IFileSourceData {
 
 interface IFilenameSourceData extends IVariableGuess {}
 
-interface IAssignmentOption {
+interface IBaseAssignmentOption {
   id: number;
   guess: TDimensions; // Guessed dimension
   size: number; // Number of elements on this dimension
-  data: IFileSourceData | IFilenameSourceData | null; // To compute which image should be taken from the tiles
   name: string; // Displayed name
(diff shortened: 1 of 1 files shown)


arjunrajlab:
Looks good to me. Although other file types (Leica, Zeiss) may also specify the metadata explicitly, like ND2. But let's just go with this for now because I don't have other files for testing. So good to merge!

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -455,17 +486,31 @@ export default class MultiSourceConfiguration extends Vue {
     );
   }
 
+  isAssignmentImmutable(assignment: IAssignment) {
+    // Assignemnt is immutable when it comes from the metadata of an nd2 file
+    const value = assignment.value;
+    if (!(value.source === Sources.File)) {
+      value;
```
luciemac, line 493: ??

luciemac:
LGTM

PR #457, opened 2023-07-05
Undo redo

bruyeret:
Add Undo/Redo feature
Add an history to the database which contains history entries used to undo/redo recordable actions.

Mechanism to record an API call:
- All models in the database inherit `ProxiedAccessControlledModel` (except for History)
- The endpoint method is decorated with `@recordable(...)` (and eventually `@cacheBodyJson`)
- The recordable decorator first call a girder event starting the recording for all `ProxiedAccessControlledModel`
- It calls the original endpoint function ; the calls of the endpoint function to models method (`save`, `remove`, ...) are recorded by the `ProxiedAccessControlledModel`
- It calls a girder event stoping the recording for all `ProxiedAccessControlledModel` and gathering the records
- It adds the history entry to the database

Close #412

PR #461, opened 2023-07-06
Fetch property values when config changes

### LINKED ISSUE #455: Calculated properties disappear when changing datasets
If you calculate property values in one dataset, then switch to a different dataset and compute the same properties, then go back to the original dataset, the property values are gone. See video.

https://github.com/Kitware/UPennContrast/assets/13998824/dfe3c625-469f-48d1-89c6-40a6944b35de

bruyeret:
Must be merged after #457 

Fetch properties when configuration changes
Fetch propertyValues when dataset changes
Move the watchers up in the component tree

Close #455


luciemac:
LGTM

arjunrajlab:
Confirmed, it works! Thanks!

PR #469, opened 2023-07-12
Adjust docker to use the latest master of large_image.

manthey:
We were inconsistently using master for large_image, but the versioned packages for all of the sources.  We should do one or the other; this uses the master.  Otherwise, we should have just pip installed large_image[all] (and girder[mount]) without cloning the large_image repo.

Note, currently using large_image master enables image conversion on the local instance rather than using the worker.  This will make it easier to test transcoding images on a deployed instance to see if that results in substantial speed gains.


bruyeret (approved):
I just tested and it works flawlessly 👍

PR #473, opened 2023-07-19
Worker properties send dicts

### LINKED ISSUE #451: Return tuple from property workers
It would be helpful if property workers could return "tuple" output. For instance, right now, we need to have separate workers to return the x and y coordinates for the centroid of a blob. It would be good to have a single "centroid" worker that returns "centroid.x" and "centroid.y".

In order to specify what the return parameter is, we suggest that we specify that with a label to the docker image. I.e., you would have

--label "outputKeys=key1, key2, key3"

Then the return value would be a dictionary with keys key1, key2, key3. That way, the interface would know that there are multiple return keys and will parse it appropriately. It might be even better if the interface could automatically detect that the return value is a dictionary and automatically pull out the keys, but I'm not sure how complex that would be.

### LINKED ISSUE #470: Bug in column ordering for property values in annotation table
To show a property in the annotation list, you click "List property value". Depending on the order in which you click that checkbox for multiple properties, the numbers (or the column header) seem to get assigned improperly. See in the screenshots below that the numbers get assigned to different columns even though the UI state appears to be the same.

image

image

bruyeret:
## Add possibility to use dict as property values

Remove the restriction on NaN values for properties (too tedious to check and should be done by the client, not the server)

Properties can have "sub-properties"
The architecture enables the use of dict of dict without depth limitation
Keys in the dict can be any valid mongoDB field (see [mongodb field names restrictions](https://www.mongodb.com/docs/v3.2/reference/limits/#Restrictions-on-Field-Names))

For the annotation client:
- addAnnotationPropertyValues is retrocompatible and accept more types for the "value" argument
- getPropertyHistogram is also retrocompatible, the propertyPath is more complex than before (see docs)

The server Docker image needs to be rebuilt (`docker-compose build girder`).

### Before this PR is merged

To be able to send dict property values using the ImageAnalysis project docker images:
- if you use `addAnnotationPropertyValues`, it should be fine with the cached version of the annotation client
- if you use `add_annotation_property_values`, you should use the new version of the annotation client by adding these two lines right after `WORKDIR /UPennContrast`:
```
RUN git fetch --all
RUN git switch worker-tuples
```

### After this PR is merged

To use the new annotation client, rebuild the docker images using `--no-cache` flag so that it clones the new version of the repo instead of using the cached one.

Close #451 
Close #470


arjunrajlab:
@bruyeret Looks great! Minor issue:

I made a property with a few dictionary elements:

image

The problem is that in the annotation list, it shows the "parent" property as a column. Like here, it shows cell "Blob Blob metrics" as a column, but there is no actual property for that "parent".

image

bruyeret:
I added some checks on properties and removed the "show property when created" feature as it is impossible to know which paths are valid before computing the property
What do you think @arjunrajlab? Ready to merge?

arjunrajlab:
@bruyeret Sorry for the delay, just tested again, looks great! Ready to merge!

luciemac:
LGTM

PR #476, opened 2023-07-21
Add options to delete or rename folders/items

### LINKED ISSUE #446: Rename and delete folders
Can we have the ability to rename and delete folders?

bruyeret:
Factorize the code
Actions on files reload the file list

Close #446


arjunrajlab:
@bruyeret When I rename a file, like here where I added "AAAAA_" to the beginning of the filename, it doesn't update the "Recently viewed" area:

image

image

Also, the "max merge" file doesn't appear to get renamed (although it still seems to work):

image

arjunrajlab:
Also:

After running below operation

image

The popup menu stays open somewhat:

image

Also, the "Selected items actions" button has the wrong color in light mode:

image

luciemac:
LGTM

bruyeret:
I don't have the typescript error of the CI in my VS Code, do you have it @luciemac?

bruyeret:
I rebased on master because of the big UI changes
The issues you mentioned should be fixed now @arjunrajlab
Except for the max merge file, but if it doesn't cause problems I think we should leave it as it is, what do you think?

arjunrajlab:
I agree, fine to merge!

PR #478, opened 2023-07-24
Add auto tag of annotations

### LINKED ISSUE #454: Automatically generate a default tag for annotations
It would be nice to have a default tag for annotations. It can be deleted by the user and replaced if needed, but good to have something there be default. I think a good default tag is [_layer shape_]. In the case below, it would be [cy point]. It would be the same tag whether it a manual or automated, which I think is the right behavior.

image

bruyeret:
@arjunrajlab
I added an automatic tag but the auto naming of the annotation becomes redundant
What do you think?

Close #454

### DIFF
--- a/src/tools/creation/ToolCreation.vue
+++ b/src/tools/creation/ToolCreation.vue
@@ -152,21 +152,12 @@ export default class ToolCreation extends Vue {
     const toolNameStrings: string[] = [];
     if (this.toolValues?.annotation) {
       toolNameStrings.push(this.toolValues.annotation.tags.join(", "));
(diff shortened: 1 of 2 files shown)


luciemac:
LGTM

arjunrajlab:
Ah, I see what you mean (screenshot is I assume what you are talking about). I would say let's change the autoname to just:
"[tag] (manual|automated)"

Like, in the below, it would be "Brightfield point (manual)".

image

bruyeret:
I took your feedback into account @arjunrajlab 
It should be ready to merge now

luciemac:
LGTM

PR #479, opened 2023-07-24
Various minor UI updates

arjunrajlab:
Made a view UI updates, moving around some items and changing some behaviors to streamline the app. Also, added a new icon :).


src/components/AnnotationViewer.vue
```diff
@@ -394,15 +394,36 @@ export default class AnnotationViewer extends Vue {
...
+                    // Idea here: keep up to 2 decimal points or 3 significant digits, whichever is more precise
+                    // Not entirely sure the logic is sound, but close enough.
+                    let fixedValue = numValue.toFixed(2);  // preserve up to 2 decimal points
+                    let precisedValue = numValue.toPrecision(3);  // preserve up to 3 significant digits
+                    value = parseFloat(fixedValue) < parseFloat(precisedValue) ? fixedValue : precisedValue;
+                  }
+                  return `${property.name}=${value}`;
+                } else {
+                  return "";
+                }
```
bruyeret, line 426: You can do this instead, `value` can only be a `number` or `undefined`:

```js
const value = this.propertyValues[annotation.id]?.[propertyId];
if (value === undefined) {
  return "";
}
const fixedValue = value.toFixed(2);
const fixedError = Math.abs(value - parseFloat(fixedValue));
const precisedValue = value.toPrecision(3);
const precisedError = Math.abs(
  value - parseFloat(precisedValue)
);
const stringValue =
  fixedError < precisedError ? fixedValue : precisedValue;
return `${property.name}=${stringValue}`;
```

src/components/SettingsPanel.vue
```diff
@@ -16,9 +17,11 @@
 import { Vue, Component } from "vue-property-decorator";
 import AnnotationToggles from "@/components/AnnotationBrowser/AnnotationToggles.vue";
 import ViewerSettings from "@/components/ViewerSettings.vue";
+import UISettings from "@/components/UISettings.vue";
+import UserProfileSettings from "@/components/UserProfileSettings.vue";
 
-@Component({ components: { AnnotationToggles, ViewerSettings } })
+@Component({ components: { AnnotationToggles, ViewerSettings, UISettings, UserProfileSettings } })
```
bruyeret, line 23: You can rename `UISettings` to `UiSettings` to avoid weird `<u-i-settings />` above
Also I think that you don't need to import `UserProfileSettings`

src/App.vue
```diff
@@ -1,36 +1,28 @@
...
         color="primary"
         class="ml-4"
         :to="{ name: 'newdataset' }"
         :disabled="!store.isLoggedIn"
+        title="Upload a new dataset"
       >
         Upload Data
       </v-btn>
-      <user-menu class="ml-4" />
-      <v-btn text icon class="ml-4" @click="dark = !dark">
```
bruyeret, line 18: You can remove the getter and the setter for `dark`

src/components/UISettings.vue
```diff
@@ -0,0 +1,37 @@
...
+    </v-expansion-panel-content>
+  </v-expansion-panel>
+</template>
+
+<script lang="ts">
+import { Vue, Component } from "vue-property-decorator";
+import store from "@/store/index";
+
+@Component
+export default class InterfaceSettings extends Vue {
```
bruyeret, line 26: Use the same name for the class and the file name

src/components/UserProfileSettings.vue
```diff
@@ -0,0 +1,47 @@
...
+    </v-card-title>
+    <v-card-text>
+      <v-container>
+        <div>
+          <h3>Username:</h3>
+          <p>{{ store.userName }}</p>
+        </div>
+        <div>
+          <h3>Email:</h3>
+          <!-- Add email from store -->
```
bruyeret, line 14: Is it something you want to do before merging?

src/components/ValueSlider.vue
```diff
@@ -1,43 +1,62 @@
...
+.label-column {
+  transform: translateY(2px); /* Adjust this value to move the label vertically */
+}
+
+.slider-column {
+  padding-left: 0px; /* Adjust this value to reduce padding around the slider */
+}
+
+.text-field-column {
+  transform: translateY(-4px); /* Adjust this value to move the text input field vertically */
```
bruyeret, line 55: You can keep the comments but remove the `Adjust this value to` at the beginning to make it more concise

src/components/ViewerToolbar.vue
```diff
@@ -63,12 +64,21 @@
...
     </div>
+    
   </div>
 </template>
+
+<style lang="scss" scoped>
+.my-checkbox::v-deep .v-input__control {
+  transform: scale(0.9) translateY(5%);
+}
+</style>
```
bruyeret, line 80: Instead of creating another block for style, you can use the existing one

bruyeret (changes requested):
Good work! Are you sure you never used Vue before?

I made all the feedback which were relevant to me
An important note: if you use VSCode, it looks like you don't use the eslint extension
This extension flags syntax errors, type errors, and format errors
For example, this is what I see in a file that you modified:
image
After installing the extension, you can browse all the files that you modified in this PR and `ctrl+s` in each of them, it should fix formatting
If you still have errors, you will have to fix them manually
There are a few files with errors because the rules changed during the project but the new changes should not introduce eslint errors (only fix them)

If you hover the red or orange lines, it shows you what the problem is
Also, for the format errors, you can configure VSCode to autoformat on save

Also, you should use `const` in general instead of `let`, and use `let` only when you need to change the value of a variable
Don't reassign a variable if the new assignment doesn't have the same meaning
For example in the picture above, `value` was the property that you get from the `propertyValues` and in the if statement, it looks like it becomes a formated string approximating this number value

src/components/ViewerToolbar.vue
arjunrajlab: Thanks!

src/components/AnnotationViewer.vue
arjunrajlab: Done, thanks!

src/App.vue
```diff
@@ -242,3 +241,10 @@ body > div {
   z-index: 100;
 }
 </style>
+<style lang="scss" scoped>
+.logo-icon {
+  height: 50px; // Adjust as needed
```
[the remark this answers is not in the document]
arjunrajlab: Sorry, that was from GPT!

src/components/UserProfileSettings.vue
arjunrajlab: Yes, although I don't know how to access the email currently.

arjunrajlab: I'll just remove for now.

src/components/ValueSlider.vue
arjunrajlab: Thanks!

src/tools/toolsets/Toolset.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <v-expansion-panels v-model="panels">
     <v-expansion-panel expand v-model="panels">
-      <v-expansion-panel-header class="pa-4">
-        <v-toolbar-title> Annotation Tools </v-toolbar-title>
+      <v-expansion-panel-header class="pa-2">
+        <v-toolbar-title :style="{fontSize: '16px'}"> Make objects </v-toolbar-title>
```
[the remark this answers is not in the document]
arjunrajlab: Makes sense, fixed!

arjunrajlab:
OK to review again

src/App.vue
bruyeret: Sorry, I meant that you removed the only part of the code that uses the variable "dark"
So you can remove the function `get dark` (the getter of dark) and `set dark` (the setter of dark) in this file
`dark` is a special variable that is computed, you can use it as a normal variable but each time you need to get its value, the getter is used (with caching) and each time you need to set it, the setter is used

.gitignore
```diff
@@ -1,4 +1,5 @@
 .DS_Store
+package-lock.json
```
bruyeret, line 2: The package-lock.json shouldn't be in the `.gitignore`, as we need to have it in the repo
Instead you can fix the commit which edits `package-lock.json` using an interactive rebase
Or I can do it for you if you prefer?

src/App.vue
```diff
@@ -242,3 +245,10 @@ body > div {
   z-index: 100;
 }
 </style>
+<style lang="scss" scoped>
+.logo-icon {
+  height: 50px;
+
+  margin-top: 10px;
+}
+</style>
```
bruyeret, line 254: I didn't notice during my first review, but you can also merge the two `<style>` blocks

bruyeret (commented):
Thanks for all the changes!
You can `ctrl+s` to auto format these files: `UISettings`, `AnnotationToggles`, `UserMenu`
If you need help with git, feel free to ask me

src/App.vue
arjunrajlab: Done!

.gitignore
arjunrajlab: Uhhhh, I think you better do that :), I don't think I can figure that out without messing something up!

package-lock.json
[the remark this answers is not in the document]
arjunrajlab: Unfortunately, I'm not sure how to remove it from the PR… help! :)

src/App.vue
arjunrajlab: Ahhh, okay, I see, I didn't see those functions in there. Deleted!

bruyeret (approved):

bruyeret:
I rolled back the `package-lock.json` and the `.gitignore` and rebased on master
We should merge this PR as soon as possible because it changes a lot of files and it makes the chances of conflict higher
Tell me when you are done with this PR @arjunrajlab, I will review it, merge it and you can create a new branch from the updated master right away

arjunrajlab:
It can be merged @bruyeret !

bruyeret:
Thanks, I review the last commits ASAP!

PR #480, opened 2023-07-24
Fix NaN values when typing in width/height of Snapshot

### LINKED ISSUE #453: Text fields for controlling snapshots are not working properly
In snapshots, when you delete “Width”, then the “Left” field become blank. Similarly, when you delete “Height”, “Top” become blank. When either of the field become blank this way, the selection frame disappears. It can be reset by clicking “set frame to maximum/viewport". See video:

https://github.com/Kitware/UPennContrast/assets/13998824/57e003c9-b667-40c1-a7cf-dc7cec8951d7

bruyeret:
Close #453

### DIFF
--- a/src/components/Snapshots.vue
+++ b/src/components/Snapshots.vue
@@ -349,6 +349,11 @@ interface IDownloadParameters {
   tiffCompression?: string;
 }
 
+function intFromString(value: string) {
+  const parsedValue = parseInt("0" + value, 10);
+  return Number.isNaN(parsedValue) ? 0 : parsedValue;
+}
+
 @Component({ components: { TagPicker } })
 export default class Snapshots extends Vue {
   readonly store = store;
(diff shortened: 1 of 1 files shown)


src/components/Snapshots.vue
```diff
@@ -446,16 +446,20 @@ export default class Snapshots extends Vue {
...
   }
 
   get bboxHeight(): number {
     return (this.bboxBottom || 0) - (this.bboxTop || 0);
   }
 
-  set bboxHeight(value: number) {
-    this.bboxBottom = (this.bboxTop || 0) + parseInt("" + value, 0);
+  set bboxHeight(value: string) {
+    const parsedValue = parseInt("0" + value, 10);
```
luciemac, line 460: Can you factorize this part, please?

bruyeret: Of course!

arjunrajlab:
Looks good to me1

src/components/Snapshots.vue
```diff
@@ -349,6 +349,11 @@ interface IDownloadParameters {
   tiffCompression?: string;
 }
 
+function intFromString(value: string) {
```
luciemac, line 352: LGTM.
In my opinion, it would be great to have a math helper file with such functions, WDYT?

luciemac:
LGTM

src/components/Snapshots.vue
bruyeret: After more thoughts, this function shouldn't be as is in a math helper file, because it makes the assumption that "not a number" should be treated as zero

PR #483, opened 2023-08-01
Add transcoding to tiff file when uploading dataset

### LINKED ISSUE #482: Implement TIFF transcoding for performance gains
For most datasets, I think having the option to transcode the dataset, especially multi-source datasets, to a single optimized TIFF would be ideal. The place to have this in the UI is I think here:

image

I would like to have a checkbox that says "Transcode into optimized TIFF file". It should be selected by default for all files except ND2 files, which are already fairly optimized.

Here's the note from David on how to do it:
The endpoint for the "un-large-image" is `DELETE /item/{id}/tiles` and the redo is `POST /item/{id}/tiles` with `force=true`.

Here is a test dataset to try:

https://www.dropbox.com/scl/fo/1mhmz9ue3hmrjkvj11777/h?rlkey=pv2lhyrqrfcvmsuijkx21x948&dl=0

It's a set of TIFF files that I uploaded all at once as a multi-source with 4 channels of z-stacks. The performance was quite poor but got a lot better after transcoding, so hopefully is a good test case.

bruyeret:
When transcoding, the logs from the server appear below the submit button

Close #482

### DIFF
--- a/src/store/GirderAPI.ts
+++ b/src/store/GirderAPI.ts
@@ -160,12 +160,14 @@ export default class GirderAPI {
     }
   }
 
-  generateTiles(itemId: string) {
-    return this.client.post(`item/${itemId}/tiles`);
+  generateTiles(itemId: string, force: boolean, localJob: boolean) {
+    return this.client.post(`item/${itemId}/tiles`, undefined, {
+      params: { force, localJob }
+    });
   }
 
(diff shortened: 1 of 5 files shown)


arjunrajlab:
Works great! Looks good to merge for me. Would be good to have a progress bar above the text window, but that can wait until we have the notification system in place.

luciemac:
LGTM

Issue #475, opened 2023-07-21
Make some sort of progress/notification system for workers and annotations, etc.


arjunrajlab:
From David, here's a file that may be helpful in implementing based on XML fragments emitted from the worker through stdout:
https://github.com/DigitalSlideArchive/superpixel-classification/blob/main/superpixel_classification/SuperpixelClassification/progress_helper.py

From the user perspective, it doesn't have to be super granular or accurate, but just something to show that something is working. The key interface parts that I've noticed people waiting for have been:
1. Waiting for interface to load up the image/annotations. Usually for large number of annotations.
2. Waiting for tile_caches to compute. There is a progress bar for it already, but it is somewhat unreliable.
3. Waiting for large jobs to compute. So far mostly for making annotations, but also in principle for property workers as well.
4. We haven't implemented yet, but also probably the TIFF transcoding.

I think that would cover the most important cases.

From the developer perspective, as I write my worker, I'd like to have a call that is like:

[annotation_client].sendProgress(title="Cy3 point annotations", info="Finding points", progress=0.33)

The only required parameter would be the progress=0.33, I think. That would be easy for the programmer to use.

PR #485, opened 2023-08-03
Add progress/notifications for property workers

bruyeret:
As in #473, you need to checkout to this branch using:
```
RUN git fetch --all
RUN git switch property-progress
```
in the Dockerfile of the workers to use the sendProgress function.
See [this commit](https://github.com/arjunrajlaboratory/ImageAnalysisProject/commit/beb094c8306cf71f1423ae4771afeafcdf08f977) of [this branch](https://github.com/bruyeret/ImageAnalysisProject/tree/example-branch) to see an example of how to send progress.

No need to rebuild the girder image.

This PR is part of #475

### DIFF
--- a/devops/girder/annotation_client/annotation_client/utils.py
+++ b/devops/girder/annotation_client/annotation_client/utils.py
@@ -1,0 +1,17 @@
+import sys
+import json
+
+def sendProgress(progress, title, info):
+    """
+    Send progress to the front end
+
+    :param float progress: Progress between 0 and 1 for the linear progress bar
+    :param str title: Text to show in bold in the progress bar
+    :param str info: Text to show after the title in the progress bar
+    """
+    print(json.dumps({
+        "progress": progress,
+        "title": title,
+        "info": info
+    }))
+    sys.stdout.flush()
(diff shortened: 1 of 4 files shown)


devops/girder/annotation_client/annotation_client/utils.py
```diff
@@ -0,0 +1,17 @@
+import sys
+import json
+
+def sendProgress(progress, title, info):
```
luciemac, line 4: Is it really necessary as it only prints progress info and the method does not seem to be called ?

bruyeret: The `stdout` of the worker is sent to the front end, this is why printing works 👍 
These method in the `annotation_client` are not called in this repo but they are used in the Workers themselves
For example, see [this commit](https://github.com/arjunrajlaboratory/ImageAnalysisProject/commit/beb094c8306cf71f1423ae4771afeafcdf08f977) which shows how to use it

luciemac: sounds good !

luciemac:
LGTM

PR #487, opened 2023-08-04
Minor UI update.

arjunrajlab:
Just a small update on the worker menu

### DIFF
--- a/src/components/AnnotationWorkerMenu.vue
+++ b/src/components/AnnotationWorkerMenu.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-card v-if="tool" class="menu">
+  <v-card :class="{ menu: true, loaded: !fetchingWorkerInterface }" v-if="tool">
     <v-card-title class="subtitle-1">
       Worker menu
     </v-card-title>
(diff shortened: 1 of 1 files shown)


src/components/AnnotationWorkerMenu.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <v-card v-if="tool" class="menu">
+  <v-card :class="{ menu: true, loaded: !fetchingWorkerInterface }" v-if="tool">
```
luciemac, line 2: I think you can go with 
```
<v-card
  class="menu",
  :class="{ loaded: !fetchingWorkerInterface}"
>
```
as menu is static and always true.

It should be rendered as:
```
<v-card class="menu loaded"> // if fetchingWorkerInterface = false
<v-card class="menu"> // if fetchingWorkerInterface = true
```

arjunrajlab:
Hmm. I tried your suggestion, but then the worker menu didn't appear at all for some reason. ChatGPT offered up the following reason:

The syntax you tried (class="menu", :class="{ loaded: !fetchingWorkerInterface}") is incorrect because it treats class and :class as separate attributes, and the comma between them is invalid syntax in HTML.

So I guess the original is right?

luciemac:
From vue.js [documentation](https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes) it should work, but that is ok !


LGTM

PR #490, opened 2023-08-11
Change when we schedule caching tile_frames and maxmerge

manthey:
We were scheduling these when a file was uploaded, but then converting that item into a different record via the multi-source file.  This largely invalidated the usefulness of the tile_frames and maxmerge, especially if the multi-source file was not a direct pass through.

This moves the scheduling of these to after the multi-source file becomes a large image or is transcoded into a tiff.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -675,7 +675,7 @@ export class Main extends VuexModule {
     parentId: string;
     metadata: string;
     transcode: boolean;
-    eventCallback?: (data: IJobEventData) => void;
+    eventCallback?: (data: IJobEventData | null) => void;
   }) {
     try {
       sync.setSaving(true);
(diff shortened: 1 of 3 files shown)


arjunrajlab:
This definitely helps, but I was still able to get Girder to crash by navigating around a lot before all the caches were done computing. It just took a little more navigating before it went down. I wonder if it has something to do with the dask array implementation. Still, it is certainly an improvement. It is interesting that the UI is basically not useful until the tile_frames are done. I wonder if that computation is just eating up a lot of memory somehow.

arjunrajlab:
@bruyeret @manthey I think this can be merged because it is definitely an improvement. Not sure how it will work with the progress bars, though. Also should we do progress bars for the z merge computation?

manthey:
I don't think we show progress from the z-merge computation.  There is a girder job that runs as it generates that; it uses server resources.  If the UI asks for a max-merge before it is done, the server has to compute it on demand.

I'll merge this.

PR #494, opened 2023-08-22
Add annotation fetching progress and green tooltip

### LINKED ISSUE #437: Minor UI tweak: Tile cache UI progress bar
The tile cache UI progress bar could be improved. I would suggest the progress bar says “Optimizing performance” with a percentage underneath. When it’s done it should turn into a green check with “Performance optimized” as a tooltip.

image

### LINKED ISSUE #475: Make some sort of progress/notification system for workers and annotations, etc.

bruyeret:
When downloading annotations (or annotation connections), show this linear progress:

image

When downloading of annotations and caching of dataset is done, show this green alert:

image

Close #437
Close #475

### DIFF
--- a/src/components/ImageViewer.vue
+++ b/src/components/ImageViewer.vue
@@ -32,40 +32,38 @@
         :key="'geojsmap-' + index"
       />
     </div>
-    <div class="loading" v-if="readyLayersCount < readyLayersTotal">
-      <v-progress-linear
-        :value="(100 * readyLayersCount) / readyLayersTotal"
-        style="height: 100%;"
-      >
-        <strong>
-          {{ ((100 * readyLayersCount) / readyLayersTotal).toFixed(1) }}% layers
-          ready {{ readyLayersCount }} / {{ readyLayersTotal }}
-        </strong>
-      </v-progress-linear>
-    </div>
     <div
-      class="loading"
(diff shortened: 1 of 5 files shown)


arjunrajlab:
Looks great, ready to merge!

arjunrajlab (approved):
Looks good to me!

luciemac:
LGTM

PR #495, opened 2023-08-23
Quick select

### LINKED ISSUE #427: Shift click on annotation to select it, shift click and drag to lasso
It would be nice if a shift-click on an annotation would select it (rather than just highlight it). That could be very useful. It could also be used as a toggle: if already selected, then would unselect.

Another nice tool would be if shift click and drag would create a "lasso" that would select all annotations within the list. Same toggle behavior. But if that's too hard, then not as important.

bruyeret:
Select annotations using shift+click or shift+drag

Close #427


zjniu:
Just tested this from my end. Works really well! Let's merge it.

arjunrajlab (approved):
Looks great to me!

luciemac:
LGTM

PR #496, opened 2023-08-23
Add search bar to the custom file manager

### LINKED ISSUE #403: Add search/filter bar for the file browser
It would be helpful to have a search bar for the file browser so people can look for particular files.

bruyeret:
Close #403

### DIFF
--- a/src/components/CustomFileManager.vue
+++ b/src/components/CustomFileManager.vue
@@ -1,5 +1,24 @@
 <template>
   <div>
+    <girder-search @select="searchInput">
+      <template #searchresult="item">
+        {{ renderItem(item) }}
+        <v-icon class="mr-2">{{ iconToMdi(iconFromItem(item)) }}</v-icon>
+        {{ item.name }}
+        <div class="d-flex flex-wrap">
+          <v-chip
+            small
+            v-for="(chipItem, i) in debouncedChipsPerItemId[item._id]"
+            :key="'chip ' + i + ' item ' + item._id"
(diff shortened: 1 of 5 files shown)


arjunrajlab (approved):
Overall, love it! I think it would be nice to have a little text that says "Search" next to the search bar to make it a little more discoverable, because right now it's a little hard to notice (see screenshot). But the functionality is great!

image

src/components/CustomFileManager.vue
```diff
@@ -156,15 +178,41 @@ export default class CustomFileManager extends Vue {
...
+      case "file":
+        return "file";
+      case "folder":
+        return "folder";
+      case "item":
+        return "file";
+      case "user":
+        return "user";
+    }
+  }
```
luciemac, line 210:
```
switch (selectable._modelType) {
      case "file":
      case "item":
        return "file";
      case "folder":
        return "folder";
      case "user":
        return "user";
    }
```

luciemac:
then, LGTM

bruyeret:
There is a loupe icon but it is white even in light mode, I don't know why
The icon to the right of the search bar is dark in light mode
I added the "search" text anyway @arjunrajlab 

image

image

Also to switch between dark mode and light mode, I had to open a dataset now that the button is in the settings panel, shouldn't it be also in home page?

zjniu:
LGTM!

luciemac:
LGTM

PR #497, opened 2023-08-24
Add Papaparse to dependecies, fix csv export

### LINKED ISSUE #471: In CSV, it lists property ID, but would be better to list property name
image

bruyeret:
Close #471


zjniu:
Just tested from my end, LGTM!

luciemac:
LGTM

PR #507, opened 2023-08-31
Add overview panel

bruyeret:
Close #450

PR #509, opened 2023-09-04
Histograms precache

### LINKED ISSUE #502: Precache histograms for improved responsiveness of UI

bruyeret:
To be merged after the overview PR #507 
Close #502

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18323,9 +18323,9 @@
       "dev": true
     },
     "qs": {
-      "version": "6.11.0",
-      "resolved": "https://registry.npmjs.org/qs/-/qs-6.11.0.tgz",
-      "integrity": "sha512-MvjoMCJwEarSbUYk5O+nmoSzSutSsTwF85zcHPQ9OrlFoZOYIjaqBAJIqIXjptyD5vThxGq52Xu/MaJzRkIk4Q==",
+      "version": "6.11.2",
+      "resolved": "https://registry.npmjs.org/qs/-/qs-6.11.2.tgz",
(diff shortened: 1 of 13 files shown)


arjunrajlab:
This is great, definitely shows very nice performance once loaded!

I am getting a lot of CPU usage and girder seems slow even after progress bar says "Optimized and fully loaded". I'm assuming the histogram caching is running then. Is it possible to show a progress bar for that as well? Whatever it is, there is some process running on Girder that seems to take up CPU for a LONG time after and makes Girder a bit sluggish.

Also, the overview panel seems to load from the full image, so only shows once that is loaded. I've noticed now that the tile_cache can load much faster, but doesn't show up in the overview panel. One could consider using the tile_cache for the overview panel, but I don't think it's super important, just a note.

This is on my laptop because I'm at home today (holiday, technically), but I'll also test this on my server at work tomorrow morning.

arjunrajlab:
@bruyeret Generally looks good, but I have a question. Right now, the progress bar shows from the very beginning. Is it actually running at the same time as the tile caching? Or is it scheduled to wait until after the tile_cache to begin computation? If it is scheduled to compute after the tile_cache, is it hard to make the progress bar only start once is actually starts computing?

arjunrajlab (approved):
@bruyeret Generally looks good, but I have a question. Right now, the progress bar shows from the very beginning. Is it actually running at the same time as the tile caching? Or is it scheduled to wait until after the tile_cache to begin computation? If it is scheduled to compute after the tile_cache, is it hard to make the progress bar only start once is actually starts computing? If it is a pain to do, then this is fine as it is!

bruyeret:
It's done! But there is a small time between tile caching and histogram caching when the frontend thinks that everything is cached because histogram caching has not started yet

arjunrajlab:
Looks great to me, thank you!

luciemac:
LGTM

PR #513, opened 2023-09-07
Change style of hovered annotations

### LINKED ISSUE #464: When line annotation is hovered/clicked in the list view, its style doesn't change
When going around in the list view, it doesn't change the style (points and blobs do). Can we make some sort of subtle change when highlighted? It does change styling (thicker line) when selected, which is good. Probably the reason is that the hover style is only for the interior and not the edge. Perhaps a mid-size increase in the line width?

bruyeret:
Hovered style wasn't different from base style for lines
Change stroke width and color when hovering annotations
Close #464

### DIFF
--- a/src/utils/annotation.ts
+++ b/src/utils/annotation.ts
@@ -35,11 +35,15 @@ export function getAnnotationStyleFromLayer(
     style.fillColor = layerColor;
     style.strokeColor = geoColor;
   }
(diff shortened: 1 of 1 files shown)


arjunrajlab (approved):
Works great, thanks!

luciemac:
LGTM

PR #522, opened 2023-09-11
Add image scale settings

### LINKED ISSUE #498: Image scale information and scalebar
It would be very helpful, for both computations and otherwise, to have a dataset/configuration metadata option. Here's what I'm thinking. Somewhere, we would need to have a button for [Configuration information]. This should bring us to the configuration view:

image

In that view, I think we should have the option of pulling in xy pixel size, z size, t step from the metadata, with the option to override or revert to metadata. We will want to have dropdowns that show units, like "nm, µm, mm, m" for length, "ms, s, h, d, m" for time.

Using that, we should be able to show a scale bar that is accurate. Also, this information should be accessible to workers (annotations and properties).

I think it should be set at the configuration level and not the dataset level.

bruyeret:
Add a new panel in the collection view
Make scalebar clickable, open a dialog to tweak scales

I added most of the features of issue 498:
- In the collection view, the user can set the values for the different scales
- In the image viewer, the user can open a dialog by clicking on the scalebar to have more options. In this dialog, the scale is saved in the dataset view and the user has the option to save the scale in collection and to reset to dataset default. It behaves like the contrasts.
- The scalebar uses the pixel size given in the dataset view or in the configuration.
- Workers have access to scales: `params["scales"]["pixelSize"]` contains both `value` and `unit`

Close #498

### DIFF
--- a/src/store/model.ts
+++ b/src/store/model.ts
@@ -146,6 +146,22 @@ export interface ISnapshot {
 
 export type IDimensionCompatibility = "one" | "multiple";
 
+export type TUnitLength = "nm" | "µm" | "mm" | "m";
+export const unitLengthOptions: TUnitLength[] = ["nm", "µm", "mm", "m"];
+export type TUnitTime = "ms" | "s" | "m" | "h" | "d";
+export const unitTimeOptions: TUnitTime[] = ["ms", "s", "m", "h", "d"];
+
+export interface IScaleInformation<TUnit> {
+  value: number;
+  unit: TUnit;
+}
+
+export interface IScales {
+  pixelSize: IScaleInformation<TUnitLength>;
+  zStep: IScaleInformation<TUnitLength>;
+  tStep: IScaleInformation<TUnitTime>;
+}
+
 export interface IDatasetConfigurationBase {
   compatibility: {
     xyDimensions: IDimensionCompatibility;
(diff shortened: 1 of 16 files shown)


arjunrajlab (changes requested):
Looks good! Only thing is the color of the scale bar would be better as white rather than black because it is often very hard to see:

image

bruyeret:
It should be fixed now, do you approve @arjunrajlab ?
image

arjunrajlab:
Is it possible to make the line a little bit thicker looking?

bruyeret:
Yes of course, it has size 2px, and now it has size 4px, do you want something thicker?
image

arjunrajlab:
Let's go for 5pt, that should be fine! Also, we don't need the little tick marks on the two ends, just the bar in the middle is fine. But also no problem if it's hard to change, minor point.

bruyeret:
Does it look good?

image

arjunrajlab:
Beautiful, perfect!

src/store/model.ts
```diff
@@ -843,21 +861,27 @@ export function copyLayerWithoutPrivateAttributes(
...
-    tDimensions: "multiple",
-    channels: {}
-  },
-  layers: [],
-  tools: [],
-  snapshots: [],
-  propertyIds: []
-};
-
+export function exampleConfigurationBase(): IDatasetConfigurationBase {
```
luciemac, line 864: Why don't you keep it as a `const` ?

bruyeret: It is to avoid aliasing

PR #524, opened 2023-09-11
Fix and improve color picker

### LINKED ISSUE #520: Color picker issues/bugs
If the color picker is in light mode, white doesn't show up. I think having a small black circle around the color should solve the problem:

image
image

The Hex code for the color is not editable. I suggest we just get rid of the hex code and just show the color itself.

The double arrow here is not particularly discoverable (I didn't know what it did until just now). I thought it was for adjusting the numbers because of the arrows, but was surprised to see I could do HSV or Hex there. Maybe we can replace with a hamburger or "…" button?

image

Also, currently, if you go around in the color picker, there is no way to pick white, which is probably one of the most commonly used options:

image

Is there a more appropriate color picker we can use?

bruyeret:
Close #520

### DIFF
--- a/src/components/DisplayLayer.vue
+++ b/src/components/DisplayLayer.vue
@@ -63,36 +63,28 @@
         :histogram="histogram"
       />
       <v-menu
-        ref="colorMenu"
-        v-model="showColorPicker"
         :close-on-content-click="false"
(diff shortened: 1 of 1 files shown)


luciemac:
LGTM

PR #526, opened 2023-09-11
Fix bound checks when typing contrast values

### LINKED ISSUE #521: "Max" field not working right in absolute contrast adjustment mode
The "max" field is hard to adjust by manually typing in the editable number field:

image

Video:

https://github.com/Kitware/UPennContrast/assets/13998824/8657f058-7ed2-4470-bc00-d93173552990

I think the issue is that if the bottom limit is, say, 6800, then I start typing "12000", it doesn't like to start with 1, because if I type 1, it think that number is < 6800 and so it has to automatically make it start with 6800. One solution would be to only allow it to update the variable when the user hits enter/return, and only do the bounds checking then.

bruyeret:
Close #521

### DIFF
--- a/src/components/ContrastHistogram.vue
+++ b/src/components/ContrastHistogram.vue
@@ -54,18 +54,16 @@
     <div class="sub">
       <v-text-field
         type="number"
-        :min="editMin"
-        :max="editMax"
         v-model="editBlackPoint"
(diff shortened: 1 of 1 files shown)


luciemac:
LGTM

PR #527, opened 2023-09-11
Default "useAutoTags" to false for edit mode

### LINKED ISSUE #525: Auto-tag bug
We have noticed that the auto-tag creates a bug. Right now, if you remove the automatic tag, it will reinsert it when you edit the tool. Here's what I mean:

image

Now I edit it:
image

But now when I click the pencil icon here:
image

It shows the default auto-tag name, even though the original tag is still associated with it:

image

bruyeret:
Close #525


arjunrajlab (approved):
Looks great!

luciemac:
LGTM

PR #529, opened 2023-09-11
Fix minor bugs at login

bruyeret:
This doesn't close any issue
This PR fixes bugs that occurs at login
You don't need to review @arjunrajlab, but you can see that when logged out the file manager was always there


arjunrajlab:
Nice, thank you!

luciemac:
LGTM

PR #532, opened 2023-09-12
Add current frame filter

### LINKED ISSUE #499: Add option to restrict/filter annotation list to show only those in current frame only
Here, where we have the filtering options, it might be nice to have a toggle for "Current frame only" that filters annotations in the list to only show those in the current frame. The toggle would be here:

image

bruyeret:
Close #499

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationFilters.vue
+++ b/src/components/AnnotationBrowser/AnnotationFilters.vue
@@ -6,6 +6,13 @@
       v-model="shapeFilter"
     />
     <tag-filter-editor class="filter-element" v-model="tagFilter" />
+    <v-switch
+      class="filter-element"
+      label="Current frame only"
+      v-model="onlyCurrentFrame"
+      dense
+      hide-details
+    />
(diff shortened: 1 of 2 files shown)


src/components/AnnotationBrowser/AnnotationFilters.vue
```diff
@@ -58,6 +65,14 @@ export default class AnnotationFilters extends Vue {
     this.filterStore.setShapeFilter(filter);
   }
 
+  get currentFrameFilter() {
```
luciemac, line 68: I am no sure about this variable name.
What do you think of something like "onlyCurrentFrame"?
In my opinion, it does not feel right to have a variable named "xxxFilter" and typed as a boolean..

bruyeret: Yes indeed, it was bad naming!

luciemac:
LGTM

arjunrajlab (approved):
Works great, LGTM!

PR #533, opened 2023-09-12
Fix getAnnotationsByDatasetId when no shape is used

### LINKED ISSUE #530: Strange pattern in annotations.py
In the file:

https://github.com/Kitware/UPennContrast/blob/master/devops/girder/annotation_client/annotation_client/annotations.py

There is this code snippet
```
    def getAnnotationsByDatasetId(self, datasetId, shape=None, limit=50, offset=0):
        """
        Get the list of all annotations in the specified dataset

        :param str datasetId: The dataset's id
        :param str shape: optional filter by shape
        :return: A list of annotations
        """
        url = PATHS['annotation_by_dataset'].format(datasetId=datasetId)

        if shape:
            url = '{url}&shape={shape}&limit={limit}&offset{offset}'.format(url=url, shape=shape, limit=limit, offset=offset)

        return self.client.get(url)
```
In that code, it acts a little bit weirdly because if you don't specify shape at all, it gives you 50 annotations of any shape. Ideally, the behavior would be to give you all annotations up to the given limit if you don't specify shape. For instance, right now, this code:

`allAnnotationList = annotationClient.getAnnotationsByDatasetId(datasetId, limit = 1000000)`

only gives you 50 annotations and not the full list.

bruyeret:
It was indeed a bug!
It should be fixed now

Close #530

### DIFF
--- a/devops/girder/annotation_client/annotation_client/annotations.py
+++ b/devops/girder/annotation_client/annotation_client/annotations.py
@@ -37,7 +37,7 @@ class UPennContrastAnnotationClient:
 
     # Annotations
 
-    def getAnnotationsByDatasetId(self, datasetId, shape=None, limit=50, offset=0):
+    def getAnnotationsByDatasetId(self, datasetId, shape=None, tags=None, limit=1_000_000, offset=0):
         """
         Get the list of all annotations in the specified dataset
 
@@ -50,6 +50,8 @@ class UPennContrastAnnotationClient:
         url = f'{url}&limit={limit}&offset{offset}'
         if shape:
             url = f'{url}&shape={shape}'
+        if tags:
+            url = f'{url}&tags={tags}'
 
         return self.client.get(url)
 
--- a/devops/girder/annotation_client/annotation_client/annotations.py
+++ b/devops/girder/annotation_client/annotation_client/annotations.py
@@ -47,8 +47,9 @@ class UPennContrastAnnotationClient:
         """
         url = PATHS['annotation_by_dataset'].format(datasetId=datasetId)
 
+        url = f'{url}&limit={limit}&offset{offset}'
         if shape:
-            url = '{url}&shape={shape}&limit={limit}&offset{offset}'.format(url=url, shape=shape, limit=limit, offset=offset)
+            url = f'{url}&shape={shape}'
 
         return self.client.get(url)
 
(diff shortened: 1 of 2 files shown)


devops/girder/annotation_client/annotation_client/annotations.py
```diff
@@ -46,8 +46,9 @@ def getAnnotationsByDatasetId(self, datasetId, shape=None, limit=50, offset=0):
         """
         url = PATHS['annotation_by_dataset'].format(datasetId=datasetId)
 
+        url = f'{url}&limit={limit}&offset{offset}'
         if shape:
-            url = '{url}&shape={shape}&limit={limit}&offset{offset}'.format(url=url, shape=shape, limit=limit, offset=offset)
+            url = f'{url}&shape={shape}'
```
luciemac, line 51: I think you should add a else case here

bruyeret: What should I put in the else case?

arjunrajlab: I think that if shape is not given, it should just return all annotations (subject to limit and offset).
Also, while we are at it, is there a way to add a filter by tag? Like, to get all annotations by tag? That would be useful.
Also, let's change default limit to 1,000,000. I think that is a reasonable number. 50 is pretty small.

bruyeret: Returning all annotations and taking limit and offset into account is now the default behavior when no shape is given
I can add the feature to filter by tags and increase the default limit
In the tag filtering, what would be the algorithm?
- Get all annotations with exactly the given tags
- Get all annotations with all the given tags and eventually other tags
- Get all annotations with at least one of the given tags

luciemac:
LGTM

devops/girder/annotation_client/annotation_client/annotations.py
arjunrajlab: Yeah, I wasn't sure exactly what to do there :). I think it should be all of the given tags and potentially other additional tags.

bruyeret:
I added the tags filter to the endpoint and propagated the changes to the annotation client

arjunrajlab:
I will try and test it later today or tomorrow!

bruyeret:
I forgot to mention that girder has to be rebuilt

arjunrajlab (approved):
Looks good to me, ready to merge!

PR #534, opened 2023-09-12
Add point annotation scaling options

### LINKED ISSUE #505: Point annotation display options
In Settings -> Viewer Settings, it would be good to have two options for point annotations:

1. A toggle for "Scale with zoom" that either keeps the point annotation circle marker the same size no matter the zoom level, or scales it by the zoom level.
2. A slider that sets the size of the point annotation circle marker.

bruyeret:
Close #505


luciemac:
LGTM

PR #537, opened 2023-09-14
Revert "Make dataset updates less frequent"

### LINKED ISSUE #536: "Unroll" not working anymore
I'm not sure which update broke this functionality, but if you check "Unroll" for XY, Z, or T, it doesn't actually do the unroll. Instead, it seems to just show annotations from all frames. Very strange bug. I think I tested unroll with the minimap functionality and it worked, so it must be in some update after that.

bruyeret:
This reverts commit da686bb70c8eca20d0a167ce36e564809b9f7304.

Close #536

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -364,9 +364,6 @@ export class Main extends VuexModule {
     id: string | null;
     data: IDataset | null;
   }) {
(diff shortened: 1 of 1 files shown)


arjunrajlab (approved):
Works great, thanks!

luciemac:
LGTM

PR #545, opened 2023-09-19
Increase z-index of menus to put them in front of images

### LINKED ISSUE #540: Clicking on the color picker goes through to the image view
When you click on the color picker, it goes through to the view, so you can't actually interact with the color picker. Somehow this is not a bug when the server is on my Mac, but it is if the server is Linux. That's why I didn't notice it before. If the server is Linux and the client is Mac, the issue is still there.

image

bruyeret:
Close #540

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -263,4 +263,8 @@ body > div {
 
   margin-top: 10px;
 }
+
+.v-menu__content {
+  z-index: 10000 !important;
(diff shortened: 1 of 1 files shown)


luciemac:
LGTM

arjunrajlab (approved):
Fixed the issue, thanks!

PR #555, opened 2023-09-25
Updated Snapshots with minor UI edits

arjunrajlab:
Minor edits to wording and spacing.

PR #569, opened 2023-09-28
Fix snapshots error

bruyeret:
Some errors I didn't notice from PR #555 

`justify-left` doesn't exist, it should be `justify-start` (see errors in the console)
eslint doesn't accept two newlines one after each other


arjunrajlab (approved):
Thanks, looks good to me!

luciemac:
LGTM

PR #572, opened 2023-10-04
File manager reload items and progress

### LINKED ISSUE #556: In the file manager, if I do Actions->Delete, the list doesn't update
If I click on the Delete action for an item in the file list:

image

It deletes the file, but requires a page refresh to show that. It should probably update the list after deletion.

bruyeret:
Close #556


luciemac:
LGTM

PR #574, opened 2023-10-05
Compositing options

### LINKED ISSUE #504: Add different compositing options
Should be added to Settings -> Viewer options.

bruyeret:
The base composition modes are the one that work with images that have a black background
The advanced ones sometimes kind of work
Close #504


luciemac:
LGTM

PR #577, opened 2023-10-13
Fix networking mode for worker.

manthey:
The worker needs to run in the same network as the rest of the docker containers OR girder must be publicly visible to the worker and to girder_worker both.  By not specifying the network mode of the worker, girder_worker assigns it the same network as girder_worker's container. By specifying http://girder:8080/api/v1 as the alternative worker URL, this is reachable from within that network.


arjunrajlab:
Works, thank you!! I have not yet tested with cancel-workers, but can rebase and try that once this change is merged.

PR #582, opened 2023-10-18
New file parser

arjunrajlab:
Created new file parser. This should help parse many different filename patterns. It uses a new algorithm for determining the relevant filename tokens and takes a reasonable guess at what they might represent. I tested with a couple different datasets and it seems to work.


luciemac:
LGTM

arjunrajlab (commented):
Overall, the refactoring looks great, sorry for being so sloppy :).

The algorithm seems to work well, but for whatever reason, it seems to fail on this set of files:
https://www.dropbox.com/sh/ary1jz01q6e9on7/AACw-N-LOjigBrlR05_hgw_da?dl=0
I tried this set of files on the raw algorithm itself, and it does parse properly, so I don't know what is going on. I am wondering if some files are not getting uploaded or put into the list or something? Is there some limit on the number of files?

arjunrajlab:
Like, if you take just half the files, or specific subsets, it seems to work, but if you take all of them, it somehow doesn't seem to load the variables properly in the assignment page.

arjunrajlab:
@bruyeret are you okay for merging?

bruyeret (approved):

PR #583, opened 2023-10-19
Home panels

### LINKED ISSUE #558: Rearrange panels on home page, add scrollbars
I like this suggested arrangement of panels on the home page. Reminder to add scroll bars to the panels for recent and browse.

image

Also, in the "Recent dataset" panel, it shows the configuration as some small text. Perhaps we can make that text clickable? I think we should not make it a whole chip, though, because the chips are kind of loud.

bruyeret:
Refactor Home Panels
Add the quickupload feature
Add a custom dropzone component

Close #558


luciemac:
LGTM

PR #586, opened 2023-10-24
Enhance visibility of code blocks and quickupload progress

### LINKED ISSUE #584: On small screens, the image conversion panel is not immediately visible
Here's what the panel looks like:

image

Sometimes it is completely off screen, so it's hard to know something is happening. It might also be nice to have a spinner or something going in the meantime to provide some more feedback, because the console output is a bit cryptic to most users.

bruyeret:
Close #584


luciemac:
LGTM

PR #588, opened 2023-10-27
Add a quick import progress

### LINKED ISSUE #585: Progress bar for quick import
It might be nice to have an overall progress bar at the top of the screen while the "quick import" is going. That way, the user knows that there is a process going automatically so they know what is happening.

bruyeret:
It takes all steps into account and uses a weight per step to account for the inequal time taken per step

Close #585

### DIFF
--- a/src/components/GirderLocationChooser.vue
+++ b/src/components/GirderLocationChooser.vue
@@ -3,7 +3,7 @@
     <template #activator="{ on }" v-if="activatorDisabled === false">
       <div class="d-flex">
         <slot name="activator" v-bind="{ on }">
-          <v-btn v-on="on">
+          <v-btn v-on="on" :disabled="disabled">
             Choose...
           </v-btn>
         </slot>
@@ -12,6 +12,7 @@
           class="pl-4"
           root-location-disabled
           :location="selected"
+          readonly
         />
       </div>
     </template>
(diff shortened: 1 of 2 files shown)


arjunrajlab (changes requested):
Progress bar looks great! Perhaps could make it a bit thicker. I also think somehow dimming out the interface underneath would be good, and perhaps hiding the buttons. That would be a bit less confusing for the user, who might want to click on the buttons while the quick import is proceeding.

luciemac:
LGTM

bruyeret:
Is it better now @arjunrajlab ?

arjunrajlab (approved):
Yes, looks great!

bruyeret:
Do you want to review the two last commits @luciemac ?

src/components/GirderLocationChooser.vue
```diff
@@ -12,6 +12,7 @@
           class="pl-4"
           root-location-disabled
           :location="selected"
+          :readonly="true"
```
luciemac, line 15: can't you go with only "readonly" property instead of ":readonly="true""?

luciemac:
then, LGTM

src/components/GirderLocationChooser.vue
bruyeret: Yes you're right!

PR #589, opened 2023-10-27
Move maxmerge folder into dataset folder

### LINKED ISSUE #557: All the "_maxmerge" files are sort of confusing
It is confusing in the interface to have a lot of "_maxmerge" files around. Would it be better to place them in the directory for the dataset?

bruyeret:
I moved the maxmerge folder into the dataset folder but there are maybe paths to change somewhere that I didn't notice
What do you think @manthey?

Close #557


arjunrajlab (approved):
I tried it and it seems to work very well, thanks!

luciemac:
LGTM

PR #591, opened 2023-10-27
Fix normal upload

### LINKED ISSUE #590: Upload data button on top right not working. I think it is still a nice feature.
I guess the issue is that now, comprehensive upload takes files from the Dropzone. But it still brings us to a naming page. I think we could just go to that page.

bruyeret:
Doing `this.files = ...;` didn't work

Close #590


arjunrajlab (approved):
Works great, thanks!

luciemac:
LGTM

PR #593, opened 2023-11-02
Use a docker with newer python

manthey:
Girder's default docker uses a base image that has python 3.7 and doesn't explicitly override it.  This uses a girder test docker that has multiple version of python.  Use 3.11


arjunrajlab:
Works great, fine to merge!

PR #598, opened 2023-11-10
Add the "Add dataset to collection" option in breadcrumbs

### LINKED ISSUE #562: "Add dataset" modal for directly adding to existing collection
I like this idea of having a modal for directly adding to existing collection a lot!

image

I feel like the wording for the top option should be "Add an existing dataset to the current collection" and the other option should be "Upload new dataset and add to current collection". Overall, the second option is the one that will probably be used more, but there's a lot more UI to it, so having it be the first option might be a bit weird looking.

Also, the "edit/review variables" checkbox should be more clearly associated with the "Add new dataset" option, because for an existing dataset, you cannot edit the variables. Right now, it's sort of by the main submit button, so it seems like it applies to either option.

image

This button would be better named "Add dataset":
image

bruyeret:
I replaced the select component with a location picker
I also did minor changes to make the design of this feature simpler
Tell me if there is anything to change @arjunrajlab 

Close #562

### DIFF
--- a/src/components/CustomFileManager.vue
+++ b/src/components/CustomFileManager.vue
@@ -28,7 +28,7 @@
       v-if="currentLocation"
       :location.sync="currentLocation"
       new-folder-enabled
-      :selectable="menuEnabled"
+      :selectable="menuEnabled || selectable"
       v-model="selected"
       v-bind="$attrs"
       v-on="$listeners"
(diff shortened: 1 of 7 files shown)


bruyeret:
All review should be fixed @arjunrajlab

luciemac:
LGTM

arjunrajlab:
After adding a new dataset, this dropdown doesn't move to the new dataset, but it probably should go to the new dataset by default:
image

arjunrajlab:
I tried adding an incompatible dataset and didn't get a warning. Instead, it just left me on this screen:
image

arjunrajlab:
If there are a lot of files, the UI (button for submitting) is a bit hard to find. I suggest making a scroll box for the file list (like on the main screen).
image

arjunrajlab (approved):
Approved!

PR #600, opened 2023-11-17
Use filename variables only when uploading several files

### LINKED ISSUE #599: Don't guess variable for filename if there is only one

bruyeret:
Close #599


luciemac:
LGTM

PR #603, opened 2023-11-21
Updated UI to Collection view 

### LINKED ISSUE #563: Updated UI to Collection view
I really like this updated UI for the Collection view:

image

The most important part is the "Add dataset to current collection" button:
image

This button here should say "Delete collection", but we have to figure out the logic of this option.

image

The thing is that every dataset has to belong to a collection. So if you delete the collection, what happens? It's not bad per se, but if you have a dataset that has no collection, then it should create a new collection when you try and reopen it. I'm not sure if that's currently the case.

bruyeret:

Fixes request changes of #598 
Close #563

### DIFF
--- a/src/components/AddDatasetToCollection.vue
+++ b/src/components/AddDatasetToCollection.vue
@@ -1,10 +1,10 @@
 <template>
-  <v-card>
+  <v-card class="smart-overflow">
     <v-card-title class="mb-4">
       <span class="text--secondary">Adding dataset to collection:</span>
       <span class="text--primary">{{ collection.name }}</span>
     </v-card-title>
-    <v-card-text>
+    <v-card-text class="smart-overflow">
       <v-radio-group v-model="addDatasetOptionType">
         <v-radio
           label="Add an existing dataset to the current collection"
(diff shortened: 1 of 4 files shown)


arjunrajlab:
Actually, never mind, I think there was an issue with that particular file, this works!

arjunrajlab (approved):
Looks great!

src/components/AddDatasetToCollection.vue
```diff
@@ -0,0 +1,371 @@
...
+
+  set addDatasetOptionType(type: TAddDatasetOption["type"]) {
+    switch (type) {
+      case "add":
+        this.option = defaultDatasetAddOption();
+        break;
+      case "upload":
+        this.option = defaultDatasetUploadOption();
+        break;
+    }
```
luciemac, line 180: no default case ?

bruyeret: The two possible values for `type` are checked
TypeScript agrees with it and does not ask for more
I can add a `default:` above `case "upload":` but it should never be used

luciemac: sounds good, just wanted to make sure there is no possible troubles here

luciemac:
LGTM

Issue #595, opened 2023-11-08
Zeiss file causes skyrocketing memory usage and crash, doesn't load


arjunrajlab:
Tried to upload this pretty small CZI (Zeiss) file. It got stuck after the upload and never stopped, and eventually used up all the RAM on my machine.

https://www.dropbox.com/scl/fi/uv4sc4zkj4xenr4k7dxsu/well3_raw_2.czi?rlkey=x8srk0vzdcmedov3rps8fva63&dl=0

It's a simple 4 color z stack.

arjunrajlab:
This appears to be a Mac-specific issue.

arjunrajlab:
Actually, the performance on this relatively small dataset seems very poor…

http://ec2-3-83-221-74.compute-1.amazonaws.com:8081/#/datasetView/654bf24e7854b6d57ff106e1/view?z=16

arjunrajlab:
It appears to be a front end issue @bruyeret

arjunrajlab:
@bruyeret an interesting data point: if you transcode to TIFF, you get the issue. If you do NOT transcode to TIFF, the issue is not there.

bruyeret:
The performance issue comes from Vue's reactivity as said before
It occurs when calling `map.draw();` in `ImageViewer.vue` at line 966
The reason is that the geojs fetch queue is reactive and the update function is adding around 1000 elements in it
The complexity of doing so when the _queue attribute is reactive is quadratic
Commenting out `mapentry.params.layer.queue = [...mapentry.imageLayers[0].queue];` l.798 fixes the issue
I still have to investigate how to properly solve the issue tomorrow but maybe you have a clue @manthey ?

arjunrajlab:
Interesting! Why does it only happen when converted to TIFF? Does that provide some clue?

bruyeret:
I don't know how this is linked to converted TIFFs yet
Maybe the queue is bigger when using converted TIFFs
It doesn't seem to come from `queue` being reactive, as it is reactive even when commenting out the line
It is the assignation itself which causes issues
`mapentry.params` is built using [`pixelCoordinateParams`](https://opengeoscience.github.io/geojs/apidocs/geo.util.html#.pixelCoordinateParams)
Is it really necessary to do this assignation @manthey? What is its purpose?

bruyeret:
I saw in an earlier commit this commit message, which could explain why :
> Used a shared tile queue across layers.
> 
[...]

PR #618, opened 2023-12-08
Removed default transcoding for lif czi files

arjunrajlab:


bruyeret:
Note: do not merge yet as issue #595 could be soon resolved and the file extensions added should not be transcoded

arjunrajlab:
Turns out that we should probably leave transcoding on to make these files more performant.

PR #633, opened 2024-02-09
Add step to worker interface sliders

bruyeret:


luciemac (approved):

bruyeret:
@zjniu This PR adds another property that you can set for the worker interface, that is "step"
If not defined, it will default to -1 (no step, maximum precision)
For example:, this will force the "Diameter" slider to be an integer between 0 and 200 (defaults to 10)
```json
'Diameter': {
    'type': 'number',
    'min': 0,
    'max': 200,
    'default': 10,
    'step': 1
},
```

arjunrajlab (approved):
Works great, thanks!

PR #637, opened 2024-03-09
Add an annotation list filter and column checkboxes

arjunrajlab:
Updated the list to have a filter by annotation ID. People in lab requested this feature to help them find specific annotations.

I also put in a dropdown selection menu for which columns to show.


src/components/AnnotationBrowser/AnnotationList.vue
```diff
@@ -216,7 +263,16 @@ export default class AnnotationList extends Vue {
...
+    const query = this.filterQuery || ""; // Ensure that filterQuery is always treated as a string
+    return this.filterStore.filteredAnnotations
+      .filter(annotation => {
+        // If there is no query, return all annotations
+        if (!query.trim()) return true;
+
+        // Otherwise, filter by the id
+        return annotation.id.toString().includes(query.trim());
+      })
+      .map(this.annotationToItem);
```
bruyeret, line 275: The issue with this approach is that you add a filter without adding it to the `filterStore`.
This will cause issues with other components that need to get the filtered annotations, for example when displaying only filtered annotations.
The right approach is to add a new filter "`idFilter`" in `filters.ts` and set it in `AnnotationFilters.vue`.
Do you need help for this?

src/components/AnnotationBrowser/AnnotationList.vue
```diff
@@ -180,9 +213,23 @@ export default class AnnotationList extends Vue {
   readonly filterStore = filterStore;
   readonly getStringFromPropertiesAndPath = getStringFromPropertiesAndPath;
 
+  columnOptions = [
+    { text: "Annotation ID", value: "annid" },
+    { text: "Index", value: "id" },
```
bruyeret, line 218: `annid` is not easy to differentiate from `id`
The issue here is that `id` have been badly named, it should be `index` or at least `idx`
Suggested modification:
- `id` -> `index`
- `annid` -> `id`

bruyeret (changes requested):
I can do the changes if you prefer @arjunrajlab

src/components/AnnotationBrowser/AnnotationList.vue
arjunrajlab: Ohhh, I see, yes, if you can do that, it would be great, thanks! I think it's a bit too complicated for me to figure out… I will probably make a lot of mistakes :)

src/components/AnnotationBrowser/AnnotationList.vue
arjunrajlab: Agreed, I was scared to make the edit :). I'll do that now.

arjunrajlab: @bruyeret I just pushed a new commit to the branch fixing this problem.

src/components/AnnotationBrowser/AnnotationList.vue
arjunrajlab: I guess I am wondering if it is better to keep it the way I implemented it. I think we would still want every other component that is accessing the item list to not worry about this particular filter, because this is meant to be just a quick "lookup" filter.

src/components/AnnotationBrowser/AnnotationList.vue
arjunrajlab, line 1: I'm not sure how to comment on a line, but this is for **line 355** :)

@bruyeret Also, I noticed here that "value" is not added in here. It's fine for now, but one problem is that if we want to eventually implement sorting in the table (which is a much requested feature), then if these values are all the same (since they are set to ""), then the arrows will all turn up and down together. You can verify that issue right now by clicking the arrows in the header items for property values—they all go up and down together. I could easily fix it, but not sure what you recommend putting in for unique values here. Also not critical for this particular PR; I figured we would do the sorting of the table in a separate PR.

src/components/AnnotationBrowser/AnnotationList.vue
bruyeret: Yes it is possible to have the "real" filters that work for all components and local filters only for the annotation list
But it should be clear for the users and in the code that it is a separate filter
If you think it is readable from a user point of view, we can leave the UI as it is
From a developer point of view, the filter is kind of "hardcoded" in the function, are you ok if I refactor it a bit?

src/components/AnnotationBrowser/AnnotationList.vue
bruyeret: I thought that this one was a quick fix but not at all
The line 355 sets the header for all properties, but not the values
As said in the comment at this line: `the slot "body" is used`
What I meant in this comment is that the value of `value` doesn't matter because it is not used
The body of the vuetify component that is used to display the annotations is overridden, and the sorting has no effet
We should do the sorting in a separate PR

bruyeret (approved):

src/components/AnnotationBrowser/AnnotationList.vue
arjunrajlab: Yes, that took me a while to figure out in the code :). Normally, the header would set the values, but the way it is set up is more custom, so the management of the header appeared to be separate from the management of the table itself. That's why I had to use all the v-if statements for the columns :). The issue with the "value" is that we need to put in the value in order for the sort arrows to work independently, but yes, fixing the sorting overall will be a bit of a headache because of the manual table management. Anyway, not critical at this point :).

PR #648, opened 2024-04-05
SAM tool important features

### LINKED ISSUE #646: Remaining SAM interface items
Task list for remaining items for SAM implementation in NimbusImage:

- [ ] **Enable "live preview".** It would be great to show a live preview of the mask that would be generated if the user clicks. It should just run SAM on the current point as the prompt. It doesn't need to process the mask at all—if it generates a complex mask, that is fine, the user will know that it may get simplified upon committing it to an annotation.
- [ ] **Progress bar for the encoder.** The encoder takes some time to compute, especially on older computers. It would be good to show some indicator on the screen or on the tool itself to show that it is doing the encoding. Perhaps also could be merged with this? https://github.com/Kitware/UPennContrast/issues/549
- [ ] **Interface for the tool.** I think there should be a toggle between "turbo mode" and "complex prompt mode". In turbo mode, you can shift click or shift drag, and it will automatically commit the annotation upon click/end of drag. That way, if someone just wants to quickly click on a bunch of objects, they can do so. In turbo mode, if you make an annotation, it gives you an "Oops, cancel!" link/button in the bottom left that allows you to undo the latest annotation quickly (lasts for 1 second). In complex prompt mode, it can allow for more complex prompts with multiple foreground and background prompt points/rectangles. In that mode, the existing buttons should be sufficient.
- [ ] **Simplify the polygon.** Give an optional "smoothing parameter" (default 1.5) to send to the algorithm to smooth out the resulting polygon so that it doesn't have quite so many points.

bruyeret:
Add SAM annotation tool features requested in #646 

Note: you need to rebuild the itk pipeline for SAM because of a small change (`npm run emscripten-build`)

Close #646

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -27,7 +27,7 @@ import {
   IGeoJSFeatureLayer,
   IGeoJSLineFeatureStyle,
   IGeoJSMap,
-  IGeoJSPoint,
+  IGeoJSPosition,
   IGeoJSPointFeatureStyle,
   IGeoJSPolygonFeatureStyle,
   IImage,
@@ -88,7 +88,7 @@ export default class AnnotationViewer extends Vue {
 
   // Compute the centroid of each annotation, taking unrolling into account
   get unrolledCentroidCoordinates() {
-    const centroidMap: { [annotationId: string]: IGeoJSPoint } = {};
+    const centroidMap: { [annotationId: string]: IGeoJSPosition } = {};
     const annotationCentroids = this.annotationStore.annotationCentroids;
 
     const anyImage = this.store.dataset?.anyImage();
@@ -530,7 +530,7 @@ export default class AnnotationViewer extends Vue {
 
   // Transform a list of coordinates, taking unrolling into account.
   unrolledCoordinates(
-    coordinates: IGeoJSPoint[],
+    coordinates: IGeoJSPosition[],
     location: IAnnotationLocation,
     image: IImage,
   ) {
@@ -549,7 +549,7 @@ export default class AnnotationViewer extends Vue {
       const tileX = Math.floor(locationIdx % this.unrollW);
       const tileY = Math.floor(locationIdx / this.unrollW);
 
-      return coordinates.map((point: IGeoJSPoint) => ({
+      return coordinates.map((point: IGeoJSPosition) => ({
         x: tileW * tileX + point.x,
         y: tileH * tileY + point.y,
         z: point.z,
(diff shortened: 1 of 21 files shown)


arjunrajlab (approved):
Tested it out, and it works great!

src/components/AnnotationViewer.vue
```diff
@@ -272,29 +298,69 @@ export default class AnnotationViewer extends Vue {
...
+      this.annotationLayer.removeAnnotation(this.samLivePreviewAnnotation);
+      this.samLivePreviewAnnotation = null;
+    }
+
+    // Create the new annotation
+    const vertices = this.samLivePreviewOutput;
+    if (!vertices) {
+      return;
+    }
+    const style = {
```
luciemac, line 329: Don't you have the possibility to customise your annotation style?

src/pipelines/computePipeline.ts
```diff
@@ -81,15 +80,46 @@ export class ComputeNode<
...
+  /**
+   * Create and return a callback function that will trigger recomputation of this node
+   * @param parentNode The parent node that will trigger a recomputation
+   * @returns The callback that needs to be called when the parentNode's output changes
+   */
+  private createComputeCallback(parentNode: ComputeNode<any, any>) {
+    let previousParentOutput = parentNode.output;
+    const computeCallback = () => {
+      const currentParentOutput = parentNode.output;
+      // Don't recompute when the parent output changes from NoOutput to NoOutput
```
luciemac, line 102: if it changes from NoOutput to NoOutput, then, it did not change :)

src/components/AnnotationViewer.vue
bruyeret: Not yet but it could be possible in the future :D

src/pipelines/computePipeline.ts
bruyeret: Good point!

src/pipelines/samPipeline.ts
```diff
@@ -341,127 +356,255 @@ async function runItkPipeline({ masks }: IDecoderOutput): Promise<IItkOutput> {
...
+  const { encoderNodes, decoderNodes, previewNodes } = nodes.allNodes;
+  const computingMessageMap: [ComputeNode<any, any>, string][] = [
+    [encoderNodes.sessionNode, "Creating encoder"],
+    [encoderNodes.inferenceNode, "Encoding"],
+    [decoderNodes.sessionNode, "Creating decoder"],
+    [previewNodes.sessionNode, "Creating decoder preview"],
+  ];
+  const recomputeLoadingMessage = () => {
+    const messages: string[] = [];
+    for (const [node, message] of computingMessageMap) {
```
luciemac, line 575: Can't you use a `filter` here ?

luciemac:
Then, LGTM

src/components/AnnotationViewer.vue
luciemac: Don't know if it is useful for now, but maybe you can generalize this style, to make it easily customizable, what do you think?

Issue #643, opened 2024-03-15
Recenter view upon click in the annotation list


arjunrajlab:
Currently, when you click on an item in the annotation list, it will bring you to the frame (XY, Z, T) of the annotation, but if you are zoomed in to a different part of the image, it doesn't recenter the view around the clicked annotation, which can be tricky. It would be good to recenter around the clicked annotation.

There is some choice about specifying the level of zoom. I'm not 100% sure what the right answer is, but for polygons, I think something that is 10x the size of longest side length of the bounding box of the polygon would be good. So if the polygon is 50 pixels on the longest edge of the bounding box, then show a 500 pixel window (long side of the main view) centered around the polygon. As for points… hmm. I think maybe just a 300 pixel window (long side of the main view) by default? I think that will probably capture most scenarios. We can adjust these defaults as needed (I don't think they need to be interface options necessarily).

bruyeret:
Waiting for user feedback after merging #649, before deciding to close the issue or add the zoom to the recenter

PR #649, opened 2024-04-09
Add a mechanism to center the view on the annotation when clicking in the list

bruyeret:
It is part of #643 but there is no change in the zoom when clicking the annotation in the list
So this PR will not close the issue but it is a step in the right direction


luciemac (approved):

arjunrajlab (approved):
Looks great!

PR #651, opened 2024-04-09
Fix API call that required admin privileges

### LINKED ISSUE #605: Allow NimbusImage usage without requiring admin permissions
Currently, in order to use NimbusImage, the user has to have admin privileges in Girder. Can we fix that?

bruyeret:
The fix given by David seems to not cause any issue.
I tested uploading a new dataset, creating tools, properties and running a property worker.
I found no issue for now, maybe I fixed the worker issue without noticing in a different PR.

Close #605

### DIFF
--- a/src/store/GirderAPI.ts
+++ b/src/store/GirderAPI.ts
@@ -88,8 +88,7 @@ export default class GirderAPI {
     id: string,
     type: IGirderSelectAble["_modelType"],
   ): Promise<IGirderSelectAble> {
-    const config = { params: { type } };
-    const response = await this.client.get(`resource/${id}`, config);
+    const response = await this.client.get(`${type}/${id}`);
     const resource = response.data;
     if (resource) {
       resource._modelType = type;


luciemac (approved):

arjunrajlab (changes requested):
So I was able to get an annotation worker to run (cellpose, on the normmedia_8well_col2_livecellgfp dataset, but probably anything would be fine). However, the property workers would not work. I did both blob metrics and blob intensity. In both cases, the progress bar goes, but it still shows a red number badge, and no properties show up in the Object List panel. If I do it with admin privileges on the same branch, it works as expected, so it is definitely a problem requiring admin privileges. Is there anything else I can do to help debug?

luciemac (approved):

luciemac (approved):

bruyeret:
After more digging, I found out that some API calls did not set the proper user permissions
It affects history (you can undo or redo but this will not effectively change anything) and annotation property values (you can push some values but you can't get them)
I added these permissions but you need to rebuild girder (`docker-compose build girder`)
Also, the history entries and the annotation property values that have been created with the wrong permissions will still be unaccessible from non admin users
Does it work now @arjunrajlab? (if it works, you can approve the PR)

arjunrajlab (approved):
Now it works!

PR #652, opened 2024-04-09
Fix UI when uploading many files

### LINKED ISSUE #613: In multi-file import, it will let you click "Submit" before it's read and assigned the variable names.
For example, try uploading these files:

https://www.dropbox.com/scl/fo/8nffbjuulk8o7qx5qdpmi/h?rlkey=n05zxfqzznjwdwdcg7j03m8w8&dl=0

The goal is the following:

image

The auto-assign is a little off, which is fine:

image

The problem is that the green submit button becomes available *before* the auto-assignment. If you assign the variables manually, it doesn't work. If you wait a bit, then it will assign the variables automatically, and then you can reassign. I think it is either running the algorithm for filename parsing, or it is reading one of the files to determine what file variables there are. Either way, I think the submit button should be greyed out until the auto-assignment/filename parsing completes. It would also maybe be good to give an indication saying "File name variable parsing underway…" or something like that.

bruyeret:
Close #613

- Better guess for dataset name
- Max height for the file list
- Show loading and prevent submit until all variables are computed

### DIFF
--- a/src/views/dataset/MultiSourceConfiguration.vue
+++ b/src/views/dataset/MultiSourceConfiguration.vue
@@ -5,7 +5,13 @@
       <v-divider class="my-2" />
       <v-data-table :headers="headers" :items="items" item-key="key" />
     </v-card>
-    <v-card class="pa-4 my-4">
+    <v-card v-if="initializing" class="my-4">
+      <v-card-title>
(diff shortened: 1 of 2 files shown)


arjunrajlab (approved):
Looks great, thank you!

luciemac (approved):

PR #655, opened 2024-04-10
Add an endpoint for adding multiple annotation property values

### LINKED ISSUE #550: Add endpoint to upload multiple properties at once
Right now, we have to upload each property value one at a time, which is very inefficient. It would be better if we could upload a list of properties.

Current call is:
`workerClient.add_annotation_property_values(annotation, prop)`

I guess it would be nice to have:
`workerClient.add_multiple_annotation_property_values(annotation_ids, prop_list)`
where `annotation_ids` is a list of the annotations_ids. Or maybe better for it to be a dictionary, with annotation_id as a key and property as a value? Whatever is most efficient.

bruyeret:
Close #550

This PR is based on PR #651 to avoid a conflict
It should be merged after

You need to rebuild girder to see the effects of this PR

The workers can use the function `addMultipleAnnotationPropertyValues()`
For convenience, here is the signature and documentation of this function:
```python
    def addMultipleAnnotationPropertyValues(self, entries):
        """
          Save one or multiple computed property values for the specified annotations
          The recursive_dict_of_numbers can be (and usually is) just a number
          :param list entries: A list of property values for an annotation. Each entry is of type { "datasetId": string, "annotationId": string, "values": { [propertyId: string]: recursive_dict_of_numbers } }
        """
```

### DIFF
--- a/devops/girder/annotation_client/annotation_client/annotations.py
+++ b/devops/girder/annotation_client/annotation_client/annotations.py
@@ -12,6 +12,7 @@ PATHS = {
     'connect_to_nearest': '/annotation_connection/connectTo/',
 
     'add_property_values': '/annotation_property_values?datasetId={datasetId}&annotationId={annotationId}',
+    'add_multiple_property_values': '/annotation_property_values/multiple',
     'get_dataset_properties': '/annotation_property_values?datasetId={datasetId}',
     'get_annotation_properties': '/annotation_property_values?datasetId={datasetId}&annotationId={annotationId}',
     'histogram': '/annotation_property_values/histogram?propertyPath={propertyPath}&datasetId={datasetId}&buckets={buckets}'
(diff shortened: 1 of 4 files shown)


luciemac (approved):

arjunrajlab (commented):
@bruyeret I put in a new function for adding multiple properties as a helper function (similar to what already exists). Slight difference from the "single" function: it requires the datasetId to be passed along inside the dictionary, which allows to theoretically push properties across multiple datasets, in case someone wants to do that in future.

bruyeret:
You just missed a `.annotationClient`  @arjunrajlab, so I changed your commit, is it good for you?
I could do something so that the user doesn't have to use the id of the property (that should always be `self.propertyId`)
For example a kind of tree structure like this:
```
{
  "myDatasetId": {
    "myAnnotationId": 42,
    "anotherAnnotationId": 84,
  },
  "anotherDatasetId": {
    "lastAnnotationId": 2,
  }
}
```

arjunrajlab:
Ah okay, that's cool, thanks for catching that! Even in the current structure, I don't think it needs the propertyId, does it? Just the datasetId and the annotationId? The property values are just set with "values", right?

bruyeret:
It's not the property name but the ID, for example `6617bb27c2e0ea61cecb7636`

arjunrajlab:
Oh, that is weird. In the previous call, we didn't need that, right? Here, it's unclear how we would even get the propertyId if the property has not yet been created, no?

bruyeret:
When the user creates a property in the "Measure Objects" panel, it creates a property and this is its ID that is given to the worker
You can see the function to add a single property value right above the function that you wrote:
```python
    def add_annotation_property_values(self, annotation, values):
        """
        Add property values to this annotation
        :param dict annotation: The annotation having the properties
        :param (number|dict) values: The value for this property (a number or a dict of numbers)
        """
        property_values = {self.propertyId: values}
        self.annotationClient.addAnnotationPropertyValues(self.datasetId, annotation['_id'], property_values)
```
It uses `self.propertyId`

arjunrajlab:
Ahhh okay, I see now, sorry. I think your proposed solution makes the most sense then. So sorry it took me a while to understand!

arjunrajlab:
In that case, it would be like this?
```
{
  "myDatasetId": {
    "myAnnotationId": 42,
    "anotherAnnotationId": 84,
  },
  "anotherDatasetId": {
    "lastAnnotationId": 2,
  }
}
```

And you could put in nested dictionaries as well instead of the numbers?

bruyeret:
Exactly! Should I do this ?

arjunrajlab:
Yes, that would be great, thank you!On Apr 11, 2024, at 9:50 AM, Thibault Bruyère ***@***.***> wrote:﻿
Exactly! Should I do this ?

arjunrajlab:
Bump @bruyeret

luciemac (approved):

bruyeret:
@arjunrajlab
You can merge this PR whenever you want
I tested on small samples of data but not at full scale
So if you want to test this branch before merging, you can 👍

PR #658, opened 2024-04-11
Make dataset view remember last location

### LINKED ISSUE #654: Breadcrumb for location
Is it possible to mark/save the current location (XY, Z, Time) on a dataset so that if you reopen the dataset, it brings you to the same location?

bruyeret:
If you stay more than 5 seconds at the same location, the last location is saved in the dataset view
When reopening the the same dataset view, the location will be the same as before
If the url query contains a location (for example `z=9`), it will override the last location saved in the dataset view

Close #654


luciemac (approved):

arjunrajlab (approved):
Works great, thank you!

PR #659, opened 2024-04-11
Fix black screen when changing dataset

### LINKED ISSUE #657: Screen goes black when changing datasets
When you switch between different datasets within a collection, the main display screen goes black. A page refresh seems to solve the issue.

I'm not quite sure when this bug got introduced, but I can say that I started noticing this happening intermittently (even without switching between datasets) after the live-SAM update. 

Here's a link a video displaying the problem: https://www.dropbox.com/scl/fi/dhs7p0wxm5ysn1vnf5h82/Screen-Recording-2024-04-10-at-1.06.55-PM.mov?rlkey=obph7erpw1ebwy0xyd43ur2vu&dl=0

bruyeret:
Close #657 

It was caused by an error thrown by oldWidgetLayer.deleteWidget when the layer was deleted

### DIFF
--- a/src/components/ImageViewer.vue
+++ b/src/components/ImageViewer.vue
@@ -875,12 +875,14 @@ export default class ImageViewer extends Vue {
     const oldWidgetLayer = this.scaleWidget?.layer();
     if (
       this.scaleWidget &&
-      oldWidgetLayer &&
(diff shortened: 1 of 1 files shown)


arjunrajlab (approved):
It works now, thanks!

PR #662, opened 2024-04-19
Add an action button to delete connections

### LINKED ISSUE #660: Provide an action button to remove all connections
Some users would like to remove all existing connections. Can we put in a button in the "Actions" subpanel in the Object List panel called "Delete all connections" that deletes all connections?

Bring up a dialog to remove all from dataset, remove all from location, remove all connected to selected annotations.

bruyeret:
Close #660


arjunrajlab (approved):
Works great, thank you!

luciemac (approved):

PR #663, opened 2024-04-19
Add tool to connect or disconnect annotations using click

### LINKED ISSUE #661: Connection tool using right click
Right now, there is a useful add connection tool that makes a lasso and connects items that way. What would also be helpful is something that allows you to connect with a couple clicks. In the ideal scenario, you click on one annotation, then right click on the annotation you want to connect to, and it bring up a little right click menu with the option to add connection. The tool itself would have the option of whether the first or second object is the "parent".

Or, even easier: let's make a new tool. You click once and then click again, and it connects the two. If you click once and hit escape, then it resets.

bruyeret:
Close #661

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -40,6 +40,7 @@ import {
   SamAnnotationToolStateSymbol,
   TSamPrompt,
   TToolState,
+  ConnectionToolStateSymbol,
 } from "../store/model";
 
 import { logError, logWarning } from "@/utils/log";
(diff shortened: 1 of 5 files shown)


src/components/AnnotationViewer.vue
```diff
@@ -1264,20 +1279,74 @@ export default class AnnotationViewer extends Vue {
...
+      } else {
+        // Delete any connection between a parent and a child
+        await this.annotationStore.deleteAllConnections({
+          parentIds,
+          childIds,
+        });
+      }
+    }
+
+    // Reset or set the slected annotation id in the tool state
```
luciemac, line 1339: typo: slected -> selected

luciemac (approved):
Then LGTM

PR #664, opened 2024-04-19
Snapshot

### LINKED ISSUE #615: Snapshot doesn't work for Z max merge.
Can be reproduced with this file: https://www.dropbox.com/scl/fi/hyg3bou153fnq6lye3zlb/DDX58_AXL_EGFR_well2.nd2?rlkey=lf00zmmkqv4hc7c6fy9qgqvfp&dl=0.

Expected behavior is that a snapshot composite image would show the max merge as in the image viewer.
image (7)
However, what we get instead is the following. It shows only the Z level you're currently on and in the wrong color.
2023-12-01 04 11 - DDX58_AXL_EGFR_well2-2023-12-01 04 11 - DDX58_AXL_EGFR_well2 Configuration-test-2023-12-08 05 43

### LINKED ISSUE #623: Snapshots are not aware of multiple files within a collection
If you take a snapshot in dataset1 from a collection, then move to a different dataset (dataset2) and select the snapshot from dataset1, the snapshot navigates to the location of the snapshot in dataset2. It should navigate back to dataset1. I am guessing we just forgot to consider this possibility when we moved from single datasets to collections of datasets. We should probably also include an extra field in the snapshot browser list that shows the snapshot dataset so that people know which dataset the snapshot is from.

### LINKED ISSUE #640: Make an option to download all snapshots, not just the current one

### LINKED ISSUE #650: Snapshot issues
#640 
#623 
#615

bruyeret:
This pull request fixes and improve snapshots

Close #650 
Close #640 
Close #623 
Close #615 

The way that "screenshots" are made have been reworked
The ImageOverview component (the minimap) has changed thanks to some code factorization, and this fixes some issues with Z max merge, but the scale of the minimap is now wrong
I have tried a lot of different things but nothing worked
I now use the endpoint `${apiRoot}/item/${itemId}/tiles/region` instead of `${apiRoot}/item/${itemId}/tiles/zxy/{z}/{x}/{y}` and this is likely the reason why scaling is messed up

We could disable the minimap for now, merge this and wait for @manthey to come back to give us some advice


arjunrajlab:
I have also noticed, even before this PR, that the little square in the minimap doesn't move around properly anymore. Like, if you click and drag the square in the minimap, the image moves but the square doesn't move.

bruyeret:
Oh yes I noticed that too, it should be fixed by this PR

arjunrajlab (changes requested):
It would be good to also, for each snapshot, list which dataset it is in from the collection in the Snapshot list:
image

arjunrajlab:
Also, I tried the Z-merge snapshot feature, but that didn't seem to work properly.

arjunrajlab:
Download all does work nicely, though!

bruyeret:
@arjunrajlab is it better now? Dataset name is shown and the snapshots already saved the layers but didn't apply them when being loaded

luciemac (approved):

arjunrajlab (approved):
Nice, everything is looking good, thanks!

arjunrajlab:
Actually, could you enable the minimap, even if it is not quite right? I think people might like to have it as a way to get a general sense of where they are in the image.

bruyeret:
I put it back!
If you are ok, I can merge right now 😄

arjunrajlab:
Yes, let's merge, thank you!

PR #689, opened 2024-06-05
Pixel scalebar

### LINKED ISSUE #669: Scalebar is black
I think we changed it to white, but it seems to be black again:

image

### LINKED ISSUE #671: Add a pixel scale bar
It would be nice to have another scale bar that shows the number of pixels. I created a branch with the desired look:

https://github.com/Kitware/UPennContrast/tree/pixel-scale-bar

Here's what it should look like:

image

I made a very small edit to the code where I just add the widget and don't even track it, but probably needs to keep track of the widget and allow for deletion and so on. I think the interface for controlling it should be the following:

1. Put an option here called "Show pixel scalebar" (enabled by default):
image
2. If you click on either scalebar, brings up this dialog:
image
Add a toggle option to the dialog for "Show pixel scalebar".

bruyeret:
Close #669 
Close #671

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -404,7 +404,7 @@ export default class AnnotationViewer extends Vue {
     } else {
       // Preview shift drag select
       const vertices = mouseState?.path ?? [];
-      if (vertices.length > 0) {
+      if (vertices.length > 1) {
         this.selectionAnnotation = geojs.annotation.lineAnnotation({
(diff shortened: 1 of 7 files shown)


bruyeret:
@arjunrajlab I also fixed a big performance issue, by changing a 0 into a 1 🙃

arjunrajlab (approved):
Works great, thank you!

luciemac (approved):

PR #691, opened 2024-06-06
Snapshot fixes

### LINKED ISSUE #676: Opening snapshot panel while manual blob tool is selected causes a huge blob to be generated
If you have a manual blob tool selected and then open the snapshot panel, it tries to make a large annotation with the snapshot selection tool.

image

### LINKED ISSUE #684: Snapshot "download all" doesn't take into account which channels are on or off
Let's say you make two snapshots. One of them has layer 1 turned on and layer 2 turned off, the other has layer 2 turned on and layer 1 turned off. If you "Download all" of "composite", it downloads the composites of all the snapshots, but only with the layer settings of the current snapshot. So, in the example above, both downloaded composites would have layer 1 on and layer 2 off. It seems that the download all doesn't take into account the layer differences between the different snapshots.

bruyeret:
Close #676 
Close #684

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -655,7 +655,7 @@ export default class AnnotationViewer extends Vue {
 
   drawTooltips = throttle(this.drawTooltipsNoThrottle, THROTTLE).bind(this);
   drawTooltipsNoThrottle() {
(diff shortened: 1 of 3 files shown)


arjunrajlab (approved):
Works great, thank you!

luciemac (approved):

PR #692, opened 2024-06-06
Fix max merge when in absolute mode

### LINKED ISSUE #675: Z-merge doesn't work when layer display mode is in "absolute"
If you do z-merge while the layer display mode is set to "absolute" (instead of "percentile"), nothing happens and it just shows the current z.

image
image
image
image

bruyeret:
Close #675


luciemac (approved):

arjunrajlab (approved):
Works great, thanks!

PR #693, opened 2024-06-06
Keep a global folder location

### LINKED ISSUE #626: Navigating back to the home screen should preserve folder location
When you navigate to a folder and then go to a dataset, then when you return to the home screen (either by clicking back or the Nimbus "cloud" icon), it goes back to the home folder. Could we instead keep track of the current folder and preserve that as the location?

bruyeret:
Close #626


luciemac (approved):

arjunrajlab (approved):
It works!

PR #694, opened 2024-06-06
Add display order to worker interface

### LINKED ISSUE #678: Allow for ordering of the interface elements in annotation and property workers
Right now, when you specify the interface in a worker, the interface is rendered with the interface elements in alphabetical order. Sometimes, however, it would be good to organize the interface options in non-alphabetical order. Is there a good way to do that? One option I was thinking was to add an "_order" key that specifies the ordering. Like `'_order': 1`, etc. If some item is missing that key, then it shows at the end of the list in alphabetical order as before. Would that work?

```
def interface(image, apiUrl, token):
    client = workers.UPennContrastWorkerPreviewClient(apiUrl=apiUrl, token=token)

    # Available types: number, text, tags, layer
    interface = {
        'Channel': {
            'type': 'channel'
        },
        'Radius': {
            'type': 'number',
            'min': 0.5,
            'max': 10,
            'default': 1,
        },
    }
    # Send the interface object to the server
    client.setWorkerImageInterface(image, interface)
```

### LINKED ISSUE #681: Add a "no_cache" key to interface elements
In some workers, we would like to ensure a particular interface element (specified by the dictionary in the "interface" method) is "fresh" and not taken from the cache. It would be nice to have a "no_cache" key that could be set to True that tells the interface to show a spinning circle until the interface is refreshed, at which point it is updated to reflect the new interface dictionary.

bruyeret:
The property is `displayOrder` and is optional

Close #678 
Close #681

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationProperties/PropertyCreation.vue
+++ b/src/components/AnnotationBrowser/AnnotationProperties/PropertyCreation.vue
@@ -212,10 +212,7 @@ export default class PropertyCreation extends Vue {
   @Watch("dockerImage")
   dockerImageChanged() {
     this.isNameGenerated = true;
-    if (
-      this.dockerImage &&
-      this.propertyStore.workerInterfaces[this.dockerImage] === undefined
-    ) {
+    if (this.dockerImage) {
       this.propertyStore.fetchWorkerInterface({ image: this.dockerImage });
     }
   }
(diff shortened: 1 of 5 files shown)


luciemac (approved):

bruyeret:
@arjunrajlab the property to use is `noCache` in the property worker interface to bypass the caching of the interface by girder

luciemac (approved):

arjunrajlab (approved):
Works great, thank you!

Issue #503, opened 2023-08-25
Single call for getting pixel value


arjunrajlab:
We saturate the browser queue via pixel requests. These are fired one per layer debounced based on time. We could make a single-call endpoint and debounce them based on promise status, not time.

bruyeret:
Debouncing based on time avoids sending requests until you stop moving your mouse
Requests are only sent when the mouse stops for long enough
But I agree that a single request for all pixels would be a lot better
@manthey should we add this endpoint to the large image plugin or should I create a custom endpoint?

arjunrajlab:
True, yes, the debouncing is nice… but it does feel a little bit "laggy" with the current debouncing. I'm wondering if we could speed that up a bit if we just had a single call.

bruyeret:
What do you mean by laggy? It is almost instantaneous on my computer, when I stop moving the mouse
Maybe you have a dataset with many many channels?
I don't think that a single call will speed things up, but it is a good thing to avoid making many requests at once

arjunrajlab:
So if you use e.g. ImageJ, it updates essentially in real time. Of course, that would be a bit much considering we are going back and forth over the internet :), but perhaps we can lower the debounce time? Actually, even better would be to live update even while the mouse is moving. But I'm not sure if that's just too much server traffic. Let me know what you think, perhaps we can talk Friday. This is also really pretty low priority.

bruyeret:
Oh yes, for a desktop application that already has all the data locally, it is very easy to get the data
ImageJ just has to do `imageBuffer[mouse.x + width * mouse.y]` to get the pixel value
The equivalent on NimbusImage is to do a `getPixel` call to the server
We could eventually download all the raw images for the current location (when not unrolling), to do as ImageJ does, but it is very heavy and we lose all the optimizations of tiling the image

arjunrajlab:
In discussion with @bruyeret, I think perhaps it's good to just reduce the debounce time slightly. If moving to a single call would help avoid spamming the server with a lot of calls, then that's great. But if just reducing the debounce time to 1/2 or 1/3 is still acceptable, then that would probably achieve much of the effect on its own.

PR #702, opened 2024-06-07
Hotfixes

### LINKED ISSUE #698: Issue with file view after upload
HIGH PRIORITY: issue with file view after upload.

Uploaded:
image

Then I click "Submit" and get this:
image
(At this point, the little "disk" icon in the top right is pulsing. If I hover on it, the tooltip says "Saving…")

If I then click View Dataset, it does nothing and I get a server connection error:
image

### LINKED ISSUE #699: Data upload for multi-source TIFF doesn't work anymore
We have found that uploading files like the following doesn't work anymore:

https://www.dropbox.com/scl/fo/j9d89r7r3ruqfdlepzeq6/AG3glq2veQT_BWZUbtBwGfQ?rlkey=ufx6eitbz4yim6rs0dyoihkty&st=9uspbwtn&dl=0

It appears to get stuck on the transcoding step, but not sure.

bruyeret:
- Land in Public folder instead of user's folder
- Decrease debounce for pixel values to 25ms

Close #698 
Close #699 
Part of #503

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -1796,7 +1796,7 @@ export default class AnnotationViewer extends Vue {
 
   handleValueOnMouseMoveDebounce = debounce(
     this.handleValueOnMouseMoveNoDebounce,
-    100,
+    25,
   );
   async handleValueOnMouseMoveNoDebounce(e: any) {
     if (!this.dataset) {
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -411,6 +411,16 @@ export class Main extends VuexModule {
     girderRest: RestClientInstance;
   }) {
     await this.loggedInImpl({ girderRest, girderUrl });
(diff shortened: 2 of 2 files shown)


arjunrajlab:
Public folder fix is great! I am not noticing much speedup with the shorter debounce time on AWS or locally on my Mac. I didn't need to rebuild girder or anything, right?

arjunrajlab (commented):
Public folder fix is great! I am not noticing much speedup with the shorter debounce time on AWS or locally on my Mac. I didn't need to rebuild girder or anything, right?

arjunrajlab:
Comparing nimbusimage.com (no update) to nimbusimage.org (after update) seems to be the same…

arjunrajlab:
Actually, it is definitely faster, so it's good!

arjunrajlab (approved):
Works great!

PR #708, opened 2024-06-13
Fix indexing in tool creation and annotation creation

### LINKED ISSUE #704: Small index error in advanced options
If you set the assign to Z here:
image

You end up assigning to z=2:
image

I think the problem is that it is assuming the input is zero indexed in the "Assign to" field, but then the interface is mostly 1 indexed.

bruyeret:
There was a confusion with 0 indexing and 1 indexing:
- 1 indexing is used for the user interface and in tool configuration
- 0 indexing is used everywhere else inside the app

Close #704


arjunrajlab (approved):
Works great, thank you!

PR #720, opened 2024-06-20
Changed to chip selection

arjunrajlab:
@bruyeret I changed the selection thing for the columns to something that I think is more intuitive. It basically works, but it doesn't look quite as good when the interface is in light mode. The problem is that when a chip is selected, it looks like it is active, even if it is not active. In dark mode, it looks fine because the white outline dominates. But the underlying issue is still there. I couldn't figure out how to change the styling. Anyway, overall, it is still an improvement, so we can also just merge the PR, but if you have any quick fixes, go for it!

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationList.vue
+++ b/src/components/AnnotationBrowser/AnnotationList.vue
@@ -517,7 +517,9 @@ tbody tr.is-hovered:hover {
 }
 
 .v-chip {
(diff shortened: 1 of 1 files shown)


bruyeret (approved):
I tried playing with the focus style of the elements but I can't find a way to remove this "focus" effect on the chips

PR #723, opened 2024-06-20
Remove repeat words

arjunrajlab:
Removes repeated words in property autonaming to avoid "blob Blob blob" things from happening.


bruyeret (approved):
I made very small changes to avoid joining the words and splitting them right after

arjunrajlab:
Ah yes, that's how I did it at first, until I realized it doesn't work :)…

image

The problem is that the tokens are actually not words. If you join "YFP blob" and "Blob metrics", it doesn't know they are repeated. So actually the simplest thing was to just let it join it and then re-split. Is there a way to undo the commit? Although yes, we should remove the `//annotationNames` thing.

bruyeret:
Oh ok yes, I'll revert the changes, sorry for the mistake!

PR #724, opened 2024-06-21
Put tool list in bigger panel

arjunrajlab:
@bruyeret I made the tool list bigger and made it come from a button. I think the interface is a bit more discoverable now, because the list of items was so long. The only problem is that the floating panel is now to the left of the button, would be better to the right of the button, I think. Also, we have a lot of old tools or non-functional tools. Perhaps we can comment them out for now?

### DIFF
--- a/public/config/templates.json
+++ b/public/config/templates.json
@@ -1,6 +1,6 @@
 [
   {
-    "name": "Manual annotation tool",
+    "name": "Manual object tool",
     "type": "create",
     "shortName": "Manual",
     "interface": [
@@ -125,7 +125,7 @@
     ]
   },
   {
-    "name": "Select annotation tools",
+    "name": "Selection tools",
     "type": "select",
     "interface": [
       {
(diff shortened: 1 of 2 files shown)


arjunrajlab:
@bruyeret, I mostly just made this to give some sense of a potential interface improvement. Feel free to trash and start from fresh :).

bruyeret:
I brought back the divider, as the item were still generated but unused
Do you want to remove them? Or restyle them?

bruyeret:
Also, we can maintain a list of hidden tools, to use as filter if you want to hide them, what do you think?

arjunrajlab:
That looks great! And yes, a list of hidden tools would be nice so that we can keep drafts for testing, etc. Great idea!

bruyeret:
I added tool hiding, you can use the text from the UI to either remove tools one by one or a whole section
For example, you can remove `"Point"`, `"Line"` or `"Blob"` tools, or even the whole section if you remove `"Manual annotation tool"`
Simply add the tools you want to hide to `hiddenToolTexts` in `ToolTypeSelection.ts`

PR #725, opened 2024-06-21
Change default to composite

arjunrajlab:
Mostly, people want to get the composite, so I made that the first option. I also put (zip) after "All layers" so that people know they will be getting a zip of the individual layers.


bruyeret (approved):
LGTM!

PR #730, opened 2024-06-24
Constrast panel

arjunrajlab:
Adds a pop-up contrast panel, allowing for quick contrast adjustment for all channels.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -259,26 +259,6 @@ export default class App extends Vue {
   text-overflow: unset;
   overflow: unset;
 }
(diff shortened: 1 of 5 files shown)


arjunrajlab:
@bruyeret Perhaps we can remove the title bar and close icon to save a little space? We could also make the background transparent to give the individual layer contrast panels a "floating" look.

luciemac (approved):

bruyeret:
Is it better @arjunrajlab ?

arjunrajlab:
Yes!

PR #732, opened 2024-06-25
Add a little filter for the property list

arjunrajlab:


arjunrajlab:
@bruyeret I made some changes to the property list to make it a bit more usable when we have large numbers of properties. For instance, this is quite common:
image
Let me know what you think!

bruyeret:
We could have a single tab and two togglable icons instead of a checkbox: an eye and a funnel, with tooltips on hover to explain what each icon is for ("Show in list" and "Use as filter")
What do you think @arjunrajlab ?
image

arjunrajlab:
Interesting! Yes, I was debating whether to do a tabbed interface or the multiple boxes thing (although definitely the two icons is better than two checkboxes). In common usage, I have found that very few people are using the "Filter" option, and almost exclusively they are using the "display in list" option. So I figured it would be best to just make that the default, but leave a "Filter" in the tab to leave it at least somewhat discoverable. Does that make sense?

I think we don't need to show the tabs by hovering, though, probably might make the interface a little jumpy.

bruyeret:
The only issue I have with the two tabs is that I find it less intuitive than using a single tab
Users could for example be in the "filter" tab without noticing, check the properties that they want to show and think that there is a bug because the property doesn't show up in the annotation list

To avoid the interface to be jumpy, I was planning on using menus instead of opening new lists

arjunrajlab:
Hmm, yes, I could see that… although again, I think very few people use the filters, so as long as the default is set to the filter tab, I think that will cover almost all use cases.

I was thinking about nested menus, but I think the column view is actually a bit better. The reason is that people often want to click multiple checkboxes at a time from the list. With nested menus, they'd have to click through each time, which is a bit less easy to navigate.

I can ask in lab today what people would prefer!

bruyeret:
I refactored the code to use computed properties, factorize it and remove unnecessary code (mostly styling)
I also changed a little bit the behavior of search, and use icons instead of checkboxes to differentiate the tab selection
Is it better? If so I will ask Lucie to review my code because I changed a lot of things

arjunrajlab:
Looks very nice! I think people still prefer checkboxes to the icons, so could we revert that? Otherwise, it is great!

arjunrajlab (commented):
Looks great, thank you!

bruyeret (approved):

PR #733, opened 2024-06-27
Disable auto tags when editing the tools

### LINKED ISSUE #712: Changing tag in the "edit tool" doesn't seem to persist
We have found that if you change the tag added in the "edit tool" dialog, it says that it has the new tag, but then if you click again, the tag is changed back. See this video.

https://github.com/Kitware/UPennContrast/assets/13998824/2cc4810a-01f6-432e-811c-0014b14b2189

bruyeret:
Close #712


arjunrajlab:
Looks good to me!

luciemac (approved):

PR #734, opened 2024-06-28
Fix snapshot refreshing page on creation with a small refactor

### LINKED ISSUE #716: Page reloads when making snapshots
I have found that when making a snapshot, the page will refresh. Sometimes, it happens when you make the first snapshot. Sometimes it happens when you make the second snapshot in a row. It is a little hard to reproduce, but I have seen it several times. Perhaps it requires a new dataset, not sure.

I also wonder if it happens when you first start a server.

bruyeret:
Close #716

### DIFF
--- a/src/components/FileManagerOptions.vue
+++ b/src/components/FileManagerOptions.vue
@@ -46,7 +46,7 @@
         <v-card>
           <v-card-title> New name </v-card-title>
           <v-card-text>
-            <v-form @submit="rename">
+            <v-form @submit.prevent="rename">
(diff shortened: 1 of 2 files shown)


luciemac (approved):

arjunrajlab (changes requested):
I am still getting the error :(. To reproduce, I ran `npm run dev` and then closed the tab, opened a new tab and went to localhost:5173, then made a new snapshot. Same page reload happened.

luciemac (approved):

arjunrajlab (approved):
It works, thanks!

PR #735, opened 2024-06-28
Tool hotkey fix and tooltip

### LINKED ISSUE #715: Hotkey persists to new tools; also add hotkey to tooltip
If you add a hotkey to a new tool, then add a new tool, the new tool has that hotkey assigned already as default in the dialog.

Also, while we're at it, it would be nice to show the hotkey in the tooltip (like add [Hotkey: "g"] to the below). 

image

bruyeret:
Close #715


luciemac (approved):

arjunrajlab (approved):
Looks great!

PR #738, opened 2024-06-28
Small fixes

bruyeret:
- Fix blinking of user menu (login screen)
- Fix loading of recent datasets sometimes not happening
- Disable tooltip when there is nothing to show in the tooltip


arjunrajlab (approved):
Works great, thank you!

luciemac (approved):

PR #741, opened 2024-07-03
Use private folder by default

### LINKED ISSUE #739: Put files in private folder by default
I have a feeling people would expect their files to be private by default. Perhaps that should be the first location that is selected by default as a place to put files instead of public.

bruyeret:
The default is now the Private folder for everything.
Also fix the datasets that are uploaded in the private folder, as the images didn't send the `girder-token` cookie because of the `crossOrigin` setting of images that was badly set.
Also fix the default CORS allowed origins in girder, because a simple wildcard is not compatible with `crossOrigin="use-credentials"`.

For exisiting builds, that use a different IP for the front end and girder:
In girder (http://localhost:8080/):
- Go to Admin Console -> Server Configuration
- Go to Advanced Settings at the bottom and find `CORS Allowed Origins` setting
- Replace the `*` with the full adress of the backend (for example `http://localhost:5173` or `http://localhost:*`), or even `**` if it is in a private network
- Save the settings
image

Close #739

### DIFF
--- a/devops/girder/provision.py
+++ b/devops/girder/provision.py
@@ -21,7 +21,7 @@ def provision(opts):
         # Allow cross origin requests, but since this may be undesired,
         # only do it if we are creating an admin user so it doesn't happen
         # on subsequent starts.
-        Setting().set("core.cors.allow_origin", "*")
+        Setting().set("core.cors.allow_origin", "http://localhost:*")
 
         # Increase the allowed files that can be cached, as this is used for
         # tiled frames
--- a/src/components/CustomFileManager.vue
+++ b/src/components/CustomFileManager.vue
@@ -184,8 +184,8 @@ export default class CustomFileManager extends Vue {
 
   @Watch("isLoggedIn")
   async fetchLocation() {
-    const publicFolder = await this.store.api.getUserPrivateFolder();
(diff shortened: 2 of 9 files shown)


luciemac (approved):

arjunrajlab:
I think this mostly works. It worked on my AWS instance. When running the front end on localhost:5173, I had to set the CORS Allowed Origins to localhost:*. Is that right? I thought I only needed to do that if I had a static IP?

arjunrajlab:
Also, perhaps unrelated, but when I make an AWS S3 assetstore, I get a bunch of 400 errors. The files themselves are actually placed in the assetstore, but it seems that it can't read them. I was able to read them from the AWS CLI, so I don't think it's a permissions issue.

arjunrajlab:
Ah okay. So for a dev install, someone will have to change that in the Girder settings? Which is fine, I just have to add that to the instructions. I am wondering if it might be better is the default was localhost:* so that it runs out of the box, and then we give instructions to change to * if you want to run on a static IP or on AWS? I think most people who are installing for the first time would be running via localhost. But either is really fine.

arjunrajlab (approved):
It works, thank you!

manthey (approved):

PR #742, opened 2024-07-03
File manager options

### LINKED ISSUE #740: Can we come up with a way to go into the folder of a dataset in the navigator and download a file?
Let's say that I put the results of an analysis within a folder corresponding to a dataset. Right now, there is no way from within the UI to see the files within the dataset or to download them. Can we allow for a listing of the items within the dataset and also for downloading of them? Perhaps in the dataset view?

image

bruyeret:
Add the "Download" option for items and files and the "Browse" option for datasets

Close #740

### DIFF
--- a/src/components/CustomFileManager.vue
+++ b/src/components/CustomFileManager.vue
@@ -148,6 +148,7 @@ export default class CustomFileManager extends Vue {
   })
   useDefaultLocation!: boolean;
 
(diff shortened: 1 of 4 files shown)


arjunrajlab (approved):
Works great!

luciemac (approved):

PR #746, opened 2024-07-09
Shortened timeout for SAM to 1s

arjunrajlab:


bruyeret (approved):
This timeout will apply for any addition of annotation, not only SAM annotations
I set it to 3s initially because it is the time I thought a user would take to reach the `Undo` button at the bottom right of the screen if it doesn't use the shortcut `ctrl+z`

arjunrajlab:
Yes, sorry, I should have said all annotations instead of SAM annotations! I did some user testing and playing with a bunch of different timings, and I think 1s turned out to be ideal for most people in that it matched their "flow", if that makes sense.

PR #749, opened 2024-07-11
Tag cloud picker filter

arjunrajlab:
Makes a clickable tag cloud for showing and hiding various annotations, a much requested feature.


src/utils/annotation.ts
```diff
@@ -199,3 +199,14 @@ export function tagFilterFunction(
...
+export function tagCloudFilterFunction(
+  inputTags: string[],
+  filterTags: string[],
+  exclusive: boolean,
+) {
+  if (exclusive && inputTags.length !== filterTags.length) {
+    return false;
+  }
+  return inputTags.some((inputTag) => filterTags.includes(inputTag));
+}
```
bruyeret, line 212: I don't understand what this function should do.
For example, if you you have:
- `inputTags = ["x", "y", "z"]`
- `filterTags = ["a", "b", "c", "x"]`
It will return `true` when `exclusive` is `false` and return `false` when `exclusive` is `true`.

Also, the logic for tags should be consistent through all the application, so why not changing the logic of `tagFilterFunction` instead of creating a new function (`tagCloudFilterFunction`)?

src/components/TagCloudPicker.vue
```diff
@@ -0,0 +1,81 @@
...
+    for (const tool of this.store.tools) {
+      if (tool.values?.annotation?.tags) {
+        const tags = tool.values.annotation.tags;
+        for (const tag of tags) {
+          tagSet.add(tag);
+        }
+      }
+    }
+    return Array.from(tagSet);
+  }
```
bruyeret, line 59: This function is copy pasted from the tag picker, maybe we should instead have it computed in the annotation store and in the store

src/components/TagCloudPicker.vue
```diff
@@ -0,0 +1,81 @@
...
+    this.tags = [...this.tagList];
+  }
+
+  selectNone() {
+    this.tags = [];
+  }
+}
+</script>
+
+<style scoped>
```
bruyeret, line 71: This is a good thing to use scoped styles, you should also use `lang="scss"` like this: `<style lang="scss" scoped>`.

src/components/TagCloudPicker.vue
```diff
@@ -0,0 +1,81 @@
...
+.selected-chip {
+  border-color: #ffffff !important;
+}
+
+.v-chip {
+  transition:
+    background-color 0.3s,
+    color 0.3s;
+}
```
bruyeret, line 80: Vuetify components can not be styled in scoped style blocks.
You should not style them in a non-scoped block either, as it will apply to the whole application (it is done in specific places, but you should generally not have to do it).
So in this case the `v-chip` style will not have any effect and you can just remove it.

bruyeret (changes requested):
I pushed a small commit in AnnotationFilers.vue, make sure to pull before taking this review into account, otherwise you will have to rebase 😉

bruyeret:
Also I completely removed the shape filter from the code base, as you removed the UI to use it.

src/utils/annotation.ts
arjunrajlab: Good question! What I found is that people wanted to be able to basically have a little "pad" on which they could click to turn on and off various annotations (mostly, people use a single tag for annotations). This logic allows them to easily turn them on and off. When exclusive is false, it runs in basically an "or" mode, in which selecting a particular tag will show every annotation that has any of those tags. When exclusive is true, it runs like a real "exclusive" mode, in which show only objects that are an exact match for the selected tags. Honestly, not many/any people use the exclusive mode anyway, but I felt like there may be moments when someone wants to very specifically find a set of annotations, so I left that in. TL;DR: I came up with logic that matches user expectations with a feature many people had been requesting (and that I had noticed people outside the lab were looking for when using it).

src/components/TagCloudPicker.vue
arjunrajlab: Yes, that would be fine! It seems like a function that should be generally available.

src/components/TagCloudPicker.vue
arjunrajlab: Will do!

src/components/TagCloudPicker.vue
arjunrajlab: Good to know, thanks!

src/utils/annotation.ts
bruyeret: Then I think we should not create a new function, but apply this logic to all the codebase.
The logic should work as you expect when `exclusive` is `false`, but when `exclusive` is `true`, it doesn't really do what you expect
Do you want me to fix it?

src/components/TagCloudPicker.vue
arjunrajlab: Actually, can you make this edit? I'm not sure exactly how best to do it, so I'd probably mess it up :).

arjunrajlab:
@bruyeret There is another problem I'm not sure what to do about. Let's say you create a new tool with a new tag. The tag will now show up in the tag cloud, but by default is not selected, so the new annotations you make do not display. I suppose we could have an "active tags" variable that contains what is selected in the tag cloud, and when a new tool is created, it makes that new tag active by default. How do you think we could implement that? I was thinking of having an activeTags variable in the annotationStore, and am happy to implement that, just let me know!

src/utils/annotation.ts
arjunrajlab: I thought about that a lot, and I think the logic for the tag cloud is actually quite different than what is expected in the tag picker, which is why I made a separate function.

bruyeret:
Note: Add the component at the end of the right panel, below the layers

arjunrajlab:
^ left panel, I think.

arjunrajlab:
Another minor request: for the tag cloud, can we implement a filter like the one we have for annotation IDs?

image

Some use cases might have a LOT of tags, and being able to search for the one you want to toggle on and off would be very helpful.

arjunrajlab:
I think it is still missing the ability to pull from the filters. Like, it should show all as active when there is no tag filter, and when you select all, it should clear the filter. Then, if we add a new annotation with a new tag, then it should show that tag as active because there is no filter.

arjunrajlab:
It is looking very good! One minor last point: https://github.com/Kitware/UPennContrast/pull/749#issuecomment-2233497985

bruyeret:
I just pushed the search zone @arjunrajlab 👍

arjunrajlab:
Looks great!

PR #755, opened 2024-07-18
Fix import of annotation properties

### LINKED ISSUE #751: JSON import doesn't reflect properties
Let's say I export a JSON with a set of properties and property values. Then I import that into a "clean" dataset. It doesn't seem to show any of the properties in the property panel. If I export the JSON from the new dataset, the properties are there, so they exist, but they just aren't recognized by the UI. Page refresh doesn't seem to help.

### LINKED ISSUE #752: I'm wondering: does the JSON import code use the new multiple property value endpoint?
I'm not sure I fully understand this code, but it seems like it uploads the property values one at a time? That would be very slow, I think.

https://github.com/Kitware/UPennContrast/blob/f7408f7811d437464d4cbbdbe0b73f2ab3b4df62/src/components/AnnotationBrowser/AnnotationImport.vue#L312C13-L317C1

bruyeret:
I fixed the import of annotation properties
Overwriting annotations works on my side, are you sure that annotations are not removed?

Close #751
Close #752

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationImport.vue
+++ b/src/components/AnnotationBrowser/AnnotationImport.vue
@@ -283,7 +283,7 @@ export default class AnnotationImport extends Vue {
     if (this.importProperties) {
       for (const oldProperty of this.properties) {
         const newPropertyPromise =
-          this.store.propertiesAPI.createProperty(oldProperty);
+          this.propertyStore.createProperty(oldProperty);
         const idx = propertyPromises.push(newPropertyPromise) - 1;
         propertyOldIdToIdx[oldProperty.id] = idx;
       }
--- a/src/components/AnnotationBrowser/AnnotationImport.vue
+++ b/src/components/AnnotationBrowser/AnnotationImport.vue
@@ -1,5 +1,5 @@
 <template>
-  <v-dialog v-model="dialog">
+  <v-dialog v-model="importDialog">
     <template v-slot:activator="{ on, attrs }">
       <v-btn
         v-bind="{ ...attrs, ...$attrs }"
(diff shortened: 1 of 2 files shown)


luciemac (approved):

arjunrajlab (dismissed):
When you do overwrite, it doesn't overwrite the Property Worker definitions, so they get duplicated if you repeatedly import with overwrite.

bruyeret:
Oh ok so you would like to "merge" the properties that are exactly the same (except for their ID)?
We could add another checkbox "merge identical properties" that would be checked by default
It would avoid creating a new property when there is already one with the same settings

arjunrajlab:
It is a good question! I was actually thinking they would be completely replaced if the user selects "overwrite". I think that would be the expected behavior. I don't think we need to merge, although I suppose we could give the option to merge the workers.

bruyeret:
I added a commit to fix it in the same way as for annotations, is it better?

arjunrajlab:
That works great! Here's another small issue.
Here's a snippet from a sample JSON:
```

        {
            "id": "abc12345",
            "name": "Magicness"
        },
        {
            "id": "def54321",
            "name": "Evilness"
        }
    ],
    "annotationPropertyValues": {
        "6692561b6fbb27ec7b06ea42": {
            "669256da6fbb27ec7b06ea98": {
                "Area": 5840,
                "Centroid": {
                    "x": 678.4908105022831,
                    "y": 279.0122431506849
                },
                "Circularity": 0.8205950927590248,
                "Compactness": 0.8205950927590248,
                "Convexity": 0.9756912538635034,
                "Eccentricity": 0.8007082748936661,
                "Elongation": 0.3771186440677967,
                "Fractal_Dimension": 1.3146444632102863,
                "Perimeter": 299.0521284760509,
                "Rectangularity": 0.7680445059379685,
                "Solidity": 1.0075347154358751
            },
            "6692570b6fbb27ec7b06eaa3": {
                "25thPercentileIntensity": 1234,
                "75thPercentileIntensity": 12345,
                "MaxIntensity": 19999,
                "MeanIntensity": 100000,
                "MedianIntensity": 3333,
                "MinIntensity": 8888,
                "TotalIntensity": 123456,
                "FakeIntensity": 76543
            },
            "abc12345": {
                "Magicness": 100
            }
        },
        "669256326fbb27ec7b06ea54": {
            "6692570b6fbb27ec7b06eaa3": {
                "25thPercentileIntensity": 4843,
                "75thPercentileIntensity": 27636,
                "MaxIntensity": 47803,
                "MeanIntensity": 15642.933621933622,
                "MedianIntensity": 8704,
                "MinIntensity": 3333,
                "TotalIntensity": 10840553
            },
            "def54321": 1000
        }
    }
}
```

I can import it just fine. However, it throws some Vue errors because I didn't define a lot of the variables in the property definition that are specific to workers. Consequently, it doesn't have a "delete property" button. We can decide what to do about it (including maybe nothing).

image

PR #756, opened 2024-07-19
Add feature to move folder to a different assetstore

### LINKED ISSUE #477: Add a way to remove the underlying data, but keep all the annotations and rest of the dataset
As our server gets full, it would be nice to be able to remove the underlying data but keep the rest of the dataset, including annotations and configurations and so forth. Then, if someone needs to look again later, they can just reupload the data into the dataset. Not sure exactly how to do this, but it seems that Girder has the appropriate functionality.

bruyeret:
Rebuilding girder is needed
`docker-compose build girder`

Close #477

### DIFF
--- a/devops/girder/Dockerfile
+++ b/devops/girder/Dockerfile
@@ -11,7 +11,7 @@ WORKDIR /src
 RUN git clone https://github.com/girder/large_image.git
 WORKDIR /src/large_image
 RUN pip install -e .[all] -r requirements-test.txt --find-links https://girder.github.io/large_image_wheels
-RUN pip install girder[mount] girder girder-user-quota
+RUN pip install girder[mount] girder girder-user-quota git+https://github.com/DigitalSlideArchive/import-tracker.git
 # RUN pip install .[all] ./girder[tasks] girder[mount] --find-links https://girder.github.io/large_image_wheels
 
 COPY ./provision.py /src/provision.py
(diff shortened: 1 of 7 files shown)


luciemac (approved):

arjunrajlab:
I _think_ this works, but it's a little hard to check because I can't tell which assetstore a particular folder is located. Is there some easy way to indicate that somewhere? We can discuss how much time this might take.

bruyeret:
A folder doesn't have an assetstore, only the files have a designated assetstore
So if you want to know "where" a folder is, you will have to iterate on all its files and there can be a lot of them

bruyeret:
The good news is that when moving a folder from an assetstore to another, the files that are already at the right location will not be moved at all, which means that it will be very fast to move a folder that already has its files in the right assetstore

arjunrajlab:
Sorry, I guess what I meant is that I don't know how to tell which assetstore an item is in, even through girder. Is there some place to see that information? It is fine if not, just curious.

bruyeret:
Oh ok, so you can do it like this:
- In girder, open your item
- Go at the bottom of the page in the "Files & links" section and click the "i" icon next to the file's size
- Copy its "Unique ID"
- Go to the "Wep API" (at the very bottom of the page, there is a link)
- Open the "file" section and open the second API endpoint: `GET /file/{id}`
- Paste the Unique ID of the file and click "Execute"
- In the "Response body", copy the "assetstoreId" property
- Open the "assetstore" section and open the third API endpoint `GET /assetstore/{id}`
- Paste the ID of the assetstore and click "Execute"
- The "Response body" should contain the property "name"

arjunrajlab (approved):
Works!

PR #759, opened 2024-07-29
Add a support for annotation worker shape label

### LINKED ISSUE #744: Annotation worker does not read the "annotationShape" label
Right now, if you choose a worker that has `--label "annotationShape=polygon"`, it doesn't seem to obey it. The problem then is that the default naming convention is sort of weird. For instance, here is cellpose (which should make blobs), but is automatically tagged with "Brightfield point" and then auto-named "Brightfield point (Automated)", even though it should be a blob.

image

bruyeret:
Use `annotationShape` as for Property workers

Close #744

### DIFF
--- a/src/tools/creation/ToolCreation.vue
+++ b/src/tools/creation/ToolCreation.vue
@@ -146,7 +146,7 @@ export default class ToolCreation extends Vue {
       return;
     }
     const toolNameStrings: string[] = [];
-    if (this.toolValues?.annotation) {
+    if (this.toolValues?.annotation?.tags) {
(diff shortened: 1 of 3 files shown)


arjunrajlab (commented):
So I tried running it for the cellpose worker with `--label "annotationShape=polygon"`. Screenshot below. It seems to recognize that it is a blob, but the auto-tagging isn't working.

image

Also, then, when I make the worker, it doesn't seem to recognize the container:

image

Any idea what is happening?

bruyeret:
Yes, sorry I missed some small issues ! Everything should be fixed now 👍

PR #760, opened 2024-07-29
Color selected

### LINKED ISSUE #750: Add a "Color selected" button to annotation list panel
It would be nice to be able to change the colors of selected annotations (much requested feature). I think adding a "Color selected" button here would be good:

image

It would just bring up a color picker dialog and assign the colors to the selected annotations. Also, it might be good to have a "Clear color" option in the dialog to remove the color property so that the annotation goes back to inheriting the color from the layer.

bruyeret:
You need to rebuild girder because of the addition of an endpoint to edit multiple annotations at the same time

Close #750

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
@@ -56,6 +56,7 @@ class Annotation(Resource):
         self.route("GET", (), self.find)
         self.route("POST", (), self.create)
         self.route("PUT", (":id",), self.update)
+        self.route("PUT", ("multiple",), self.updateMultiple)
         self.route("POST", ("compute",), self.compute)
         self.route("POST", ("multiple",), self.createMultiple)
         self.route("DELETE", ("multiple",), self.deleteMultiple)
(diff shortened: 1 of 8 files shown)


arjunrajlab (approved):
Work great, thank you!

arjunrajlab (changes requested):
This PR has a subtle bug in it. When I hit the Girder endpoint for multiple annotations:
`upenn_annotation/multiple` (which happens from e.g. the random square worker), I'm getting a response of `{"message": "data must contain ['color'] properties", "type": "validation"}`. This bug is affecting all our workers. I think somewhere a validation was added? Color should be optional for these workers.

bruyeret:
Oh yes, I didn't think about it. I made the color mandatory, but it can be `null`.

When you update an annotation, it finds the annotation with this ID, do the changes you asked for, and validate the new annotation. If it didn't have a color property, then the new annotation won't have a color either, hence the validation error.

So this error could happen:
- If you create an annotation from a worker without a color
- If you update an old annotation that didn't have a color

We could discuss a solution on Friday 👍

bruyeret:
It should all work fine now!
I made it compatible with the old annotations that don't have a color @arjunrajlab

arjunrajlab (approved):

arjunrajlab:
So I'm still getting this error:

```
[2024-09-05 15:49:31,299: WARNING/ForkPoolWorker-16] girder_client.HttpError: HTTP error 400: POST http://girder:8080/api/v1//upenn_annotation/multiple
Response text: {"message": "data must contain ['color'] properties", "type": "validation"}
```
I pulled the branch and rebuilt the worker (this was the cellpose worker). And I rebuilt girder. Is there something else I could be missing?

bruyeret:
Did you pull correctly? I forced push, so you could have issues with that
I am sure that the validation doesn't need a color anymore, I even specified it in a comment:
```
# color is optional (legacy, equivalent to null)
```

bruyeret:
You can:
```
git fetch
git reset origin/color-selected --hard
```

arjunrajlab:
That worked, thank you!

PR #764, opened 2024-08-20
Updated API to add getPropertyById

arjunrajlab:
Made a getPropertyById to get property information from a propertyId.

Also, fixed the naming convention for some paths that are for property values.


bruyeret (approved):
Perfect, looks good to me!

PR #769, opened 2024-09-05
Fix connection display

bruyeret:
Coordinates should not have a z coordinate
You can merge right away if important


arjunrajlab (approved):
Works great, thank you!

PR #771, opened 2024-09-05
Fix girder crash when no Anthropic API key and handling of errors

### LINKED ISSUE #770: Girder exits if you don't have the ANTHROPIC_API_KEY environment variable set
GIrder seems to quit out if the ANTHROPIC_API_KEY environment variable is not set. I'm not sure why, but it would be good to raise a warning and disable the chat rather than crashing entirely.

bruyeret:
I also added a better handling of errors so that users can keep sending messages even when an error occurs

Close #770


arjunrajlab (approved):
Works well, thank you!

PR #777, opened 2024-09-15
Implement signup

arjunrajlab:
This implements a signup page:

1. It removes the Girder domain if the Girder domain environment variable is set. Otherwise, it shows an input field like before.
2. It allows you to click on "sign up".
3. It will inform you if the username or email is already in use.
4. Returns to the login page after sign up is successful while the user has to go and complete their email verification.

@bruyeret Currently I am getting a typescript error for "error" in UserAPI.ts. Not sure what to do there. I tried a few things but couldn't resolve it.


bruyeret:
I removed the UserAPI file, we can recreate it later if needed, but the rest client managment should be done in the store
Also the domain should be configurable when signing up, and the logic was broken there

PR #778, opened 2024-09-16
More comprehensive system prompt and prompt caching

arjunrajlab:
The more comprehensive system prompt should give a better experience. In theory, the prompt caching should be much more cost-effective and faster. I didn't notice much difference, but we will see.


bruyeret (approved):
That is a lot of context! The chat bot will be very helpful!

Issue #766, opened 2024-08-27
Changing advanced options while making a new tool causes the tags to revert to default auto-fill


arjunrajlab:
Let's say I make a tag like this:

<img width="985" alt="image" src="https://github.com/user-attachments/assets/8b4941b9-1a18-4a00-8e9e-da425368ad3e">

Then I open the Advanced options and change, for example, the color. Then it changes the tag back to "DAPI blob", which is the auto-name tag:

<img width="983" alt="image" src="https://github.com/user-attachments/assets/8c0241cd-7ca6-41d8-be0e-5d510e690a4d">

arjunrajlab:
@pchoisel as discussed at our meeting, I think we can do the quick and dirty one line fix for now, then come back to fix this with a cleaner refactoring later when there is time.

PR #785, opened 2024-09-27
Configure automated tool creation panel

### LINKED ISSUE #784: Allow more programmatic control of the tool menu for automated workers
Automated workers still use this same tool dialog:

image

However, in some instances, the various options presented may not be relevant. For example, for the "Connect sequential" worker, which connects objects to each other, the "Annotation configuration" and "Advanced options" are not relevant. Perhaps we can control these with labels on the docker image. For instance: "annotationConfigurationPanel=False" and "advancedOptionsPanel=False" would turn these two panels off. If they are not set, then it uses the default.

Also, the naming could be overridden. Right now, it tries to autoname based on the tags. However, for connect sequential, this again is not right. We could control this with "defaultToolName=Connect sequential" or something like that, and if the label is not set, then it just does the default behavior.

arjunrajlab:
Now, if you label an automated worker docker container like this:

```
LABEL isUPennContrastWorker="" \
      isAnnotationWorker="" \
      interfaceName="Connect to Nearest" \
      interfaceCategory="Connections" \
      description="Connects objects to each other based on distance" \
      advancedOptionsPanel="False" \
      annotationConfigurationPanel="False" \
      defaultToolName="Connect to Nearest"
```
it will turn off the Advanced Options Panel, configuration panel, and will auto-name with "Connect to Nearest".

Closes #784

### DIFF
--- a/src/store/AnnotationsAPI.ts
+++ b/src/store/AnnotationsAPI.ts
@@ -234,8 +234,8 @@ export default class AnnotationsAPI {
     const datasetId = dataset.id;
     const { id, name, type, values } = tool;
     const image = values.image.image;
-    const { annotation, connectTo, jobDateTag } = values;
-    let tags = annotation.tags;
+    const { annotation = {}, connectTo = {}, jobDateTag } = values;
+    let tags = annotation.tags ?? [];
     if (jobDateTag) {
       const date = new Date(Date.now());
       const timeZone = date.getTimezoneOffset() / 60;
@@ -250,9 +250,9 @@ export default class AnnotationsAPI {
       tags = [...tags, computedTag];
     }
     const connectToLayerId = connectTo.layer;
-    const connectToLayer = layers.find(
-      (layer) => layer.id === connectToLayerId,
-    );
+    const connectToLayer = connectToLayerId
+      ? layers.find((layer) => layer.id === connectToLayerId)
+      : null;
     const connectToChannel = connectToLayer ? connectToLayer.channel : null;
     const augmentedConnectTo = { ...connectTo, channel: connectToChannel };
     const params = {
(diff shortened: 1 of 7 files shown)


arjunrajlab:
Hmmm, I think I'm getting some error now. I think the issue is that, without these interface elements, it is missing some parameters that are required when making the call to run the worker. @pchoisel @bruyeret Do you have any insight?

arjunrajlab:
I think I managed to fix it by adding in some defaults?

src/store/properties.ts
```diff
@@ -492,6 +492,27 @@ export class Properties extends VuexModule {
...
+      return labels ? labels.advancedOptionsPanel !== "False" : true;
+    };
+  }
+
+  get showAnnotationConfigurationPanel() {
+    return (image: string) => {
+      const labels = this.workerImageList[image];
+      return labels ? labels.annotationConfigurationPanel !== "False" : true;
+    };
+  }
```
bruyeret, line 508: I you want to make these checks case insensitive, you can do:
`myLabel.toLowerCase() !== "false"` instead of `myLabel !== "False"`

src/tools/creation/ToolConfiguration.vue
bruyeret, line 1: I think that Paul is currently working on this file for the issue #766 
What do you think @pchoisel?

src/store/AnnotationsAPI.ts
bruyeret, line 1: Why are the tool values sometimes not containing an annotation or a connect to attribute? Is there a way to create a worker without these values attributes?

bruyeret (commented):
@pchoisel should review the file I where I tagged him, otherwise LGTM!

pchoisel:
Looks good to me, you might want to apply the lowercase comments by Thibault.
I've also modified this file, but I'll just apply your modifications when I finish mine, so no worries.

PR #786, opened 2024-09-29
Add create property function

arjunrajlab:


devops/girder/annotation_client/annotation_client/annotations.py
```diff
@@ -403,3 +414,44 @@ def getPropertyValuesForAnnotation(self, datasetId, annotationId):
...
+        )
+
+    def getConfigurationsByDatasetId(self, datasetId):
+        """
+        Get the configurations for a dataset
+        :param str datasetId: The id of the dataset
+        """
+        return self.client.get(
+            PATHS["configurations_by_dataset"].format(datasetId=datasetId)
+        )
```
bruyeret, line 457: These two functions don't return exactly what they say they return:
- The first one returns the whole configuration item (let's call it `item`). So you will have to do `item.meta.propertyIds` to get the property ids, which is not even the properties themself.
- The second one returns a list of dataset views, which contain a `configurationId` but not the configuration.

To make the things a little bit cleaner you can rename them:
- `getItemById` and rename `PATHS["properties_by_configuration"]` to `PATHS["item_by_id"]`
- `findDatasetViewsByDatasetId` and rename `PATHS["configurations_by_dataset"]` to `PATHS["dataset_views_by_dataset"]`

bruyeret (commented):
Some renaming to do, but otherwise LGTM!

bruyeret (approved):

PR #789, opened 2024-10-04
Use PNPM

bruyeret:
Change the package manager from NPM to PNPM.
This branch was in the repo for a long time, it took me 5min to rebase on master 👍 

How to use pnpm:
- Install PNPM by doing `npm i -g pnpm`
- Instead of doing `npm run xxx`, it is now `pnpm xxx`. For example: `npm run dev` becomes `pnpm dev` and `npm install` becomes `pnpm install`.

It comes with cool advantages and no drawbacks:
- The version of Node is enforced and the same for everyone. The production server can't crash because it uses a different version of node that is different from the developer.
- The packages that are not in the package.json can't be used in the project. This avoids dependency on dependecies.
- There are no version conflicts anymore. If we want to bump axios because it is outdated, no other package can interfere.


arjunrajlab (approved):
Looks good!

PR #793, opened 2024-10-05
Allow no tag property creation

### LINKED ISSUE #787: Add "Create and compute" button
Right now, we first hit "Submit", then go down and click the Play button to compute a property. For many such properties, people probably want to compute right away. I think we can have 3 buttons: "Create and compute property" "Create property" and "Cancel".

image

### LINKED ISSUE #788: Allow property computation for all objects with a particular shape
Right now, you have to select a tag in order to start making a property. (It then automatically brings up the right shape, which is good behavior.) However, there are some times when you might just want to get the property for all objects, for things like e.g. Area and other blob metrics. What would be a good interface for that?

We could also put a "Compute for all blob objects" checkbox in the property worker interface, but that's a little bit odd. But honestly, that might fit better with people's workflow anyway.

image

arjunrajlab:
This PR allows the user to create properties even if no tag is specified. This helps if you want to compute, say, area for all polygon objects at once.

Closes #788 
Closes #787

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationProperties/PropertyCreation.vue
+++ b/src/components/AnnotationBrowser/AnnotationProperties/PropertyCreation.vue
@@ -276,7 +276,6 @@ export default class PropertyCreation extends Vue {
     this.dockerImage = null;
     this.originalName = "New Property";
     this.isNameGenerated = true;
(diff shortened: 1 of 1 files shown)


pchoisel:
Look good to me !

PR #794, opened 2024-10-08
Add delete unselected button

### LINKED ISSUE #762: Need a "keep only selected" button
It would be useful to have a button that deletes everything that is NOT selected. Right now, we have a "delete selected" button, but nothing to delete everything that is the opposite (i.e., keep only that which is selected).

arjunrajlab:
Allows you to delete all annotations that are not currently selected.

Closes #762

### DIFF
--- a/src/store/annotation.ts
+++ b/src/store/annotation.ts
@@ -559,6 +559,24 @@ export class Annotations extends VuexModule {
   }
 
   @Action
+  public async deleteUnselectedAnnotations() {
+    const unselectedIds = this.annotations
+      .filter(
+        (annotation) =>
+          !this.selectedAnnotations.some(
+            (selected) => selected.id === annotation.id,
+          ),
+      )
+      .map((annotation) => annotation.id);
+
+    if (unselectedIds.length === 0) {
+      return;
+    }
+
+    await this.deleteAnnotations(unselectedIds);
+  }
+
+  @Action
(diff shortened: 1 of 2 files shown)


src/store/annotation.ts
```diff
@@ -558,6 +558,24 @@ export class Annotations extends VuexModule {
...
+        (annotation) =>
+          !this.selectedAnnotations.some(
+            (selected) => selected.id === annotation.id,
+          ),
+      )
+      .map((annotation) => annotation.id);
+
+    if (unselectedIds.length === 0) {
+      return;
+    }
```
pchoisel, line 574: I would put this at the beginning of `deleteAnnotations`  
I think it would be a better responsibility share

src/store/annotation.ts
```diff
@@ -558,6 +558,24 @@ export class Annotations extends VuexModule {
...
+
+    await this.deleteAnnotations(unselectedIds);
```
pchoisel, line 576: Why do you need to await the result ?

src/store/annotation.ts
arjunrajlab: Good point!

src/store/annotation.ts
arjunrajlab: Good point, removed.

PR #795, opened 2024-10-09
Slight tweak to the tag picker interface

arjunrajlab:
Made the focus return after adding each tag. Many users were sort of confused or annoyed a the tag picker automatically going to add more tags.

### DIFF
--- a/src/components/TagPicker.vue
+++ b/src/components/TagPicker.vue
@@ -34,7 +34,6 @@
 import { Vue, Component, VModel, Prop } from "vue-property-decorator";
 import store from "@/store";
 import annotationStore from "@/store/annotation";
-import Vue from 'vue';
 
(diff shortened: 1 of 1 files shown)


src/components/TagPicker.vue
```diff
@@ -56,5 +58,18 @@ export default class TagPicker extends Vue {
   get layers() {
     return this.store.layers;
   }
+
+  $refs!: {
+    combobox: any;
```
pchoisel, line 63: Using `HTMLFormElement` for the type would be preferable

pchoisel:
LGTM otherwise

src/components/TagPicker.vue
arjunrajlab: Thanks!

PR #796, opened 2024-10-12
Fix property creation logic

arjunrajlab:
Fixes the property creation logic to guide the user through property creation more clearly. It also fixes some of the reset logic, which was broken.


pchoisel:
Looks good to me

PR #797, opened 2024-10-12
Hide preview button

arjunrajlab:
Most workers do not have preview functionality, but the preview button is always showing. This PR hides the preview button unless the docker image has a "hasPreview" label set to True.


pchoisel:
LGTM

PR #798, opened 2024-10-13
Added notes option to worker interface

arjunrajlab:
You can now have a notes field like this:

```
'Notes on using connect sequential': {
            'type': 'notes',
            'value': 'This tool connects objects sequentially across time or z-slices.'
                     'It is useful for connecting objects that are moving or changing over time or z-slices.'
        },
```


pchoisel:
This looks like a great modification, but there is a lot of redundant code.  
A quick and easy way to fix it would be to use the new `v-tooltip` directive from https://github.com/Kitware/UPennContrast/pull/804  
Since we have control of the actual tooltip, we could add a new prop to the `NimbusTooltip` component to display it or not, which would make this feature much shorter.

PR #799, opened 2024-10-13
Add tooltips to interface

arjunrajlab:
Added a few tooltips to main interface items. Many more to go, but I will do those in separate PRs.


arjunrajlab:
Agreed, it is very verbose, currently. Yes, adding the package to this PR would be great! Also, I have noticed that I can't get v-tooltips to work in side panels. But perhaps that can be addressed separately.

PR #802, opened 2024-10-15
BUG: Fix autoTags

### LINKED ISSUE #766: Changing advanced options while making a new tool causes the tags to revert to default auto-fill

pchoisel:
Autotags will only be applied to non-advanced Annotation Configuration

Fixes #766


arjunrajlab (approved):
Haha yes, that is rather dirty, but it gets the job done for now!

PR #804, opened 2024-10-17
ENH: Add better tooltip with examples

pchoisel:
Add a new package that gives a Vue directive to add a tooltip to a component.  
Also fixed the effect it has on v-tab elements by add a `display: flex`


arjunrajlab (approved):
Looks good to me, thanks!

PR #809, opened 2024-10-29
Handle worker error

arjunrajlab:
Added an error alert. It allows you to send an error using sendError just like sendProgress. This way, the worker can communicate an error to the user. We may also want to send warnings to the user through the same interface, perhaps by using a warning flag or something.


pchoisel:
The front-end part looks good to me, but the back-end part is WIP ?

arjunrajlab:
Actually, there is no back end :). The workers just emit error messages to stdout, and then those get parsed by the front end, which is quite nice. There's is an example in the random_square worker. If the front end looks good, then is it okay for me to merge?

PR #810, opened 2024-10-30
Replace tags

arjunrajlab:
When you click "tag selected", it now has an option to replace the existing tags.

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationList.vue
+++ b/src/components/AnnotationBrowser/AnnotationList.vue
@@ -18,27 +18,45 @@
         </v-row>
         <v-row>
           <v-col class="pa-0 mx-1">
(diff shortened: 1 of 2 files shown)


src/components/AnnotationBrowser/AnnotationList.vue
```diff
@@ -18,20 +18,45 @@
         </v-row>
         <v-row>
           <v-col class="pa-0 mx-1">
-            <v-dialog v-model="tagSelectedDialog" width="50%">
+            <v-dialog v-model="addRemoveTagsSelectedDialog" width="50%">
```
pchoisel, line 21: This v-model name really hides what it does  
What about renaming it to  `showAddRemoveTagsSelectedDialog`

pchoisel (commented):
LGTM otherwise

PR #812, opened 2024-10-31
Add adjust opacity option

arjunrajlab:


src/components/ViewerSettings.vue
```diff
@@ -145,6 +153,15 @@ export default class ViewerSettings extends Vue {
...
   }
 
+  get annotationOpacity() {
+    return this.store.annotationOpacity;
+  }
+
+  set annotationOpacity(value: number) {
+    const opacity = typeof value === "string" ? parseFloat(value) : value;
+    this.store.setAnnotationOpacity(opacity);
+  }
```
pchoisel, line 163: I'm not sure how many input events the slider sends when you move it, but since each time you mutate the store, I'm afraid this will slow down the application.  
My recommendation would be to add a `@Debounce(...)` on that function.  
We have examples of this in the codebase.

src/utils/annotation.ts
```diff
@@ -45,13 +45,21 @@ export function getAnnotationStyleFromBaseStyle(
   }
   if (isHovered) {
     style.fillOpacity = 0;
-    style.strokeWidth = 4;
+    style.strokeWidth = 5;
     if (annotationColor) {
```
pchoisel, line 49: Is that `if(annotationColor)` still needed ?

src/components/ViewerSettings.vue
arjunrajlab: Updated, is this right?

pchoisel:
LGTM

PR #815, opened 2024-11-01
Interactive tags in cloud picker

arjunrajlab:


pchoisel:
LGTM  
I'm just afraid the function `handleAnnotationRightClick` will take a lot of time if we have a lot of annotations, especially if the user clicks outside of an annotation.  
Isn't there a function inside GeoJS that can react to an annotation being clicked and directly give you its id ? @manthey

arjunrajlab:
Oh, interesting! I tested it on around 20K annotations and it didn't seem to slow. But I am hoping we can scale to around 1M annotations in the near future.

PR #816, opened 2024-11-05
Removed a duplicate redraw

arjunrajlab:
An extra watch in AnnotationViewer was causing everything to be drawn twice.


pchoisel:
LGTM

PR #821, opened 2024-11-12
Add AWS Global in Girder prod docker-compose

pchoisel:


pchoisel:
@arjunrajlab for information
This will only be used for the AWS prod deployment. It does not affect the standard dev setup

PR #825, opened 2024-11-13
Add color picker to tag chevron

arjunrajlab:
Added color picker to the things you can do based on clicking a tag.


pchoisel:
LGTM

PR #827, opened 2024-11-14
Add annotation id filter

arjunrajlab:


src/store/filters.ts
```diff
@@ -215,6 +262,16 @@ export class Filters extends VuexModule {
         return false;
       }
 
+      // Annotation ID filters
+      const matchesAnnotationIds =
+        enabledAnnotationIdFilters.length === 0 ||
```
pchoisel, line 267: Not important: `!enabledAnnotationIdFilters.length`

pchoisel:
LGTM

src/store/filters.ts
arjunrajlab: Cool, I'll leave as is because I think it's every so slightly more readable.

PR #830, opened 2024-11-15
Added support for rectangle annotations

arjunrajlab:
Added support for rectangle annotation type.

Edited annotation.py, which I think has the schema.


pchoisel:
LGTM

PR #834, opened 2024-11-16
Performance enhancement: alter hover behavior

arjunrajlab:
Annotation List is slow to update with large numbers of annotations because it triggers a redraw from every mouseover event. Change this to just update based on a click if you have over 5000 annotations.

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationList.vue
+++ b/src/components/AnnotationBrowser/AnnotationList.vue
@@ -531,7 +531,7 @@ export default class AnnotationList extends Vue {
   }
 
   hover(annotationId: string | null) {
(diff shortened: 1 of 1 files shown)


src/components/AnnotationBrowser/AnnotationList.vue
```diff
@@ -530,7 +531,12 @@ export default class AnnotationList extends Vue {
   }
 
   hover(annotationId: string | null) {
-    this.annotationStore.setHoveredAnnotationId(annotationId);
+    // Only update the hover if the total number of annotations is less than 1000
```
pchoisel, line 534: Shouldn't that be 5000 ?

pchoisel:
LGTM

PR #835, opened 2024-11-19
Fix large image upload with S3 assetstore

pchoisel:
The getTiles API from large_image expects the field large_image to be set when recovering the tiles, but due to the S3 assetstore, that field can take a bit of time to be set.
Retry that function a few times (10) if it fails to wait for that field to be set.


pchoisel:
Tested on prod, it indeed fixes the issue:
image

arjunrajlab (approved):
LGTM

PR #843, opened 2024-12-07
Speed selection and creation

arjunrajlab:
Annotation selection and creation was very slow with large numbers of annotations because interaction annotations were on the same layer as the main annotations. I created a separate interactionLayer to capture all the interactions, making all these interactions much faster.

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -2001,7 +2001,7 @@ export default class AnnotationViewer extends Vue {
     }
 
     // Remove the selection annotation from layer (do not show the annotation used to select)
-    this.annotationLayer.removeAnnotation(selectAnnotation);
+    this.interactionLayer.removeAnnotation(selectAnnotation);
   }
 
   private async addAnnotationFromGeoJsAnnotation(annotation: IGeoJSAnnotation) {
@@ -2010,7 +2010,7 @@ export default class AnnotationViewer extends Vue {
     }
 
     const coordinates = annotation.coordinates();
-    this.annotationLayer.removeAnnotation(annotation);
+    this.interactionLayer.removeAnnotation(annotation);
 
     // Create the new annotation
     await this.createAnnotationFromTool(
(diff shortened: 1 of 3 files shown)


src/components/AnnotationViewer.vue
```diff
@@ -2223,34 +2224,28 @@ export default class AnnotationViewer extends Vue {
...
-      this.annotationLayer.geoOn(
-        geojs.event.mouseclick,
-        this.handleTaggingClick,
-      );
-    } else {
-      this.annotationLayer.geoOff(
-        geojs.event.mouseclick,
-        this.handleTaggingClick,
-      );
+  handleModeChange(evt: any) {
```
pchoisel, line 2242: Should the type be `IGeoJSMouseState` ?

src/components/AnnotationViewer.vue
```diff
@@ -2223,34 +2224,28 @@ export default class AnnotationViewer extends Vue {
...
+    if (evt.mode === null) {
+      this.refreshAnnotationMode();
     }
   }
 
-  handleModeChange(evt: any) {
+  handleInteractionModeChange(evt: any) {
```
pchoisel, line 2248: `IGeoJSMouseState` ?

src/components/AnnotationViewer.vue
```diff
@@ -2299,14 +2294,14 @@ export default class AnnotationViewer extends Vue {
     return map.unitsPerPixel(map.zoom());
   }
 
-  handleAnnotationChange(evt: any) {
+  handleInteractionAnnotationChange(evt: any) {
```
pchoisel, line 2297: `IGeoJSMouseState` ?

pchoisel (commented):
LGTM otherwise

src/components/AnnotationViewer.vue
arjunrajlab: If I do that, I get a lot of typescript issues. I don't know why Thibault did it with `any`, but it seems to be a purposeful design choice on his part.

pchoisel: Oh yeah, that's because GeoJS is no a Typescript library, so we don't have types for those events.  
We could add more types to `src/store/model.ts` but `any` is fine.

PR #844, opened 2024-12-07
Removed several function calls

arjunrajlab:
Converted to loops in many places.
Removed looped object creation in at least one place.


pchoisel:
LGTM

PR #845, opened 2024-12-07
Fix upload buttons

arjunrajlab:
Fixes a problem where hitting return goes down the wrong path.
Fixes the Upload Data button in the top navigation bar.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -20,7 +20,8 @@
           color="primary"
           class="ml-4"
           @click="goToNewDataset"
-          :disabled="!store.isLoggedIn"
+          :disabled="!store.isLoggedIn || !store.girderUser"
+          :loading="isUploadLoading"
         >
           Upload Data
         </v-btn>
@@ -164,6 +165,7 @@ import propertyStore from "@/store/properties";
 import { logError } from "@/utils/log";
 import { IHotkey } from "@/utils/v-mousetrap";
 import ChatComponent from "@/components/ChatComponent.vue";
+import { IGirderFolder } from "@/girder";
 
 @Component({
   components: {
@@ -211,6 +213,8 @@ export default class App extends Vue {
   chatbotOpen = false;
 
   lastModifiedRightPanel: string | null = null;
+
+  isUploadLoading = false;
 
   fetchConfig() {
     // Fetch the list of available tool templates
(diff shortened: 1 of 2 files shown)


src/App.vue
```diff
@@ -291,6 +291,33 @@ export default class App extends Vue {
...
+
+  async goToNewDataset() {
+    try {
+      // Ensure we have a valid user and location before proceeding
+      if (!this.store.girderUser || !this.store.isLoggedIn) {
+        throw new Error("User must be logged in");
+      }
+
+      // Get the user's private folder
+      const privateFolder = await this.store.api.getUserPrivateFolder();
```
pchoisel, line 303: Calling an API in a button action is fine, but it might delay the UI reaction if the server is loaded.   
The user might think that they didn't click and will click again, calling the API again and loading the server even more.  
To fix this, you can:
 - Add a loader by using the `loading` prop of `v-btn`
 - Or get the privateFolder of the user when they connect and store it in the store

src/App.vue
```diff
@@ -291,6 +291,33 @@ export default class App extends Vue {
       this.toggleRightPanel(null);
     }
   }
+
+  async goToNewDataset() {
+    try {
+      // Ensure we have a valid user and location before proceeding
+      if (!this.store.girderUser || !this.store.isLoggedIn) {
+        throw new Error("User must be logged in");
```
pchoisel, line 299: This is unnecessary since the button will be disabled if `!store.isLoggedIn`  
You may want to change this condition to add `!this.store.girderUser` line 23

src/App.vue
```diff
@@ -291,6 +291,33 @@ export default class App extends Vue {
...
+        name: "newdataset",
+        params: {
+          quickupload: false,
+          defaultFiles: [],
+          initialUploadLocation: privateFolder,
+        } as any,
+      });
+    } catch (error) {
+      logError(error);
+      // Optionally show error to user
```
pchoisel, line 318: Unnecessary comment

src/views/dataset/NewDataset.vue
```diff
@@ -13,6 +13,7 @@
       ref="form"
       v-model="valid"
       :disabled="quickupload && !pipelineError"
+      @submit.prevent="valid && filesSelected && !uploading && submit()"
```
pchoisel, line 16: The `disabled` should be used to make those checks rather than `submit`

PR #850, opened 2024-12-16
ENH: Add plugin feature

pchoisel:
This Vite config modification allows node_modules names nimbus_plugin_XXX to override components in the application.

Also split UserMenu.vue into two components for an easier replacement.


src/layout/UserMenu.vue
```diff
@@ -12,122 +12,24 @@
...
+            <p class="text-subtitle-1">
+              A cloud-based image analysis platform from the Raj Lab at the
+              University of Pennsylvania and Kitware
+            </p>
+          </div>
+          <v-text-field
+            v-if="!isDomainLocked"
+            v-model="domain"
+            name="domain"
+            label="Girder Domain"
```
arjunrajlab, line 28: I have noticed that when you log out, there are two "Girder Domain" fields visible.

image

src/layout/UserMenuLoginForm.vue
```diff
@@ -0,0 +1,222 @@
...
+      <template v-else>
+        <v-text-field
+          v-model="username"
+          name="username"
+          label="Username or e-mail"
+          required
+          prepend-icon="mdi-account"
+          autocomplete="username"
+        />
+        <v-text-field
```
arjunrajlab, line 22: Should we have a "re-enter password" field as well to catch mistakes?

src/layout/UserMenuLoginForm.vue
```diff
@@ -0,0 +1,222 @@
...
+          v-model="signupLastName"
+          name="lastName"
+          label="Last Name"
+          required
+          prepend-icon="mdi-account"
+          autocomplete="family-name"
+        />
+        <v-text-field
+          v-model="signupPassword"
```
arjunrajlab, line 87: Perhaps this is where we should have a "re-enter password" field?

arjunrajlab (commented):
Looks great as far as I can tell, apart from a few minor thing!

src/layout/UserMenu.vue
pchoisel: Whoops, my bad, I removed it.
Now it's up to UserMenu.vue to handle that field, which allows me to simply swap the component UserMenuLoginForm.vue without having to care about the domain

src/layout/UserMenuLoginForm.vue
pchoisel: I added one, but it's pretty crude, it just disables the sign up button if the password and password verification fields don't match  
I can improve it if necessary, but it should probably go it another PR

arjunrajlab:
Looks good! My only comment is that after you log in, this profile popup remains open on the right hand side:

image

pchoisel:
> Looks good! My only comment is that after you log in, this profile popup remains open on the right hand side:
> image

My bad, I thought this was the normal behavior of the app...  
It should be fixed

arjunrajlab:
Perfect, thanks!

PR #856, opened 2024-12-17
Added max file size limit

arjunrajlab:
Reads environment variable VITE_MAX_TOTAL_FILE_SIZE to specify max total file size (over all files) in MB.

### DIFF
--- a/src/views/dataset/NewDataset.vue
+++ b/src/views/dataset/NewDataset.vue
@@ -90,7 +90,8 @@
       exists. Please update the dataset name field and try again.
     </v-alert>
     <v-alert v-if="fileSizeExceeded" text type="error">
-      {{ fileSizeExceededMessage }}
+      Total file size ({{ totalSizeMB }} MB) exceeds the maximum allowed size of
+      {{ maxTotalFileSize }} MB
     </v-alert>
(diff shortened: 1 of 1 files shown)


src/views/dataset/NewDataset.vue
```diff
@@ -372,6 +386,22 @@ export default class NewDataset extends Vue {
   }
 
   filesChanged(files: FileUpload[]) {
+    const totalSize = files.reduce((sum, { file }) => sum + file.size, 0);
+    const maxSizeBytes = this.maxTotalFileSize * 1024 * 1024; // Convert MB to bytes
+
+    if (totalSize > maxSizeBytes) {
+      this.fileSizeExceeded = true;
+      this.fileSizeExceededMessage = `Total file size (${(totalSize / 1024 / 1024).toFixed(1)} MB) exceeds the maximum allowed size of ${this.maxTotalFileSize} MB`;
```
pchoisel, line 394: I would expect to see that error string inside the `v-alert` component.  
Here it's a bit hidden.

You would need to set `totalSize` as a reactive value of the component to do that

arjunrajlab: Good point, thanks!

PR #858, opened 2024-12-19
Add display lock button

arjunrajlab:
People wanted a way to "lock" the view so that it wouldn't move around when doing e.g. SAM segmentation.


pchoisel:
LGTM

PR #859, opened 2024-12-20
Block large SAM annotations

arjunrajlab:
If the SAM preview annotation takes up more than 70% of the screen (X or Y), then don't show the preview.

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -422,6 +422,23 @@ export default class AnnotationViewer extends Vue {
     if (!vertices) {
       return;
     }
+
+    // Get the current screenshot dimensions
+    const viewBounds = this.map.bounds();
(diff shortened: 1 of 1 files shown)


src/components/AnnotationViewer.vue
```diff
@@ -422,6 +422,23 @@ export default class AnnotationViewer extends Vue {
     if (!vertices) {
       return;
     }
+
+    // Get the current screenshot dimensions
+    const viewBounds = this.map.bounds();
+    const srcWidth = viewBounds.right - viewBounds.left;
+    const srcHeight = viewBounds.bottom - viewBounds.top;
```
pchoisel, line 429: Minor optimization: I think this could be stored in the component data to avoid recalculating every time we move the mouse ?

pchoisel (commented):
LGTM

PR #863, opened 2024-12-25
Rgb support

arjunrajlab:
Added support for RGB images by parsing bands.

In the process, also refactored the generated multi-source JSON to be more explicit about sources for each frame. This was done to avoid a bug in the way large_image was previously parsing things in the non-composited case @manthey. Problem is solved for our scenario, so all good, but I can try and reproduce the bug if you are curious.

### DIFF
--- a/src/views/dataset/MultiSourceConfiguration.vue
+++ b/src/views/dataset/MultiSourceConfiguration.vue
@@ -7,7 +7,7 @@
       <v-divider class="my-2" />
       <v-data-table :headers="headers" :items="items" item-key="key" />
       <v-checkbox
-        v-if="isRGBFile && rgbBandCount > 1"
+        v-if="isMultiBandRGBFile"
         v-model="splitRGBBands"
         label="Split RGB files into separate channels (otherwise, only the red channel will be used)"
         class="mt-4"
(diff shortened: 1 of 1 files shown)


src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -696,6 +731,19 @@ export default class MultiSourceConfiguration extends Vue {
       channels = ["Default"];
     }
 
+    // Handle RGB expansion
+    if (this.isRGBFile && this.rgbBandCount > 1 && this.splitRGBBands) {
```
pchoisel, line 735: I see this `this.isRGBFile && this.rgbBandCount > 1` a lot  
It might be worth putting it into a helper function

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -937,6 +984,24 @@ export default class MultiSourceConfiguration extends Vue {
...
+
+  get submitError(): string | null {
+    if (!this.submitEnabled()) {
+      return "Not all variables are assigned";
+    }
+
+    if (!this.isRGBAssignmentValid) {
+      return "If splitting RGB file into channels, then filenames must be assigned to another variable";
+    }
+    return null;
```
pchoisel, line 996: Maybe return a `Unknown error` ?

arjunrajlab: Actually, if it gets this far, it means there is no error to display.

arjunrajlab:
@pchoisel This PR actually had a major bug in it that took some time to identify and fix (it is pure luck I didn't merge it already!). I _think_ I fixed it, but I think I might deploy in house for a while first and then merge if there are no complaints.

pchoisel:
LGTM, you can merge when you need to

PR #866, opened 2024-12-27
Adds progress bars to the tracking of annotation and property workers

arjunrajlab:
Adds progress bars in the progress manager. This allows for users to track the progress even if the UI is closed.


pchoisel:
This codes seems duplicated. Do you think you could factorize it into utility functions that both store reducers use ?

PR #867, opened 2024-12-30
Refactored the object panel UI

arjunrajlab:
Updated the UI for "Measure Objects":

1. Removed the separate Measure objects panel.
2. Moved it to a button on the "Properties" collapsible panel in the objects panel.
3. When you close the new "Measurements" dialog, it automatically opens the "Properties" collapsible panel, which makes it more discoverable.
4. Minor UI improvements.


pchoisel:
LGTM

PR #871, opened 2025-01-04
Updated scrolling behavior and viewer tour

arjunrajlab:
Set scrollto to false to avoid the interface shifting around unpredictably. Reduced timeout for the rotation to avoid user confusion. Swapped order of pan and zoom so that hopefully the user zooms first to make panning possible.


pchoisel:
LGTM

PR #876, opened 2025-01-26
Added a tour for adding a DAPI blob

arjunrajlab:
Added a tour for adding a DAPI blob tool.

Also, added automatic ID generation for toolsets and worker value lists (interface items from the worker).


pchoisel:
LGTM

PR #878, opened 2025-01-29
Add blob editing tool

arjunrajlab:
One common request from users is for a blob editing tool. This PR implements a blob editing tool that allows for blob editing. Rather than a brush and eraser ala Photoshop, it instead implements a "line splice" tool. You can just draw to add and remove from your blob.

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -97,6 +97,8 @@ import AnnotationActionPanel from "@/components/AnnotationActionPanel.vue";
 import TagSelectionDialog from "@/components/TagSelectionDialog.vue";
 import ColorSelectionDialog from "@/components/ColorSelectionDialog.vue";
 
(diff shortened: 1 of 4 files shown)


src/components/AnnotationViewer.vue
```diff
@@ -2159,6 +2159,245 @@ export default class AnnotationViewer extends Vue {
...
+    return {
+      ...annotation,
+      coordinates: newCoordinates,
+    };
+  }
+
+  // Helper function to check if two points are equal
+  pointsEqual(a: IGeoJSPosition, b: IGeoJSPosition): boolean {
+    return Math.abs(a.x - b.x) < 1e-10 && Math.abs(a.y - b.y) < 1e-10;
+  }
```
pchoisel, line 2399: I feel like these functions should go in a `utils` file, especially considering the size of this file

pchoisel:
LGTM otherwise

PR #880, opened 2025-01-30
Added channel checkbox group to worker interface

arjunrajlab:
Allows a checkbox group for channels as a worker interface element.

Can be added using e.g.:
```
'All channels': {
            'type': 'channelCheckboxes',
            'tooltip': 'Blur selected channels.',
            'displayOrder': 2,
        },
```

### DIFF
--- a/src/components/ChannelCheckboxGroup.vue
+++ b/src/components/ChannelCheckboxGroup.vue
@@ -1,17 +1,19 @@
 <template>
   <v-container class="pa-0">
     <v-subheader v-if="label" class="px-0">{{ label }}</v-subheader>
-    <v-row v-for="item in channelItems" :key="item.value" class="ma-0">
-      <v-col cols="12" class="pa-1">
-        <v-checkbox
-          v-model="selectedChannels[item.value]"
-          :label="item.text"
-          dense
-          hide-details
-          v-bind="$attrs"
-        ></v-checkbox>
-      </v-col>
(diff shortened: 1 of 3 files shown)


src/components/ChannelCheckboxGroup.vue
```diff
@@ -0,0 +1,60 @@
...
+
+  @Prop({ default: "" })
+  readonly label!: string;
+
+  @VModel({ type: Object })
+  selectedChannels!: { [key: number]: boolean };
+
+  get channelItems() {
+    const res = [];
+    if (this.store.dataset) {
```
pchoisel, line 34: Not important: instead of this, you can do an early return:
```ts
if (!this.store.dataset)
  return []
}
...
```

src/components/ChannelCheckboxGroup.vue
```diff
@@ -0,0 +1,60 @@
...
+      for (const channel of this.store.dataset.channels) {
+        res.push({
+          text:
+            this.store.dataset.channelNames.get(channel) ||
+            `Channel ${channel}`,
+          value: channel,
+        });
+      }
+    }
```
pchoisel, line 43: This should be:
```ts
this.store.dataset.channels.map((channel)=> {
          text:
            this.store.dataset.channelNames.get(channel) ||
            `Channel ${channel}`,
          value: channel,
        }
)
```

PR #888, opened 2025-02-05
Added scale bar color picker

arjunrajlab:


src/store/index.ts
```diff
@@ -359,6 +360,18 @@ export class Main extends VuexModule {
...
+  // Note that we are just updating a CSS variable here
+  // In principle, we could just have updated the CSS variable
+  // directly in the ScaleSettings.vue component because the CSS
+  // variable would persist accordingly. But if in future we want to save
+  // the interface state in some way, this will make it easier.
+  // Also enables multiple color pickers across the interface to synchronize.
+  @Mutation
+  public setScalebarColor(color: string) {
+    this.scalebarColor = color;
+    document.documentElement.style.setProperty("--scale-bar-color", color);
```
pchoisel, line 372: I'm not sure about this  
If we want to save the interface state, we can just export the part of the store that holds information about the interface, then we can restore it later.  
But modifying the style of the DOM here can be hard to trace later if we have a problem with this

arjunrajlab: So it seems that CSS selectors are the only way to change the color, currently:
https://github.com/OpenGeoscience/geojs/issues/1363
Is there a better pattern? I agree this is a bit clunky, but I didn't think it was worth the development effort to modify geojs.

pchoisel: Perfect, I would just remove
```js

  @Watch("scalebarColor")
  updateScalebarColor() {
    document.documentElement.style.setProperty(
      "--scale-bar-color",
      this.scalebarColor,
    );
  }
```
And replace it with `:style="{ '--scale-bar-color': scalebarColor }"` in

```js
  <div
    class="image"
    v-mousetrap="mousetrapAnnotations"
    :style="{ '--scale-bar-color': scalebarColor }"
  >
```
at the very top

That way, you can also remove
```css
:root {
  --scale-bar-color: #ffffff; /* Default white color */
}
```

This makes the information more localized instead of declaring a global variable

PR #890, opened 2025-02-07
Snapshot scalebar

arjunrajlab:
Adds a scalebar to snapshots, both for download and on the screen.

The only thing I couldn't quite figure out was how to put the text on the line. I couldn't figure out how to add the label to the geojs line annotation.

### DIFF
--- a/src/components/Snapshots.vue
+++ b/src/components/Snapshots.vue
@@ -537,6 +537,16 @@ export enum MovieFormat {
   WEBM = "webm",
 }
 
+export enum PixelSizeMode {
+  DATASET = "dataset",
+  MANUAL = "manual",
+}
+
+export enum ScalebarMode {
+  AUTOMATIC = "automatic",
+  MANUAL = "manual",
+}
+
 @Component({
   components: { TagPicker, MovieDialog },
 })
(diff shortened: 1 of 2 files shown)


src/components/Snapshots.vue
```diff
@@ -478,6 +618,16 @@ export default class Snapshots extends Vue {
...
 
+  addScalebar: boolean = true;
+  addScalebarText: boolean = true;
+  scalebarSettingsDialog: boolean = false;
+  snapshotScalebarColor: string = "#ffffff";
+  manualScalebarSettings: IScalebarSettings | null = null;
+  manualPixelSize: IScalebarSettings | null = null;
+
+  pixelSizeMode: "dataset" | "manual" = "dataset";
+  scalebarMode: "automatic" | "manual" = "automatic";
```
pchoisel, line 629: There should be enums for these strings, a typo happens so fast...

src/components/Snapshots.vue
```diff
@@ -575,6 +725,296 @@ export default class Snapshots extends Vue {
     return results;
   }
 
+  get manualScalebarSettingsLength(): number {
+    if (this.manualScalebarSettings) {
+      return this.manualScalebarSettings.length;
+    }
+    return 1.0;
```
pchoisel, line 732: This (and also the following getters) could be shortened with: `return this.manualScalebar?.length || 1.0`

pchoisel:
LGTM

PR #895, opened 2025-02-14
Show dataset owner

arjunrajlab:
Adds dataset owner to:
1. Recents menu
2. Breadcrumbs

### DIFF
--- a/src/layout/BreadCrumbs.vue
+++ b/src/layout/BreadCrumbs.vue
@@ -236,8 +236,12 @@ export default class BreadCrumbs extends Vue {
 
     const newItems: IBreadCrumbItem[] = [];
     const params: { [key: string]: string } = {};
-    if (datasetId) params.datasetId = datasetId;
-    if (configurationId) params.configurationId = configurationId;
+    if (datasetId) {
+      params.datasetId = datasetId;
+    }
+    if (configurationId) {
+      params.configurationId = configurationId;
+    }
 
     // Create dataset item
     let datasetItem: IBreadCrumbItem | undefined;
--- a/src/layout/BreadCrumbs.vue
+++ b/src/layout/BreadCrumbs.vue
@@ -43,11 +43,14 @@
           </template>
           <!-- Otherwise, simply make the item clickable -->
           <template v-else>
-            <router-link :to="item.to">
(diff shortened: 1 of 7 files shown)


src/layout/BreadCrumbs.vue
```diff
@@ -220,69 +234,96 @@ export default class BreadCrumbs extends Vue {
     this.previousRefreshInfo.configurationId = configurationId;
     this.previousRefreshInfo.routeName = this.$route.name;
 
-    this.items = [];
+    const newItems: IBreadCrumbItem[] = [];
+    const params: { [key: string]: string } = {};
+    if (datasetId) params.datasetId = datasetId;
+    if (configurationId) params.configurationId = configurationId;
```
pchoisel, line 240: It's always preferable to use the 
```
if(condition){
  action()
}
```
syntax for code readability

pchoisel (approved):
LGTM

It looks like this PR adds a lot of requests to the server, especially if the user has a lot of files  
All these requests, even small, have an overhead that can overload the server  
The best thing to do is to reorganize the code as possible to make batch requests instead of lots of small requests

arjunrajlab:
> LGTM
> 
> It looks like this PR adds a lot of requests to the server, especially if the user has a lot of files All these requests, even small, have an overhead that can overload the server The best thing to do is to reorganize the code as possible to make batch requests instead of lots of small requests

Yes, I was concerned about this… but I think it's not so bad? There appears to be a girderResources store that caches many server requests to minimize these sorts of things, and most of these "owner" requests should end up hitting that cache because it's the same owner for most files for most users. Is it okay?

pchoisel:
> > LGTM
> > It looks like this PR adds a lot of requests to the server, especially if the user has a lot of files All these requests, even small, have an overhead that can overload the server The best thing to do is to reorganize the code as possible to make batch requests instead of lots of small requests
> 
> Yes, I was concerned about this… but I think it's not so bad? There appears to be a girderResources store that caches many server requests to minimize these sorts of things, and most of these "owner" requests should end up hitting that cache because it's the same owner for most files for most users. Is it okay?

For the moment, it's fine, but when people start using the platform more, it will hinder the server's performances.

PR #896, opened 2025-02-15
Added heartbeat loop with 30 second interval

arjunrajlab:


arjunrajlab:
No longer needed because timeout issues are now handled via PR https://github.com/arjunrajlaboratory/NimbusImage/pull/900

There may still be some use for progress to be sent via a thread on long-running tasks that cannot be broken up, but that is probably best handled on a per-worker basis via threading the existing progress mechanisms.

PR #898, opened 2025-02-15
Optimize JSON export

arjunrajlab:
Chunked up the JSON export to avoid very large string variables.
Added progress indicators to give visual feedback.


pchoisel:
LGTM

PR #899, opened 2025-02-17
Fixed bugs in loading default worker menu items

arjunrajlab:
Added guard on tool being undefined.
Added guard in case tool.values.workerInterfaceValues is not defined.


pchoisel:
LGTM

PR #900, opened 2025-02-18
Add timeout to stream request

arjunrajlab:
Adding a 45 second timeout ensures that long-running tasks will remain connected despite proxy/load balancer timeouts.


pchoisel:
LGTM

PR #903, opened 2025-02-20
Switched from displaying login to email

arjunrajlab:
When showing user info, show the email instead of the Girder login.


pchoisel:
LGTM

PR #904, opened 2025-02-26
Prevent users from upload files outside of a folder

pchoisel:
Using Girder, users can only upload files in a folder. Uploading at the root of a collection, the user list or just the root of the filesystem breaks Girder.

### DIFF
--- a/src/views/dataset/NewDataset.vue
+++ b/src/views/dataset/NewDataset.vue
@@ -94,7 +94,7 @@
       </v-alert>
       <v-alert v-if="invalidLocation" text type="error">
         Cannot create datasets in this location. Please select a subfolder
-        within your user directory.
+        within your user directory or group folder.
       </v-alert>
 
       <div class="button-bar" v-if="!quickupload || pipelineError">


arjunrajlab (commented):
Thanks, we definitely need something like this! While this PR does cover the case of "Upload individual file", it still allows a dataset to be created in these root folders via the "Quick upload" and "Advanced upload".

pchoisel:
> Thanks, we definitely need something like this! While this PR does cover the case of "Upload individual file", it still allows a dataset to be created in these root folders via the "Quick upload" and "Advanced upload".

Normally I took care of that by disabling the `Select` button in `GirderLocationChooser` if the selected location is not a folder.

src/views/dataset/NewDataset.vue
```diff
@@ -83,11 +83,31 @@
...
+        >. This might happen, for instance, if a dataset by that name already
+        exists. Please update the dataset name field and try again.
+      </v-alert>
+      <v-alert v-if="fileSizeExceeded" text type="error">
+        Total file size ({{ totalSizeMB }} MB) exceeds the maximum allowed size
+        of {{ maxTotalFileSize }} MB
+      </v-alert>
+      <v-alert v-if="invalidLocation" text type="error">
+        Cannot create datasets in this location. Please select a subfolder
+        within your user directory.
```
pchoisel, line 97: `within your user directory or lab collection.` ?

pchoisel:
LGTM otherwise

PR #907, opened 2025-03-04
Added progress bar and indicators to annotation deletion

arjunrajlab:
Users have noticed that deleting large number of annotations takes a long time. We should of course fix that :), but for now, we should at least provide some progress indicator so that users know that a deletion is occurring.


luciemac:
LGTM

PR #910, opened 2025-03-06
Fix worker menu positioning

arjunrajlab:
Prevents the worker menu from going off screen slightly, which was very annoying. Calls updateDimensions after content is loaded.


luciemac (approved):

luciemac:
LGTM

PR #912, opened 2025-03-07
Add indices to make various searches more efficient

pchoisel:
Those indices are the one used in all the searches that I could find in the codebase.


arjunrajlab (approved):
Looks good to me!

PR #917, opened 2025-03-15
Job logs

arjunrajlab:
Adds the ability to see the logs of jobs, both of the currently running job and up to 20 previous jobs.

### DIFF
--- a/src/components/JobsLogs.vue
+++ b/src/components/JobsLogs.vue
@@ -152,6 +152,7 @@ import store from "@/store/index";
 import main from "@/store/index";
 import { formatDateString, formatDuration } from "@/utils/date";
 import { IJob } from "@/store/model";
+import { logError } from "@/utils/log";
 
 @Component
 export default class JobsLogs extends Vue {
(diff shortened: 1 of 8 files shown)


src/components/JobsLogs.vue
```diff
@@ -0,0 +1,397 @@
...
+        `Type: ${jobWithLog.type}\n\n` +
+        `Arguments:\n${jobWithLog.args.join("\n")}\n\n` +
+        `${
+          jobWithLog.status === 3
+            ? "Job completed successfully."
+            : jobWithLog.status === 4
+              ? "Job failed with errors."
+              : jobWithLog.status === 2
+                ? "Job is still running..."
+                : "Job status: " + this.getStatusText(jobWithLog.status)
```
luciemac, line 322: Seeing that, you probably should add a property to your Interface:
```
interface JobLogProperty {
  color: string;
  statusText: string;
  stateText ?: string;
}
```
then, update your object:
```
const jobLogStatus: {[key: number]: JobLogProperty} = {
  0: {color: "grey", statusText:  "Inactive"},
  1: {color: "blue", statusText:  "Queued"},
  2: {color: "orange", statusText:  "Running", stateText: "Job is still running..."},
  ....
}
```
and define a function:
```
function getJobState(status: number): string {
  if (status in jobLogStatus && jobLogStatus[status].stateText) {
    return  jobLogStatus[status].stateText
  }
  return "Job status: " + getStatusText(status)
}
```

and finally here, you should call `getJobState(jobWithLog.status)`

arjunrajlab: Thanks, done!

PR #922, opened 2025-03-26
Use more precise CSS selectors to change icon color

pchoisel:
This prevents Vuetify default theme to override it


arjunrajlab (approved):
Looks great!

PR #924, opened 2025-03-27
Handle OIB files

arjunrajlab:
Allows for the uploading of OIB files.
Does so by explicitly asking for a conversion to large_image. Not sure why that's required, but it doesn't work without it.

### DIFF
--- a/src/views/dataset/MultiSourceConfiguration.vue
+++ b/src/views/dataset/MultiSourceConfiguration.vue
@@ -569,8 +569,7 @@ export default class MultiSourceConfiguration extends Vue {
     if (hasOibFiles) {
       try {
         // Process each OIB file to ensure it has a large image
-        for (let i = 0; i < items.length; i++) {
-          const item = items[i];
+        for (const item of items) {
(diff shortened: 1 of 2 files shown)


src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -555,19 +557,69 @@ export default class MultiSourceConfiguration extends Vue {
...
+      return parts.length > 1 ? parts[parts.length - 1].toLowerCase() : "";
+    });
+    const hasOibFiles = fileExtensions.includes("oib");
+
+    // For OIB files, explicitly try to create a large image first
+    if (hasOibFiles) {
+      try {
+        // Process each OIB file to ensure it has a large image
+        for (let i = 0; i < items.length; i++) {
+          const item = items[i];
```
luciemac, line 573: you probably can do `for (const item of items) { ...}` then you can remove `const item = items[i];`

arjunrajlab: Thanks!

luciemac (approved):

PR #959, opened 2025-06-09
Allow for "any" as a property shape

arjunrajlab:
When finding the list of properties, include images labeled with annotationShape="any".


arjunrajlab:
This closes https://github.com/arjunrajlaboratory/NimbusImage/issues/551

luciemac (approved):

PR #986, opened 2025-10-13
Add hints to avoid query planner

arjunrajlab:
Query planner takes a long time for big queries, presumably because it tries multiple different queries. By adding a hint, that time is reduced significantly.


pchoisel:
LGTM

PR #988, opened 2025-10-15
Rename collections

arjunrajlab:
Here, we add the ability to rename collections, which got removed in a prior update. We create an endpoint for the renaming and then use that to update the name from the ConfigurationInfo.vue component.

### DIFF
--- a/src/store/girderResources.ts
+++ b/src/store/girderResources.ts
@@ -74,46 +74,21 @@ export class GirderResources extends VuexModule {
     id: string;
     type: IGirderSelectAble["_modelType"];
   }) {
-    // Helper: treat upenn_collection <-> item as equivalent on the wire
-    const typesEquivalent = (a?: string | null, b?: string | null) => {
-      if (!a || !b) return false;
-      if (a === b) return true;
-      // If backend sometimes returns 'item' for collections, accept it
-      if (
-        (a === "item" && b === "upenn_collection") ||
-        (a === "upenn_collection" && b === "item")
-      ) {
-        return true;
-      }
-      return false;
-    };
-
     if (id in this.resourcesLocks) {
       await this.resourcesLocks[id];
     }
 
     const cached = this.resources[id];
 
-    // Not cached OR cached with the wrong/ambiguous type → re-fetch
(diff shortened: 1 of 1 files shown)


devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/collection.py
```diff
@@ -169,6 +169,19 @@ def setMetadata(self, collection, metadata, allowNull=False):
...
+        if name is not None:
+            collection['name'] = Item()._validateString(name)
+            collection['lowerName'] = collection['name'].lower()
+            changed = True
+        if description is not None:
+            collection['description'] = Item()._validateString(description)
+            changed = True
+        if changed:
+            collection['updated'] = datetime.datetime.utcnow()
+        return self.save(collection)
```
pchoisel:
```suggestion
        if not changed:
            return collection

        collection['updated'] = datetime.datetime.utcnow()
        return self.save(collection)
```

src/layout/BreadCrumbs.vue
```diff
@@ -331,6 +339,46 @@ export default class BreadCrumbs extends Vue {
...
+    // Trigger fetch when undefined OR null
+    if (resource == null && this.currentDatasetId) {
+      this.girderResources.forceFetchResource({
+        id: this.currentDatasetId,
+        type: "folder",
+      });
+      return;
+    }
+
+    if (resource && resource.name) {
```
pchoisel: You can use the `?.` operator to access the property of a potentially undefined value

src/store/girderResources.ts
```diff
@@ -60,7 +60,12 @@ export class GirderResources extends VuexModule {
   }) {
     try {
       const resource = await this.requestResource({ id, type });
-      this.setResource({ id, resource });
+      // Normalize the type in cache to the logical type requested
+      const normalized =
+        resource && resource._modelType !== type
+          ? ({ ...resource, _modelType: type } as IGirderSelectAble)
+          : resource;
+      this.setResource({ id, resource: normalized });
```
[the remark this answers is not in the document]
arjunrajlab: No, it was from some previous debugging attempt. I have removed it now, thanks for spotting it!

PR #994, opened 2025-10-26
Add labels to dataset metadata

arjunrajlab:
This PR adds the ability to add more useful labels to the sliders from either the parsed filenames or from the file metadata itself (ND2 only).

- It takes the filename information and the internal_metadata information and turns that into axis labels. This metadata is added to the dataset/folder metadata in Girder.
- In principle, this information would be more naturally stored in the large_image axis labels; however, given the complexity of our large_image JSON, I wasn't able to figure this out, so I just stored it on our own. This might be something worth revisiting in the future.
- The UI for showing the sliders was updated to accommodate the sometimes long axis value strings. It might be possible to make this slightly more compact.
- The UI now only shows sliders for the dimensions that have more than one value to help save space.
- In the Action panel, there is now an option to download CSVs for the conversion between the indices and the label names.

### DIFF
--- a/src/components/AnnotationBrowser/IndexConversionDialog.vue
+++ b/src/components/AnnotationBrowser/IndexConversionDialog.vue
@@ -122,9 +122,7 @@ import Papa from "papaparse";
 import { downloadToClient } from "@/utils/download";
 import { logError } from "@/utils/log";
 
-@Component({
-  components: {},
-})
+@Component({})
 export default class IndexConversionDialog extends Vue {
   readonly store = store;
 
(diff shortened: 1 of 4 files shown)


src/components/AnnotationBrowser/IndexConversionDialog.vue
```diff
@@ -0,0 +1,269 @@
...
+                    No labels detected; index only
+                  </div>
+                </div>
+                <v-btn color="primary" @click="downloadTime">
+                  <v-icon left>mdi-download</v-icon>
+                  Download Time CSV
+                </v-btn>
+              </div>
+            </v-card-text>
+          </v-card>
```
pchoisel: Please factorize these 3 sections into a new component

src/components/AnnotationBrowser/IndexConversionDialog.vue
```diff
@@ -0,0 +1,269 @@
...
+      this.timeCount,
+      this.dimensionLabels?.t ?? null,
+    );
+    const datasetName = this.dataset?.name ?? "dataset";
+    const params = {
+      href: "data:text/csv;charset=utf-8," + encodeURIComponent(csv),
+      download: `${datasetName}_time_index_conversion.csv`,
+    };
+    downloadToClient(params);
+  }
```
pchoisel: Please factorize these functions

src/components/ValueSlider.vue
```diff
@@ -90,6 +102,14 @@ export default class ValueSlider extends Vue {
     this.$emit("input", numberValue - this.offset);
   }
 
+  get displayValue() {
+    return this.valueLabel || (this.value + this.offset).toString();
+  }
+
+  set displayValue(_value: string) {
+    // This will be handled by handleInput method
+  }
```
pchoisel: Is this set necessary ?

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -260,6 +260,122 @@ enum Sources {
...
+
+  const points = xy.parameters.points;
+  if (!points || points.length === 0) return null;
+
+  return points.map((point) => {
+    const [x, y] = point.stagePositionUm;
+    return `${trimFloat(x)}, ${trimFloat(y)}`;
+  });
+}
+
```
pchoisel: These should go in a separate file (like `ND2Helpers.ts`)  
MultiSourceConfiguration.vue is already 1600 lines long ^^'

src/views/dataset/MultiSourceConfiguration.vue
```diff
@@ -1352,17 +1473,44 @@ export default class MultiSourceConfiguration extends Vue {
...
+      this.store.scheduleMaxMergeCache(datasetId);
+
+      // Add a small delay before scheduling histogram cache to ensure the item is processed
+      setTimeout(async () => {
+        try {
+          await this.store.scheduleHistogramCache(datasetId);
+        } catch (error) {
+          logError("Failed to schedule histogram cache:", error);
+        }
+      }, 2000);
```
pchoisel: That will not play well with the cloud deployment, you cannot expect a time limit to always work  
If we always schedule the histogram cache after the item is processed, maybe the backend could be responsible for scheduling it when the item is done processing ?

pchoisel (commented):
I'm fine with the modifications, however there should be more separation between the processing of regular files and the processing of ND2 files.  
Do you think other types of files will get the same treatment as the ND2 files ?

src/components/AnnotationBrowser/IndexConversionDialog.vue
arjunrajlab: I could… but it seems like a lot of overhead and obfuscation to make a new component for just this little bit of repetition, no?

src/components/AnnotationBrowser/IndexConversionDialog.vue
arjunrajlab: Done!

src/components/AnnotationBrowser/IndexConversionDialog.vue
pchoisel: Sounds good

pchoisel:
LGTM

PR #1000, opened 2025-11-19
Improve the upload dialog

arjunrajlab:
Change the name to Create Dataset.
Add name picker and location picker.
Change the button names and assign primary and secondary. Include name exist checking feature.


pchoisel (commented):
Congrats for PR #1000 !

PR #1001, opened 2025-11-19
Allow for current folder in ImportConfiguration and also allow locati…

arjunrajlab:
…on choosing

Before, when you "Add to existing collection" from the dataset view, you would default to the private folder. This update:
1. On the above flow, uses the folder of the dataset, which is more natural.
2. Allows for changing the folder.

### DIFF
--- a/src/views/configuration/ImportConfiguration.vue
+++ b/src/views/configuration/ImportConfiguration.vue
@@ -56,7 +56,7 @@ export default class ImportConfiguration extends Vue {
     if (Array.isArray(folderId)) {
       return folderId[0] || undefined;
     }
-    if (folderId === null || folderId === undefined) {
+    if (folderId === null) {
       return undefined;
     }
     return folderId;


src/views/configuration/ImportConfiguration.vue
```diff
@@ -1,28 +1,108 @@
...
+  get folderId(): string | undefined {
+    if (this.selectedFolder && this.selectedFolder._modelType === "folder") {
+      return this.selectedFolder._id;
+    }
+    // Fallback to route query if folder not yet loaded
+    const folderId = this.$route.query.folderId;
+    if (Array.isArray(folderId)) {
+      return folderId[0] || undefined;
+    }
+    if (folderId === null || folderId === undefined) {
```
pchoisel: This can be replaced with `if(folderId ==null)`

arjunrajlab: Good point, thanks!

PR #1005, opened 2025-11-24
Public access

arjunrajlab:
Allows for public access of a datasetView even if the user is not logged in.

1. Backend updated to allow public access using a public endpoint.
2. Frontend guards in place to prevent calling API functions that are forbidden as anonymous.
3. UI added to Sharing component for making public.

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/property.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/property.py
@@ -121,8 +121,9 @@ class AnnotationProperty(Resource):
         .errorResponse()
     )
     def getAllProperties(self, params):
-        # Note that this function is analogous to the "find" function in other classes,
-        # but here we don't have any filtering criteria, so we can just get all properties.
+        # Note that this function is analogous to the "find" function in other
+        # classes, but here we don't have any filtering criteria, so we can
+        # just get all properties.
 
         limit, offset, sort = self.getPagingParameters(params, "lowerName")
         user = self.getCurrentUser()
@@ -162,8 +163,10 @@ class AnnotationProperty(Resource):
         )
     )
     def get(self, id, params):
-        # Note that params is not used in this method, but it is required by the describeRoute decorator.
-        # This is because the describeRoute decorator expects a params argument, but we don't need it in this method.
+        # Note that params is not used in this method, but it is required by
+        # the describeRoute decorator. This is because the describeRoute
+        # decorator expects a params argument, but we don't need it in this
+        # method.
 
         user = self.getCurrentUser()
 


devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/property.py
```diff
@@ -119,28 +121,71 @@ def update(self, property, params):
...
+            if 'meta' in config and 'propertyIds' in config['meta']:
+                for pid in config['meta']['propertyIds']:
+                    accessible_property_ids.add(ObjectId(pid))
+
+        # Query properties
+        if accessible_property_ids:
+            query = {'_id': {'$in': list(accessible_property_ids)}}
+        else:
+            # User has no accessible configurations, return empty
+            query = {'_id': {'$in': []}}
```
pchoisel: Those two paths do exactly the same thing ?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/datasetView.py
```diff
@@ -57,7 +57,7 @@ class DatasetViewSchema:
     }
 
 
-class DatasetView(ProxiedModel, AccessControlledModel):
+class DatasetView(AccessControlledModel, ProxiedModel):
```
pchoisel: So here you are transforming the DatasetView, AnnotationProperty, ... models in AccessControlledModel instead of ProxiedModel.  
The difference is that now each document will handle its access control individually instead of relying on the parent document.  
I don't really see why you want to do that ?  
Instead, if you set the dataset as public, the underlying documents should become public as well

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/property.py
```diff
@@ -119,28 +121,71 @@ def update(self, property, params):
...
+            user=user,
+            level=AccessType.READ
+        ))
+
+        # Collect all property IDs from accessible configurations
+        accessible_property_ids = set()
+        for config in accessible_configs:
+            if 'meta' in config and 'propertyIds' in config['meta']:
+                for pid in config['meta']['propertyIds']:
+                    accessible_property_ids.add(ObjectId(pid))
```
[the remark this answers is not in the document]
arjunrajlab: I think it actually has to do the conversion. A property _id is indeed an ObjectId, but here, we are accessing the metadata from the configuration, which stores it as a plain string. I added a small comment about it.

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/property.py
```diff
@@ -119,28 +122,70 @@ def update(self, property, params):
     )
     def find(self, params):
```
pchoisel: This function does not use the `params` anymore, so its name is confusing regarding the find functions of the other models.
Why not renaming it to something like `getAllProperties` ?

src/components/ShareDataset.vue
```diff
@@ -110,13 +133,25 @@ export default class ShareDataset extends Vue {
   userErrorString: string = "";
   showUserError: boolean = false;
   accessLevel: number = -1;
+  makePublic: boolean = false;
```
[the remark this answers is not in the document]
pchoisel: Oh wait no, that's the state of the public/shared switch.  
So what about `usePublicAccess` ? Or `isPublicSelected` ?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/property.py
arjunrajlab: Done!

PR #1010, opened 2026-01-07
Add streaming JSON export endpoint

arjunrajlab:
## Summary
- Adds a new backend `/api/v1/export/json` endpoint that streams annotation data directly to the browser
- Updates frontend to use the new endpoint instead of assembling JSON in memory
- Eliminates memory issues when exporting large datasets

### Backend
- New `export.py` API resource with streaming JSON generation
- Finds properties via DatasetViews (correct data model relationship)
- Uses `orjson` for efficient serialization
- Supports filtering: annotations, connections, properties, property values
- Comprehensive test coverage (9 tests)

### Frontend
- New `ExportAPI.ts` class for direct browser downloads
- Simplified `AnnotationExport.vue` - delegates to backend streaming
- No JavaScript memory usage for large exports

## Test plan
- [x] Backend tests pass (`tox` in AnnotationPlugin)
- [x] Manual test: export downloads correct JSON file
- [x] Verified exported data matches previous frontend export format

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/test/test_export.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/test/test_export.py
@@ -1,5 +1,4 @@
 import pytest
-import json
 
 from upenncontrast_annotation.server.models.annotation import Annotation
 from upenncontrast_annotation.server.models.connections import (


devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/export.py
```diff
@@ -0,0 +1,303 @@
...
+        includeConnections,
+        includeProperties,
+        includePropertyValues
+    ):
+        """
+        Generator function that yields JSON chunks.
+
+        This is returned as a callable to match Girder's streaming pattern.
+        """
+        def generate():
```
pchoisel: I see what you want to do here, but this kind of mechanism is setup for transferring large amounts of data, not something you would typically do for a JSON file.  
What's the size of the file you would expect ?  
If you really want to have this streaming behavior, you can keep the way the `generate` function works, but you'll have to factorize most of it since most of the code of this file is repeating.  
Unless I'm missing something, the entire `_streamCursorAsArray` function can be replaced with a `orjson.dumps()`

PR #1014, opened 2026-01-13
Fix the need to double click on the login button

pchoisel:
This commit retains the fact that logging out does not display the login screen.

### DIFF
--- a/src/layout/UserMenu.vue
+++ b/src/layout/UserMenu.vue
@@ -77,14 +77,15 @@ export default class UserMenu extends Vue {
   readonly store = store;
 
   @Watch("store.isLoggedIn")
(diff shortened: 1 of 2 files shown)


arjunrajlab (commented):
This solves the problem of the double click on login, but does so by reverting functionality back to before the public sharing dataset. The problem is illustrated by the following screenshot (in anonymous mode):

Image

When viewing anonymously, we don't want to show the login window over the screen. It can be dismissed by clicking away, but that behavior is a bit odd. I think it's better to just not show the login screen unless we are on the home route. But in doing so, I suspect that we will again have this problem reappear—hence the other solution, which allows for the flow to proceed without having the login component mounted. But I could be wrong about that.

PR #1016, opened 2026-01-15
Redesign Share Dataset dialog with real-time access management

arjunrajlab:
## Summary
- Add new `getDatasetAccess` endpoint to fetch current access list for a dataset
- Extend `share` endpoint to support revoking access (accessType: null)
- Redesign ShareDataset.vue with real-time updates - each action immediately syncs to backend
- Show current users with access, their permission levels, and ability to modify/remove
- Add confirmation dialog before removing user access
- Display user emails alongside usernames in access list

## Test plan
- [x] Open Share dialog on a dataset you own → verify current access list loads
- [x] Add a user by email/username → verify they appear in list immediately
- [x] Change a user's access level (Read/Write) → verify change persists
- [x] Remove a user → verify confirmation dialog appears, then user is removed
- [x] Toggle public checkbox → verify public status changes
- [x] Try to remove dataset owner → verify button is disabled (lock icon)
- [x] Open dataset as different user → verify access changes took effect
- [x] Test with invalid email/username → verify error message appears

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/test/test_sharing.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/test/test_sharing.py
@@ -9,7 +9,6 @@ import random
 from girder.constants import AccessType
 from girder.exceptions import AccessException
 from girder.models.folder import Folder
-from girder.models.user import User
 
 from upenncontrast_annotation.server.models.datasetView import (
     DatasetView as DatasetViewModel


devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/datasetView.py
```diff
@@ -238,8 +245,15 @@ def loadDocument(model, documentId):
...
-        accessType = AccessType().validate(body["accessType"])
+
+        # accessType can be null (to remove access) or a valid access level
+        rawAccessType = body["accessType"]
+        if rawAccessType is None:
+            # None means remove access - Girder's setUserAccess handles this
+            accessType = None
+        else:
+            # Will raise if accessType is a bad value
+            accessType = AccessType().validate(rawAccessType)
```
pchoisel: This would be simplified if you used -1 for no access

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/datasetView.py
```diff
@@ -306,20 +318,106 @@ def setDatasetPublic(self, dataset, public):
...
             self._datasetViewModel.setPublic(dv, public, save=True)
 
-        # 5. Set permissions on Configurations
-        for configId in configIds:
-            config = CollectionModel().load(configId, force=True)
-            if config:
-                CollectionModel().setPublic(config, public, save=True)
+        # 5. Bulk load configurations, then set permissions
+        # (Bulk load avoids N+1 queries; setPublic must still be individual)
+        configs = list(CollectionModel().collection.find({
```
pchoisel: I would recommend against using the find method from the PyMongo collection object as you are bypassing Girder's Model find method, which adds a few security items, such as making sure the query only relies on authorized fields and setting a timeout.  
Using `CollectionModel().find` instead of `CollectionModel().collection.find` does not prevent you from doing a bulk search.

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/datasetView.py
```diff
@@ -306,20 +318,106 @@ def setDatasetPublic(self, dataset, public):
...
+
+        # Get unique configuration IDs
+        configIds = {dv['configurationId'] for dv in datasetViews}
+
+        # Bulk load all configurations at once (avoid N+1 queries)
+        # TODO: The UI currently doesn't expose per-configuration public status
+        # because configurations are confusing for most users. Most datasets
+        # have just one configuration anyway. Consider adding this UI later
+        # if users request more granular public/private control.
+        configs = list(CollectionModel().collection.find({
```
pchoisel: No need to use the Pymongo collection

PR #1127, opened 2026-05-01
Fix @loadmodel for non-admins on Annotation/Connection (MRO regression from #1005)

arjunrajlab:
## Summary
- Flip base order on `Annotation`, `AnnotationConnection`, and `AnnotationPropertyValues` from `(ProxiedModel, AccessControlMixin)` to `(AccessControlMixin, ProxiedModel)` so the mixin's `load`/`hasAccess`/`findWithPermissions` (which proxy to the parent dataset folder) win the MRO over `AccessControlledModel`'s versions (which look for an `access`/`public` field these docs don't have).
- Adds 5 regression tests in `test_access_control.py::TestInheritedAccessLoad` covering direct `model.load(level=..., user=user)` on all three models plus end-to-end `GET /upenn_annotation/{id}` via `@loadmodel` for both shared-access and denied cases.
- Harden `PUT /upenn_annotation/multiple` so bulk annotation updates keep the existing empty/`null` response, but raise 403 if any requested annotation is not actually writable/found instead of returning `200/null` after updating zero documents.
- Adds 3 bulk-update regression tests in `test_annotations.py::TestUpdateMultiple` covering non-admin dataset-owner geometry persistence, denied updates that must not silently skip, and inherited dataset access via `findWithPermissions`.
- Ensure `recordable` stops model recording before re-raising endpoint exceptions, so failed writes do not leave stale recording state behind.

## Background
PR #1005 (Dec 2025) changed `CustomNimbusImageModel` from `Model` to `AccessControlledModel`. That pushed `AccessControlledModel.hasAccess` ahead of `AccessControlMixin.hasAccess` in the MRO of these three models, so for non-admin users `@loadmodel` always returned 403. Admins were unaffected via the `if user['admin']: return True` short-circuit, which is why this didn't surface immediately.

The same MRO regression also affected `findWithPermissions` on annotation-like models. In `PUT /upenn_annotation/multiple`, that meant non-admin dataset owners could send a valid geometry update, get `200 OK`/`null`, create an empty history entry, and still update zero annotation documents. Refreshing then restored the original coordinates.

## Scope
- **Annotation / Connection** are actively affected: master uses `@loadmodel` on their `get`/`update`/`delete` endpoints, and `findWithPermissions` is called in `updateMultiple` / `deleteMultiple` paths.
- **Bulk annotation update** now fails loudly if the permission-filtered query does not return every requested annotation, while avoiding a large response payload for successful bulk edits.
- **PropertyValues** gets the same MRO flip defensively — its endpoints all check access via `Folder().load()` and use raw `.find()`/`.save()`/`.appendValues()`, so no user-visible bug existed there. Flipping it now keeps the three sibling models consistent and prevents a regression if anyone later adds `@loadmodel` or uses `findWithPermissions` on it.

Other methods (`save`, `validate`, etc.) aren't on `AccessControlMixin` and still resolve through `ProxiedModel` as before.

## Test plan
- [x] `tox` passes locally (161 existing + 5 new = 166 tests) for the original MRO fix
- [x] New inherited-access regression tests verified to fail without the MRO flip (confirmed by temporarily reverting on `Annotation`)
- [x] `flake8 devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation` clean for the original MRO fix
- [x] `python3 -m py_compile` on updated backend files and `test_annotations.py`
- [x] `git diff --check`
- [x] Container dry-run verified patched `updateMultiple` finds the live reported annotation as writable without mutating Mongo
- [x] Manual curl against rebuilt container covering:
  - Anonymous on private dataset → 401 on get/find/count/connection get
  - Anonymous on public dataset → 200 on get/find/count for both annotation and connection (and on propertyValues find/count/histogram)
  - Non-admin without access → 403 on get/update/delete and find for annotation/connection/propertyValues
  - Non-admin with READ → 200 on get/find/count, 403 on PUT/POST/DELETE
  - Non-admin with WRITE → 200 on PUT/DELETE through `@loadmodel`

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
@@ -283,7 +283,7 @@ class Annotation(Resource):
                 newDatasetIds, self.getCurrentUser()
             )
 
-        return self._annotationModel.updateMultiple(
+        self._annotationModel.updateMultiple(
             annotationIdToUpdate, self.getCurrentUser()
         )
 
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
@@ -283,7 +283,7 @@ class Annotation(Resource):
                 newDatasetIds, self.getCurrentUser()
             )
 
-        self._annotationModel.updateMultiple(
+        return self._annotationModel.updateMultiple(
             annotationIdToUpdate, self.getCurrentUser()
         )
 
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/proxiedModel.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/proxiedModel.py
@@ -54,14 +54,16 @@ class recordable:
             if datasetId is None:
                 return fun(*args, **kwargs)
 
-            # Wrap original endpoint between a start and a stop recording
+            # Wrap original endpoint between a start and a stop recording.
+            # The finally ensures stopRecording fires even on failure so the
+            # model's recording state never leaks across requests.
             events.trigger("proxiedModel.startRecording")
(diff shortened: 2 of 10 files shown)


pchoisel:
LGTM

Issue #1136, opened 2026-05-05
Per-annotation structuredClone in updateMultipleAnnotations is slow on large updates


arjunrajlab:
## Issue

\`src/store/annotation.ts:875\` uses \`structuredClone(annotation)\` inside a per-annotation loop in the multi-update path. For each annotation being updated, the entire annotation object is deep-cloned — including \`coordinates\` arrays which can be large for polygon/contour annotations.

When the annotation is a Vue reactive Proxy (which it generally is — the annotations live in the reactive \`annotation\` store module), \`structuredClone\` reads through every Proxy.get trap on every leaf, so the cost compounds with reactivity overhead.

## Why it matters

Surfaced in an external diagnostic of NimbusImage's responsiveness regression after the Vue 2 → Vue 3 migration. Multi-update was called out as one of the hot-path culprits when batch-updating annotations (e.g., bulk tag changes, color updates).

## Suggested investigation / fix

Options:
1. \`structuredClone(toRaw(annotation))\` — bypasses the Proxy walk.
2. Replace the deep clone with a targeted shallow copy of just the fields the update path actually needs to mutate.
3. \`markRaw\` annotations stored in the annotation store (large structural change, would need careful auditing of who reads which fields reactively).

(1) is the cheapest and lowest-risk first step; profile to see if it's enough.

## Severity

Medium — perf-only, no correctness issue. Worth doing before any UI work that triggers large bulk updates.

## References

This was item flagged in the perf diagnostic that fed into branch perf/markraw-and-watcher-fixes (see code review on that branch).

Issue #1135, opened 2026-05-05
Remove stale histogramsLoaded fields on AnnotationsAPI and PropertiesAPI


arjunrajlab:
## Issue

`AnnotationsAPI` and `PropertiesAPI` both declare a public `histogramsLoaded = 0` field that nothing reads from outside the class:

- \`src/store/AnnotationsAPI.ts:26\`: \`histogramsLoaded = 0;\`
- \`src/store/PropertiesAPI.ts:32\`: \`histogramsLoaded = 0;\`

These appear to be copy-paste residue from \`GirderAPI\`, where \`histogramsLoaded\` is meaningful (it's read reactively by the \`layerStackImages\` getter to gate tile URL emission).

## Why it matters

- They look like reactive-read targets but aren't. While doing the perf branch (perf/markraw-and-watcher-fixes), we considered \`markRaw\`'ing these API instances for performance. We had to leave them reactive partly because static grep can't reliably find every reactive read, and these fields look like potential consumers — but they're not consumed anywhere.
- Removing them would unblock a future \`markRaw\` of \`AnnotationsAPI\` / \`PropertiesAPI\` to reduce per-method-call Proxy overhead.

## Suggested fix

Delete the two declarations and any (presumably no-op) writes to them inside their respective files. Verify with a grep that no consumer exists outside the class.

## Severity

Low — dead code cleanup.

Issue #1134, opened 2026-05-05
ImageViewer mutates Vuex maps state in place (maps.value.pop / maps.value = [])


arjunrajlab:
## Issue

`src/components/ImageViewer.vue` mutates the Vuex-managed `store.maps` array directly, outside any mutation:

- `maps.value.pop()` at `src/components/ImageViewer.vue:1184` (inside `draw()` when shrinking the maps array to match `mapLayerList.length`).
- `maps.value = []` at `src/components/ImageViewer.vue:1432-1435` (inside `onBeforeUnmount`, after calling `mapentry.map.exit()` on each entry).

This works only because Vuex strict mode is not enabled. Strict mode would throw on these. They violate the Vuex contract that state can only be mutated through `@Mutation` handlers.

## Why it matters

- Code review confusion: it's not obvious from the call site that `maps.value` is a reactive Vuex-state proxy, so the in-place pop reads as ordinary local state mutation.
- Surfaced as a side observation while doing the perf branch (perf/markraw-and-watcher-fixes) — the IMapEntry markRaw work in `setMaps` interacts with these in-place mutations, and we want them to be principled.
- If we ever turn on strict mode, this breaks.

## Suggested fix

Add explicit mutations on the main store:
- `popMap()` (or `removeMapAt(idx)`) for the shrink loop.
- `clearMaps()` for unmount.

Then ImageViewer calls `store.popMap()` / `store.clearMaps()` instead of touching the array directly.

## Severity

Low — pre-existing, not actively broken. Cleanup, not bug-fix.

PR #1137, opened 2026-05-05
Reduce Vue 3 reactive Proxy overhead on canvas + store hot paths

arjunrajlab:
## Summary

Targets the structural reactivity overhead from vuex-module-decorators wrapping every state field in a deep Proxy, plus a deep watcher and a GeoJS handler-accumulation issue surfaced in an external perf diagnostic.

- **`markRaw` the heavy GirderAPI caches** (`imageCache`, `histogramCache`, `resolvedHistogramCache`) at their declaration so Vue's reactive Map interceptor doesn't track every `.get` / `.set` on the per-tile-fetch hot path. The API instances themselves stay reactive (`histogramsLoaded` gates the `layerStackImages` getter — the v1 attempt to markRaw the whole instance broke initial render).
- **`markRaw` each `IMapEntry` in `setMaps`** so the canvas hot path (every read of `mapentry.map` / `.annotationLayer`) doesn't go through a Proxy. The outer maps array stays reactive so push/pop still trigger watchers.
- **`markRaw` `propertyValues` at declaration** in the properties store — largest mutable structure in the app, was getting proxy-walked at module init before the first mutation reassigned it.
- **Replace `{ deep: true }` watcher on `capturedMouseState`** with two cheap watchers (identity transitions + `path.length` growth). The deep version recursively dirty-checked the whole `IMouseState` graph (which includes the entire `IMapEntry` → GeoJS map) on every mouse event.
- **`shallowRef` the eight `IGeoJSAnnotation` refs in `AnnotationViewer`** plus `samMapEntry` in `ImageViewer`. They hold heavy native GeoJS objects whose internals must not be Proxy-wrapped.
- **`geoOff` cleanup before `geoOn` rebinds** in `bindAnnotationEvents` / `bindInteractionEvents` / `bindTimelapseEvents`, so handlers don't accumulate across layer-prop rebuilds (e.g., dataset reset). Extracted the inline mouseclick handler so it has a stable reference for `geoOff`.
- Comments on the trade-offs flagged during code review (asymmetric bind/unbind, two-watcher swap double-fire, `setMaps` mutating its input array elements).

## Out of scope (filed as follow-ups)

- #1134 — `maps.value.pop()` / `maps.value = []` mutating Vuex state in place.
- #1135 — Stale `histogramsLoaded` fields on `AnnotationsAPI` / `PropertiesAPI`.
- #1136 — Per-annotation `structuredClone` in multi-update is slow.

Per-the-original-diagnostic, intentionally **not addressed here**: ONNX runtime CPU fallback, plain `v-data-table` → `v-data-table-virtual`.

## Test plan

- [x] `pnpm tsc` clean
- [x] `pnpm lint:ci` clean (zero warnings)
- [x] `pnpm test` — 2118/2118 pass (1 pre-existing unrelated rejection in `NewDataset.test.ts`, also present on master)
- [x] Manual: image renders on initial page load on `/datasetView/:id/view`
- [x] Manual: time-slider scrub works
- [ ] Manual: 20-minute session to confirm the GeoJS handler-accumulation fix actually relieves the long-session sluggishness reported in the diagnostic
- [ ] Manual: dataset switch (triggers layer-prop watchers and the new unbind-old / bind-new path)

### DIFF
--- a/src/components/AnnotationViewer.vue
+++ b/src/components/AnnotationViewer.vue
@@ -209,14 +209,17 @@ const props = withDefaults(
 const isDragging = ref(false);
 const dragStartPosition = ref<IGeoJSPosition | null>(null);
 const draggedAnnotation = ref<IAnnotation | null>(null);
-const dragGhostAnnotation = ref<IGeoJSAnnotation | null>(null);
-const dragOriginalCoordinates = ref<IGeoJSPosition[] | null>(null);
-const pendingAnnotation = ref<IGeoJSAnnotation | null>(null);
-const selectionAnnotation = ref<IGeoJSAnnotation | null>(null);
-const samPromptAnnotations = ref<IGeoJSAnnotation[]>([]);
-const samUnsubmittedAnnotation = ref<IGeoJSAnnotation | null>(null);
-const samLivePreviewAnnotation = ref<IGeoJSAnnotation | null>(null);
-const cursorAnnotation = ref<IGeoJSAnnotation | null>(null);
+// IGeoJSAnnotation instances are heavy native objects whose internals must
+// not be Proxy-wrapped. shallowRef tracks identity (so swap/clear triggers
+// reactivity) but skips deep-walking the object graph.
+const dragGhostAnnotation = shallowRef<IGeoJSAnnotation | null>(null);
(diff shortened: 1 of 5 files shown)


arjunrajlab:
@codex review

PR #1140, opened 2026-05-07
Fix unselectable checkboxes; polish JSON import dialog

arjunrajlab:
## Summary
- Fix tool-configuration checkboxes that appeared visually selectable but never toggled. The template-level `meta.value` (a default-value seed consumed by `setDefaultValues`) was falling through onto the rendered component; on `VCheckbox`/`VSelectionControl` a `value` prop redefines `trueValue`/`falseValue` and breaks the checked toggle. `ToolConfigurationItem.vue` now strips `value` out of `meta` before binding.
- Clean up the "Import from JSON" dialog: replace the floating `mdi-code-json` glyph + generic "JSON file" label with an outlined file input, an `mdi-file-upload-outline` icon inside the field, and a clearer "Click to upload a JSON file" label (with file-size readout).
- After a successful import, the activator button briefly turns green with a check icon so the action has visible feedback before the dialog closes; the timer is cleared on unmount.

## Test plan
- [ ] Open a tool whose template defines a checkbox interface element with a default `meta.value` (e.g., a worker tool with a boolean option). Verify the checkbox toggles when clicked.
- [ ] Open the annotation browser → Actions → Import from JSON. Confirm the dialog shows an outlined input with an upload icon inside, no `{...}` glyph next to the field.
- [ ] Pick a valid JSON export and import it. After the dialog closes, the "Import from JSON" button briefly shows a green check + "Imported" before reverting.
- [ ] `pnpm tsc` and `pnpm test src/components/AnnotationBrowser/AnnotationImport.test.ts` both pass.

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationImport.vue
+++ b/src/components/AnnotationBrowser/AnnotationImport.vue
@@ -4,6 +4,7 @@
       <v-btn
         v-bind="{ ...activatorProps, ...$attrs }"
         :disabled="!isLoggedIn"
+        :color="showSuccess ? 'success' : undefined"
         v-description="{
           section: 'Object list actions',
           title: 'Import from JSON',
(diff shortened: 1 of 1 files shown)


arjunrajlab:
@codex review

PR #1144, opened 2026-05-08
[codex] Mark annotations raw in store

### LINKED ISSUE #1136: Per-annotation structuredClone in updateMultipleAnnotations is slow on large updates

arjunrajlab:
## Summary

Fixes #1136 by making annotation objects stored in the annotation Vuex module raw and cloning update drafts from `toRaw(annotation)` before `structuredClone`.

## Details

- Mark annotations raw when they enter the store via single-add, batch-add, and full replacement paths.
- Clone update drafts with `structuredClone(toRaw(annotation))` so the bulk update path avoids walking Vue proxy traps over large coordinate arrays.
- Replace the annotations array on add/update mutations so existing Vue watchers still observe annotation changes even though the entries themselves are raw.
- Add focused mutation-reactivity tests for raw annotations and array-reference replacement.

## Validation

- `pnpm exec vitest run src/store/annotation-mutations.test.ts`
- `pnpm exec eslint src/store/annotation.ts src/store/annotation-mutations.test.ts`
- `pnpm exec vue-tsc --noEmit`
- `git diff --check`

Note: the targeted Vitest run still emits the existing jsdom/Vuetify CSS parse noise and a Vite websocket EPERM warning, but exits 0 with all tests passing.

### DIFF
--- a/src/store/annotation-mutations.test.ts
+++ b/src/store/annotation-mutations.test.ts
@@ -11,7 +11,15 @@
  */
 import { describe, it, expect, vi } from "vitest";
 import { createStore } from "vuex";
-import { defineComponent, watch, computed, nextTick, h } from "vue";
+import {
+  defineComponent,
+  watch,
+  computed,
+  nextTick,
+  h,
+  markRaw,
+  isReactive,
+} from "vue";
 import { mount } from "@vue/test-utils";
 
 // Minimal mock of the annotation store's connection state and mutations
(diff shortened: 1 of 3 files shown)


arjunrajlab:
@codex review

arjunrajlab:
Actually, upon testing, this makes bulk edits a lot slower, weirdly.

PR #1155, opened 2026-05-15
Add optional CSV column-name sanitization

arjunrajlab:
## Summary
- Added an optional `sanitizeColumnNames` flag to CSV export on the backend, frontend dialog, TypeScript store client, and Python accessor.
- Sanitized headers now replace spaces, slashes, commas, punctuation, and other non-alphanumeric characters with underscores for R-friendly column names.
- Added a default-off checkbox in the CSV export dialog and kept preview/download behavior aligned with the server.
- Added backend tests for the new sanitization behavior and request plumbing, plus frontend tests for the dialog and export API.

## Testing
- `tox -e py311 -- upenncontrast_annotation/test/test_export.py` passed.
- `pnpm test -- run src/store/ExportAPI.test.ts src/components/AnnotationBrowser/AnnotationCSVDialog.test.ts` passed.
- `pnpm tsc` passed.
- `nimbusimage/tests/test_export.py` passed using the existing tox Python environment with package dependencies installed.

### DIFF
--- a/src/components/AnnotationBrowser/AnnotationCSVDialog.vue
+++ b/src/components/AnnotationBrowser/AnnotationCSVDialog.vue
@@ -297,6 +297,29 @@ const DISPLAY_CHAR_LIMIT = 10000;
 // so this client-side preview matches the server-generated CSV exactly.
 const UNSAFE_CSV_COLUMN_CHARS = /[^A-Za-z0-9_]+/g;
 
+interface CsvColumn {
+  name: string;
+  // isQuoted controls value quoting, not just header quoting: Tags joins
+  // multiple tags with ", " and Name is freeform user text, so these stay
+  // quoted regardless of sanitization. Keeping the flag on the column
+  // (rather than a separate parallel array) means it cannot drift out of
+  // alignment with the name when fixed columns are added/moved/removed.
+  isQuoted: boolean;
+}
+
+// Fixed columns for CSV export. Mirrors CSV_FIXED_COLUMNS in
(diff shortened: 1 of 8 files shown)


src/components/AnnotationBrowser/AnnotationCSVDialog.vue
```diff
@@ -430,7 +474,7 @@ async function generateCSVStringForAnnotations() {
       "Tags",
       "Shape",
       "Name",
-    ];
+    ].map(getCsvColumnName);
     const quotes = [true, false, false, false, false, true, true, true];
```
pchoisel: It would be better to have a class `CsvColumn` like you have in the backend that would have a `name` and `is_quoted` attribute to  hold the fields.  
Because using this method, if you want to add/move/remove fields, the "quotes" array will be wrong

pchoisel (commented):
LGTM otherwise

src/components/AnnotationBrowser/AnnotationCSVDialog.vue
arjunrajlab: Done in 5a14a002. Modeled columns as a single `CsvColumn` array of `{name, isQuoted}` (mirroring the backend dataclass) and derive the parallel `fields`/`quotes` arrays only at the `Papa.unparse` call, since that API requires them separate. The hardcoded positional `quotes` array is gone, so the flag travels with its column and can't drift when fixed columns are added/moved/removed.

Also rewrote the regression test to look up each fixed column's quote flag by name instead of a hardcoded index (covering all 8 fixed columns), and verified it fails on a deliberate quote-flag drift.

PR #1160, opened 2026-05-16
[codex] Fix unroll image tile redraws

### LINKED ISSUE #1159: Unroll can render annotations without image tiles
## Issue

While testing the ImageViewer map-mutation cleanup, dimension/layer unroll was observed to render annotations while the underlying image tiles were missing/blank. This also appears to reproduce on `master`, so it is likely unrelated to PR #1143.

A one-off GeoJS console error was also seen once during testing:

```text
geojs.js?v=f53b0c9f:103451 Uncaught TypeError: m_viewer.renderWindow(...)._init is not a function
    at HTMLCanvasElement.<anonymous> (geojs.js?v=f53b0c9f:103451:54)
```

That error was not reproducible and may be unrelated, but it is recorded here in case it turns out to correlate with map teardown/recreation.

## Observed behavior

- Toggle/use unroll in the live viewer.
- Annotation geometry is still visible in the unrolled layout.
- Image tiles do not appear.

## Code path notes

The annotation and image unroll paths are currently split:

- `AnnotationViewer.vue` directly consults `store.unrollXY`, `store.unrollZ`, and `store.unrollT` when filtering annotations and offsetting coordinates via `unrollIndexFromImages()`.
- Image tiles depend on `ViewerToolbar.vue` calling `store.refreshDataset()` when unroll flags change.
- `setSelectedDataset()` then refetches the dataset with `{ unrollXY, unrollZ, unrollT }`.
- `parseTiles()` rewrites unrolled dimensions to `-1`, and `ImageViewer.vue` later builds tile URLs from `store.layerStackImages`.

That means annotations can continue to render from the live store flags even if the image path fails to complete the dataset/histogram/tile URL refresh. In `_setTileUrls()`, missing `fullUrls[0]`, `urls[0]`, or `baseQuadOptions` causes the image layers to be hidden, which matches the blank-image symptom.

## Suggested investigation / fix

1. Confirm whether the repro is the XY/Z/Time `Unroll` checkboxes, the `Layers: Unroll` radio mode, or both.
2. Add a focused regression test for an unrolled dataset where `layerStackImages` contains multiple `images`, `urls`, and `fullUrls`, then assert that `ImageViewer` assigns tile URLs instead of falling back to blank/hidden layers.
3. Make the image path mirror the annotation path more explicitly, or reset/force the dependent image state when unroll flags change. Candidate fixes to evaluate:
   - reset per-layer `_histogram` state when `refreshDataset()` runs with changed unroll flags, so unrolled image histograms and tile URLs are regenerated reliably;
   - make `getLayerImages()` / `layerStackImages` derive unrolled lookup values from the active `store.unrollXY/Z/T` flags rather than relying only on the re-parsed dataset axes;
   - force a redraw after unrolled histograms resolve if `_setTileUrls()` initially hid the layers because URLs were not ready.

## Expected behavior

Unroll should show both image tiles and annotations in the same unrolled coordinate layout.

arjunrajlab:
## Summary

- Re-request image layer histograms from `ImageViewer` when tile URLs are missing during unroll and redraw after they resolve.
- Add a reactive histogram cache revision so per-layer histogram debounce state is invalidated when Girder caches are flushed during dataset refresh.
- Add focused `ImageViewer` regression tests for unrolled multi-image tile URLs and missing-histogram redraw scheduling.

## Root Cause

Unroll annotations read live unroll flags directly, but image tiles depend on refreshed dataset tile URLs and resolved histograms. After `refreshDataset()` flushed Girder histogram caches, layer-local `_histogram` state could still consider its previous image set current, leaving `layerStackImages` without URLs. `_setTileUrls()` then hid image layers while annotations continued rendering.

## Validation

- `pnpm exec vitest run src/components/ImageViewer.test.ts`
- `pnpm exec vitest run src/components/ViewerToolbar.test.ts`
- `pnpm run tsc`
- `pnpm exec eslint src/components/ImageViewer.vue src/components/ImageViewer.test.ts src/store/GirderAPI.ts src/store/index.ts src/store/model.ts`
- `git diff --check`

Closes #1159

### DIFF
--- a/src/components/ImageViewer.test.ts
+++ b/src/components/ImageViewer.test.ts
@@ -1091,7 +1091,7 @@ describe("ImageViewer", () => {
   // ---- 13. _setTileUrls ----
 
   describe("_setTileUrls", () => {
-    it("assigns unrolled tile URLs and skips redraw when hist is ready", () => {
+    it("assigns unrolled tile URLs and skips histogram fetch when hist is ready", () => {
       wrapper = mountComponent();
       vi.clearAllMocks();
 
@@ -1130,7 +1130,7 @@ describe("ImageViewer", () => {
       expect(mockedStore.getLayerHistogram).not.toHaveBeenCalled();
     });
 
-    it("requests a histogram redraw when tile URLs are not ready", async () => {
+    it("requests a histogram fetch when tile URLs are not ready", async () => {
       wrapper = mountComponent();
       vi.clearAllMocks();
 
--- a/src/components/ImageViewer.test.ts
+++ b/src/components/ImageViewer.test.ts
@@ -115,6 +115,7 @@ vi.mock("@/store", () => {
       setCameraInfo: vi.fn(),
       setDrawAnnotations: vi.fn(),
       setShowTooltips: vi.fn(),
+      getLayerHistogram: vi.fn().mockResolvedValue(null),
     }),
   };
 });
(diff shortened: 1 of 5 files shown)


arjunrajlab:
I will say that I've noticed a slight performance regression in that there's a bit of stutter when scrubbing through time. I think that the stutter seems to happen when it's loading in a new quad tile, but not exactly sure.

PR #1179, opened 2026-05-29
chatbot: document shipped Zenodo project publishing in system prompt

arjunrajlab:
## Summary
Updates the chatbot system prompt (`devops/girder/plugins/girder-claude-chat/system_prompt_2.txt`) to reflect that Zenodo project publishing has shipped.

- Replaced the stale bullet "Prepare data for future export to repositories like Zenodo" with a line describing the shipped feature, pointing to the new section.
- Fixed a related stale note in **Publication Metadata** that still said exporting to repositories was "(coming soon)".
- Added a dedicated **Publishing Projects to Zenodo** section covering: what it is, what gets uploaded, the token → upload → publish workflow, irreversibility, versioning/concept DOIs, the sandbox toggle, and limits.

This gives the chatbot accurate guidance to point users to when they ask about data-sharing policies, journal/funder archiving requirements, getting a DOI, or making imaging data publicly available.

## Test plan
Documentation-only change to a prompt text file; no code paths affected.


arjunrajlab:
Good catch — verified against the backend:

- `ProjectModel.addCollection` (and the frontend `addCollectionToProject`) only appends to `meta.collections`; member datasets are not copied into `meta.datasets`.
- `zenodo_job.py` iterates `project['meta']['datasets']` for image files + annotation JSON, and `project['meta']['collections']` only for config JSONs — it never expands a collection into its member datasets.

So a collection-only project would indeed mint a DOI with just the collection config, missing the image data. Rather than change upload behavior in this docs-only PR, I added a caveat to the prompt (5dc2279a) telling the chatbot to instruct users to add datasets *directly* when they want the actual images/annotations archived.

PR #1203, opened 2026-06-24
Stub/hydration annotation architecture + server-side list

arjunrajlab:
## Summary

Introduces a **stub / hydration architecture** so NimbusImage stays responsive on very large annotation datasets (validated on HCR 26K and Xenium **708K**). Annotations load as lightweight **stubs** (centroid + metadata, no coordinates) and are selectively **hydrated** (full coordinates fetched on demand) based on viewport, size, and selection. Property values and the annotation list go lazy/server-side to match.

Full architecture + as-built notes: `codebaseDocumentation/ANNOTATION-STUBS.md` and `ANNOTATION-LIST-SERVER-SIDE-DESIGN.md`.

## What's included

- **Stub/hydrate core** — server `/upenn_annotation/stubs` + `/hydrate`; frontend stub map, spatial index, viewport-driven hydration with an accumulating LRU cache; a size gate so datasets under the render budget behave exactly like `master`.
- **Server-side annotation list** — `POST /upenn_annotation/list` (+ `/list/ids`): filter/sort/paginate server-side via Mongo aggregation, with a dual-mode `AnnotationList.vue` (client list unchanged below threshold; `v-data-table-server` above it). PV-driven queries keep `/list` fast at 708K.
- **Lazy property values** — viewport-scoped value loading + server-side per-property "uncomputed" counts, so the full value map is never resident in stub-only mode.
- **Rendering perf** — density-adaptive, zoom-aware render/hydration budget; unified pan+zoom hysteresis; O(N) viewport selection + incremental draw (per-pan main-thread block ~3.3 s → ~0.82 s at 708K).
- **Settings** — power-user `visibilityConfig` (thresholds/budgets) with bounds + clamp-on-blur validation behind an "Advanced settings for large numbers of annotations" disclosure.

## Branch review

A full branch review (`codebaseDocumentation/BRANCH-REVIEW-stub-annotations.md`) found 26 issues; **all are addressed (TDD)** — each carries a ✅ Resolved note with its covering test. Highlights:

- Backend: shared input validation (malformed input → 400, not 500); `Annotation.distinctDatasetIds` model helper; consistent `findByAnnotationIds` shape; request-size sanity ceilings + a 5-min `maxTimeMS` query bound. (Per-IP rate limiting was intentionally **deferred to the edge proxy**.)
- Frontend: count-failure routes to stub-only mode instead of the OOM full fetch; stale-response guard on the lazy value fetch; `onBeforeUnmount` cancels all debounced/throttled fns; `stubPerf` dev-gated; assorted perf, helper extraction, and type tidy-ups.

## Verification

- `pnpm tsc` — clean
- `pnpm lint:ci` — 0 warnings
- `pnpm test` — 2377 / 2377
- backend `tox` — 271 passed · `flake8` — clean (whole plugin)
- Live smoke test against the rebuilt backend container

**Reviewer note (needs a human eye, not assertable in jsdom):** in server-list mode, change a filter while on page 3 and confirm the footer snaps back to page 1.

https://claude.ai/code/session_01WBfPTkdddxm4wbG4rTyD5w


arjunrajlab:
Thanks @codex — addressed all four findings (TDD), and audited the branch for each pattern:

**[P1] "Select all" hydrating the entire dataset.** `ensureHydrated` now caps its fetch at `visibilityConfig.maxHydrated`, and `hydrationCacheCap` is enforced as a HARD ceiling via a new pure `planHydrationEvictions` helper — it protects selected ids but evicts selected LRU too once the protected set alone exceeds the cap, so a select-all can't post a giant `/hydrate` request or grow the cache unbounded. Audit: `ensureHydrated` is the single choke point for on-demand hydration. Verified live on the 708K Xenium dataset: selecting all 708,983 fires **one** bounded `/hydrate` and the hydration cache settles at ~24K (< the 40K cap), not 708K.

**[P2] Geometry edit/combine on unhydrated stubs.** Both paths now detect an unhydrated target and show a toast ("Annotation not fully loaded — zoom in to fully load it…") instead of silently failing/no-op'ing; combine also re-checks the first-selected annotation in case it was LRU-evicted between clicks. Audit: of the selection handlers, only combine and edit need coordinates — select/connections/tagging operate on ids/tags/centroid and are stub-safe.

**[P3] Bare `ObjectId()` on client input → uncaught 500.** The new public endpoints — `stubs`, `hydrate`, `listAnnotations`, `listAnnotationIds` (and the property-values `batch`) — now route ids through `requireObjectId` (→ 400), verified live (malformed datasetId/id → 400 on all of them). Audit: the **pre-existing** `find`/`count` (annotation + propertyValues) and the `deleteMultiple` id helpers share the pattern but predate this branch — filed as follow-up #1204 rather than widening this PR's diff.

**[P4] Empty values-mode property filter routing into the PV-driven path.** Added `dropNoOpPropertyFilters`, which normalizes away no-op filters (empty `values`, or a range with neither bound) before the model chooses a path — so an inactive filter behaves like no filter on `/list`, `/list/ids`, and the count (no longer dropping annotations that lack a value document).

Verified: `pnpm tsc`, `pnpm lint:ci` (0 warnings), `pnpm test` (2385), backend `tox` (276) + `flake8` (clean), plus the live browser/curl checks above. New tests cover each finding.

@codex review

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
```diff
@@ -436,3 +478,168 @@ def compute(self, params, *args, **kwargs):
...
+
+    @access.public(scope=TokenScope.DATA_READ)
+    @describeRoute(
+        Description("Annotation IDs matching list filters")
+        .param("body", "JSON: {datasetId, filters}", paramType="body")
+        .errorResponse()
+        .errorResponse("Read access denied.", 403)
+    )
+    @memoizeBodyJson
+    def listAnnotationIds(self, params, *args, **kwargs):
```
pchoisel: I don't see why you are using params, *args and **kwargs when you are just expecting a single parameter which is the body of the request  
Why not handling it like here for example: https://github.com/arjunrajlaboratory/NimbusImage/blob/master/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/datasetView.py#L66

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
```diff
@@ -436,3 +478,168 @@ def compute(self, params, *args, **kwargs):
...
+    @access.public(scope=TokenScope.DATA_READ)
+    @describeRoute(
+        Description("List annotations (paged), stub-shaped + property values")
+        .param("body", "JSON: {datasetId, filters, sort, propertyPaths, "
+                       "offset, limit}", paramType="body")
+        .errorResponse()
+        .errorResponse("Read access denied.", 403)
+    )
+    @memoizeBodyJson
+    def listAnnotations(self, params, *args, **kwargs):
```
pchoisel: Same remark as above

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
```diff
@@ -198,6 +218,560 @@ def deleteMultiple(self, annotationStringIds):
...
+        inclusive -> $in (has any); exclusive -> exactly that set.
+        """
+        match = {"datasetId": datasetId}
+        if filters.get("shape"):
+            match["shape"] = filters["shape"]
+
+        tags = filters.get("tags") or {}
+        tagValues = tags.get("values") or []
+        if tagValues:
+            if tags.get("exclusive"):
```
pchoisel, line 319: Does this mean that one cannot create a tag named `exclusive` ?  
In that case, it should be guarded in the front-end or it should be a separate request argument

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
```diff
@@ -198,6 +218,560 @@ def deleteMultiple(self, annotationStringIds):
...
+        return stages
+
+    def _propertySortAddFields(self, sort, valueBase="_pv.values."):
+        if sort and sort.get("type") == "property":
+            ref = "$" + valueBase + ".".join(sort["key"])
+            return [{"$addFields": {
+                "_sortValue": ref,
+                "_hasSortValue": {"$cond": [
+                    {"$ne": [{"$ifNull": [ref, None]}, None]}, 1, 0,
+                ]},
```
pchoisel, line 451: This is really hard to read, can you add a comment explaining this stage ?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
```diff
@@ -198,6 +218,560 @@ def deleteMultiple(self, annotationStringIds):
...
+        pipeline = self._buildListMatchStages(datasetId, filters)
+        if filters.get("propertyFilters"):
+            pipeline += self._lookupStages()
+            pipeline += self._propertyFilterStages(filters)
+        pipeline.append({"$project": {"_id": 1}})
+        cursor = self._aggregate(self.collection, pipeline)
+        return [str(doc["_id"]) for doc in cursor]
+
+    # Annotation fields allowed as a sort key (field-type sort).
+    _SORTABLE_FIELDS = {"location.XY", "location.Z", "location.Time",
```
pchoisel: Those constants should be placed at the top of the class definition for easy access

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
```diff
@@ -198,6 +218,560 @@ def deleteMultiple(self, annotationStringIds):
...
+
+    @property
+    def _pvModel(self):
+        """The property-values model (lazy, cached) for PV-driven queries."""
+        pv = getattr(self, "_pvModelCache", None)
+        if pv is None:
+            from .propertyValues import AnnotationPropertyValues
+            pv = AnnotationPropertyValues()
+            self._pvModelCache = pv
+        return pv
```
pchoisel: Girder already lazy-loads and caches models, see https://github.com/girder/girder/blob/master/girder/models/model_base.py#L73

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
```diff
@@ -198,6 +218,560 @@ def deleteMultiple(self, annotationStringIds):
...
+        }}})
+        pipeline.append({"$replaceRoot": {"newRoot": "$_ann"}})
+        pipeline.append({"$project": {"coordinates": 0}})
+        return pipeline
+
+    def _pvHasValueCount(self, datasetId, sort):
+        valueKey = "values." + ".".join(sort["key"])
+        pipeline = [
+            {"$match": {"datasetId": datasetId, valueKey: {"$ne": None}}},
+            {"$count": "n"},
```
pchoisel: What about naming this field `count` it's easier to debug later ?

pchoisel (commented):
LGTM otherwise

pchoisel:
Actually, before merging, let's inspect the use of @memoizeBodyJson, because I don't think it is useful everywhere

Issue #1207, opened 2026-06-27
[Client] nimbusimage package robustness: retry/backoff, stable cursor iterator, compute() validation, clearer job 401


arjunrajlab:
## Summary

Robustness fixes to the `nimbusimage` Python client (`nimbusimage/`) that require **no backend change** and can land immediately. These are four issues from the June 2026 NimbusImage API report (Grant Kinsler, Bhatt Lab). A local analysis lives at `API_ISSUE_REPORT_ANALYSIS.md` (uncommitted).

## Issues addressed

- **NIM-006 (client half) — Transient 5xx under normal load.** The client has no retry; every transient 502/503/504 surfaces to the caller, forcing users to hand-roll 8-retry backoff loops.
- **NIM-002 — Offset pagination is unstable during bulk deletion.** `fetch(offset)→delete→offset+=N` skips records because `$skip` is positional and shifts after each delete.
- **NIM-004 — Batch mode silently ignores `location`.** When batch ranges are supplied, the `location=` passed to `compute()` is silently dropped with no warning.
- **NIM-005 (client half) — Confusing 401 from `job.wait()`.** Scoped API keys get a 401 whose message misattributes the missing scope.

### NIM-006 — add retry-with-jitter
`create_client` (`nimbusimage/nimbusimage/_girder.py:52`) builds a bare `girder_client.GirderClient`; upstream `sendRestRequest` issues a single request and raises `HttpError` on any non-OK status — no `urllib3.Retry`, no `HTTPAdapter`, no backoff.
- **Fix:** mount a `requests.adapters.HTTPAdapter(max_retries=urllib3.util.Retry(status_forcelist=[502,503,504], backoff_factor=…, respect_retry_after_header=True))` (plus jitter) onto the session the client uses, in `create_client`. One change centralizes retry for every call and covers the 503 (backend-saturation) case too.

### NIM-002 — expose a stable cursor iterator
`AnnotationAccessor.list()` (`nimbusimage/nimbusimage/annotations.py:22-50`) exposes only `(shape, tags, limit, offset)`. The backend `find` route **already supports a stable `_id` cursor via `afterId`** (`server/api/annotation.py:361-364`, on `master`) — the client just doesn't use it.
[...]

PR #1209, opened 2026-06-27
fix(nimbusimage): client robustness — retry, stable cursor, compute validation, clear job 401 (#1207)

arjunrajlab:
Fixes the four **client-only** issues from #1207 (June 2026 NimbusImage API report, Grant Kinsler / Bhatt Lab). All are independent of any backend change and were implemented test-first.

## Issues addressed

- **NIM-006 — transient 5xx under load.** `create_client` now mounts a `urllib3.Retry` adapter on the `requests.Session` (`status_forcelist=502/503/504`, exponential backoff + jitter, `respect_retry_after_header`). Only idempotent methods are retried — **POST is deliberately excluded** so a retried compute submission can't create duplicate jobs/annotations (NIM-007). `raise_on_status=False` keeps girder_client's usual `HttpError` on final failure.
- **NIM-002 — offset pagination unstable during deletion.** Added `AnnotationAccessor.iter_all()`, a generator that walks the backend's stable `afterId` cursor (plus an `after_id=` kwarg on `list()`). Mutation-safe for delete-as-you-go loops; documents that `offset` is not.
- **NIM-004 — batch mode silently dropped `location`.** `compute()` now raises `ValueError` when both `location=` and `assignment=` are supplied, instead of silently ignoring `location`. Docstring clarified.
- **NIM-005 — confusing 401 from `job.wait()`.** `Job.refresh()`/`wait()` catch the raw 401 from `job/{id}` (which misattributes the missing scope as `jobs.job_<id>`) and re-raise a `PermissionError` naming the real culprit, `core.user_auth`. Required scope set documented in the README.

## Testing

- **235 unit tests pass** (was 218), flake8 clean. New coverage includes a real local-HTTP-server test proving `503,503,200 → success` (and no retry on 4xx), a delete-as-you-go cursor simulation, and an infinite-loop guard.
- **Verified live** against a freshly rebuilt backend (Girder 5.0.11): `iter_all` walks the real `afterId` cursor and left 0 records behind during delete-as-you-go; the NIM-005 401 was reproduced with a real scoped API key and confirmed translated into the clear `core.user_auth` message.
- Added 2 live integration tests for `iter_all` under `tests/integration/`.

## Review notes (from `/branch-review`)
Applied all five findings before committing: declared `requests`/`urllib3` as direct deps, added an infinite-loop guard to `iter_all` (with test), documented the `_id`-ordering assumption, noted `PermissionError` in `wait()`'s docstring, and switched to `collections.abc.Iterator`.

### DIFF
--- a/nimbusimage/README.md
+++ b/nimbusimage/README.md
@@ -23,6 +23,8 @@ The recommended setup uses a **Girder API key**, which is persistent and doesn't
 - **nimbusimage.com (hosted):** Email **[support@cytopixel.com](mailto:support@cytopixel.com)** with your account email address to request an API key.
 - **Local/self-hosted server:** In the Girder admin UI, go to **Users** > select the user > **Edit User** > **API Keys** > create a new key and copy the key string.
 
+**Required scopes.** The simplest option is a **full-access key** (leave the scope list empty), which works for everything including polling job status. If you instead create a *scoped* key, it must include **`core.user_auth`** — without it, `job.wait()` / `job.refresh()` fail with a confusing `401 Unauthorized` even though you own the job. Add **`jobs.rest.list_job`** as well so job logs can be read. Recommended scoped-key set: `core.data.read`, `core.data.write`, `core.data.own`, `core.user_info.read`, `core.user_auth`, `jobs.rest.list_job`.
+
 Set environment variables for persistent access:
 
 ```bash
--- a/nimbusimage/nimbusimage/_girder.py
+++ b/nimbusimage/nimbusimage/_girder.py
@@ -10,6 +10,40 @@ from __future__ import annotations
 import os
 
 import girder_client
+import requests
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+# NIM-006: transient 502/503/504 responses happen under normal load
+# (backend saturation, proxy timeouts). Retry them with jittered
+# exponential backoff so callers don't have to hand-roll retry loops.
+#
+# Only idempotent methods are retried (urllib3's default allowed set:
+# GET/HEAD/PUT/DELETE/OPTIONS/TRACE). POST is deliberately excluded so a
+# retried compute submission can't create duplicate jobs/annotations
+# (NIM-007). raise_on_status=False lets girder_client raise its usual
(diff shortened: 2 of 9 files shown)


arjunrajlab:
Good catch — fixed in 19c966b8.

**Targeted fix:** `backoff_jitter` was added in urllib3 2.0, and pyproject declared `urllib3` with no lower bound, so on urllib3 1.26.x `create_client()` would crash with `TypeError` before any request. Rather than impose a version floor (which can cause resolution conflicts for a library), I feature-detect the kwarg via `inspect.signature(Retry.__init__)` and only pass it when supported. Backoff still applies without jitter on older urllib3 — we just lose the thundering-herd spread. Added two tests (jitter applied when supported; omitted and still-builds when not).

**Branch-wide audit (same pattern — a third-party kwarg that needs a newer dependency than the declared floor guarantees):** `backoff_jitter` was the only instance. `respect_retry_after_header`, `raise_on_status`, `status_forcelist`, and `backoff_factor` all exist in urllib3 1.26; `collections.abc.Iterator` is fine on this package's Python >=3.10 floor. No other gating needed.

arjunrajlab:
Filed #1210 for the backend root cause: the `find` handler defaults the sort to `lowerName` (`annotation.py:312`) even though the route declares `defaultSort="_id"` (`:308`) and hints `(datasetId,_id)` (`:342`). That mismatch is why `iter_all()` here forces `sort=_id&sortdir=1` (560973ab) — without it the `afterId` cursor only works by accident (annotation docs have no `lowerName`, so the null keys happen to come back in `_id` order via the hinted index).

**Sequencing:** if #1210 lands first, we can merge master in and the server will default to `_id` ordering for every caller (and drop the blocking in-memory sort).

**On keeping vs. removing the client-side `sort=_id`:** I'd lean toward keeping it. This package ships to PyPI and talks to arbitrary/older server versions that may lag the backend fix, so requesting `sort=_id` explicitly keeps the cursor correct regardless of the server's default — and it's a no-op-cost line that also benefits from the index-aligned sort on fixed backends. So #1210 is better framed as "fix it for all clients + remove the blocking sort" than as a prerequisite for deleting this line.

PR #1211, opened 2026-06-27
fix(annotation): default find sort to _id, not dead "lowerName"

### LINKED ISSUE #1210: Backend: `find` annotation route default sort is `lowerName`, contradicting declared `defaultSort="_id"` and the `(datasetId,_id)` index hint
## Summary

The annotation `find` route (`GET /upenn_annotation`) is internally inconsistent about its sort, which (a) forces a blocking in-memory sort at scale and (b) makes the `afterId` cursor stable only **by accident**.

In `devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py`:

- **Line 308** declares the route's paging default as `_id`:
  ```python
  .pagingParams(defaultSort="_id")
  ```
- **Line 312** then *overrides* it with `lowerName`:
  ```python
  limit, offset, sort = self.getPagingParameters(params, "lowerName")
  ```
- **Line 342** hints the `(datasetId, _id)` index:
  ```python
  ).hint([("datasetId", 1), ("_id", 1)])
  ```

So when a client sends no `sort` param, the server sorts by **`lowerName`** — a field annotation documents don't have — while hinting an index that provides `(datasetId, _id)` order.

## Impact

1. **Blocking in-memory sort / 504s at scale.** Because the sort field (`lowerName`) isn't provided by the hinted `(datasetId,_id)` index, MongoDB performs a blocking in-memory sort. With full `coordinates` arrays and large result sets this is slow and can exceed the proxy `timeout server`. This is the structural root cause behind the NIM-001 (silent drop on large lists) and NIM-006 (504s on big fetches) symptoms from the June 2026 API report.

2. **`afterId` cursor stability is accidental.** The cursor (`annotation.py:331-334`) advances `_id > afterId` and the client uses `page[-1]._id` as the next cursor — which is only correct if results come back in ascending `_id` order. Today that holds **only because** annotation docs have no `name`/`lowerName`, so the keys are all null and the hinted index happens to emit `_id` order. If annotations ever gained a name, or Mongo's tie-break for equal keys changed, cursor pagination could silently **skip, duplicate, or loop** on records.

## Proposed fix

Align the handler's default with the declared `defaultSort` and the index hint:

```python
# line 312
limit, offset, sort = self.getPagingParameters(params, "_id")
```

This makes the default sort `_id`, which:
- matches the route's declared `defaultSort="_id"` (line 308),
- aligns with the `(datasetId,_id)` hint → no blocking in-memory sort (fast, streams in index order, scales to any size), and
- makes the `afterId` cursor correct **by design** rather than by accident.

Clients that need a different sort can still pass `sort=`/`sortdir=` explicitly (already supported by `pagingParams`).

Optionally also bound the query with `.maxTimeMS(...)` so a pathological query fails deterministically rather than silently mid-stream (NIM-001).

## Context / related

- The `nimbusimage` Python client now works around this by forcing `sort=_id&sortdir=1` in `iter_all()` (PR #1209), but the underlying backend default should be fixed so every caller benefits and the cursor isn't relying on the all-null-`lowerName` coincidence.
- Reported originally as part of NIM-001 / NIM-006 in the June 2026 NimbusImage API report (Bhatt Lab).

arjunrajlab:
Closes #1210.

## What

Change the `GET /upenn_annotation` (`find`) handler's default-sort fallback from `"lowerName"` to `"_id"` (`annotation.py:312`), aligning it with the route's declared `.pagingParams(defaultSort="_id")` (`:308`) and the `(datasetId,_id)` index hint (`:342`). Adds a regression test.

## Why this is a cleanup, not a runtime perf fix

The investigation for #1210 overturned the issue's premise. On the current Girder 5.x backend the `"lowerName"` fallback is **dead code**: `@autoDescribeRoute`'s `pagingParams(defaultSort="_id")` registers a `sort` param with default `"_id"` and injects it into the request when the client omits `sort` (`girder/api/describe.py:760-761`), so `getPagingParameters` takes its `if 'sort' in params` branch and uses `"_id"` — never reaching `"lowerName"`.

**Proven on the live backend** (Girder 5.0.11, a 26,141-annotation dataset), `GET /upenn_annotation?datasetId=…` with no `sort`:
- `command.sort: {"_id": 1}`
- `planSummary: IXSCAN { datasetId: 1, _id: 1 }` — **no SORT stage**, 5 keys examined, 0 ms.

So there is no blocking in-memory sort, and the `afterId` cursor was already correct *by design* (the injected `_id` default), not "by accident" via all-null `lowerName` keys.

## What the change buys

- Removes dead/misleading code — the `"lowerName"` arg is exactly what made a static read infer a `lowerName` sort.
- Keeps the handler correct if `defaultSort` is ever dropped from the decorator (which would otherwise silently reintroduce a `lowerName` sort + blocking in-memory sort).
- **No observable behavior change** on the current backend.

## Tests

`test_annotations.py::TestFindDefaultSort` injects a `lowerName` on docs in reverse-`_id` order and asserts the route returns `_id` order (it would return reverse order if it sorted by `lowerName`).

Verified **red→green**: with the decorator default temporarily set to `None` (exposing the line-312 fallback), the old `"lowerName"` failed the test and `"_id"` passed; the decorator default was then restored for defense-in-depth. Full plugin suite: **188 passed**, flake8 clean.

## Note on NIM-006 (504s)

The 504s were reproduced on the latest backend, so they are **not** caused by this (nonexistent) blocking sort. Real suspects — unbounded query (no `.maxTimeMS`), pre-stream `cursor.count()` second pass, `coordinates` payload size vs proxy timeout, backend saturation — are being investigated separately and are out of scope here.

## Deploy note

Baked-in backend plugin change — needs `docker compose build girder && docker compose up -d` to take effect on a running container.

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/annotation.py
@@ -309,7 +309,7 @@ class Annotation(Resource):
         .errorResponse()
     )
     def find(self, params):
-        limit, offset, sort = self.getPagingParameters(params, "lowerName")
+        limit, offset, sort = self.getPagingParameters(params, "_id")
 
         # First, check dataset permissions explicitly
         datasetId = ObjectId(params["datasetId"])
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/test/test_annotations.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/test/test_annotations.py
@@ -561,3 +561,54 @@ class TestInlinePropertiesField:
             {"annotationId": result["_id"]}
         ))
         assert values == []
+
+
+@pytest.mark.usefixtures("unbindLargeImage", "unbindAnnotation")
+@pytest.mark.plugin("upenncontrast_annotation")
+class TestFindDefaultSort:
+    """The GET /upenn_annotation find route must default its sort to _id.
+
+    The route declares ``pagingParams(defaultSort="_id")`` and hints the
+    ``(datasetId, _id)`` index, so the handler's default must match. If it
+    defaults to a field the hinted index doesn't provide (e.g. ``lowerName``),
+    MongoDB performs a blocking in-memory sort, and the ``afterId`` cursor —
+    which advances on ``_id > afterId`` and uses ``page[-1]._id`` as the next
+    cursor — is only stable by accident (real annotation docs simply lack
+    that field, so the null keys happen to emit in ``_id`` order).
+    """
+
+    def testFindDefaultsToIdSort(self, admin, server):
+        folder = utilities.createFolder(
+            admin, "ds", upenn_utilities.datasetMetadata
+        )
+        datasetId = str(folder["_id"])
+
+        # Create several annotations and record their _ids ascending.
+        ids = [
+            Annotation().create(
+                upenn_utilities.getSampleAnnotation(folder["_id"])
+            )["_id"]
+            for _ in range(5)
+        ]
+        idsAscending = sorted(ids)
+
+        # Give each doc a `lowerName` whose ascending order is the REVERSE of
+        # _id order. Real annotations have no lowerName, so a lowerName-default
+        # sort returns them in _id order by accident; injecting a contradictory
+        # lowerName exposes which field the route actually sorts by.
+        for rank, annotationId in enumerate(idsAscending):
+            Annotation().update(
+                {"_id": annotationId},
+                {"$set": {"lowerName": "%03d" % (len(idsAscending) - rank)}},
+            )
+
+        resp = server.request(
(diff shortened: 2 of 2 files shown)


arjunrajlab:
**For reviewers — low priority, no runtime behavior change.** Our testing (live MongoDB profiler on Girder 5.0.11 + a regression test) confirmed the `find` route **already sorts by `_id` by default** today: `@autoDescribeRoute`'s `pagingParams(defaultSort="_id")` injects `sort="_id"`, so the `"lowerName"` fallback at line 312 is never reached (profiler showed `IXSCAN {datasetId:1,_id:1}`, no SORT stage).

So this is a **correctness-by-design cleanup**, not a fix for a live bug — it removes the misleading dead arg and keeps the handler correct if `defaultSort` is ever dropped from the decorator. No rush to merge; review at your convenience.

pchoisel:
LGTM

PR #1231, opened 2026-07-06
AI panel: agent-driven interface control (spec + Option A implementation)

arjunrajlab:
## AI Panel — agent-driven interface control

A "Nimbus AI" panel where the user converses with an agent that **drives the NimbusImage interface** through a curated tool surface (not the raw REST API). Started as a design spec (`codebaseDocumentation/AI_PANEL_SPEC.md`), grew into the full **Option A** implementation, and now **absorbs Nimbus Chat** into the panel — the standalone chat is removed, and the agent covers Q&A too.

### Architecture (Option A)
- **Backend** (`girder-claude-chat`): `POST /claude_agent` relay attaches the API key, server-held system prompt, and tool definitions (`agent_tools.json`), and returns raw content blocks. **Streams** (`client.messages.stream`, 32k output cap) and strips SDK output-only fields (`parsed_output`) so multi-turn tool loops don't 400. Per-user rate limit + request body-size cap. Also serves NimbusImage help topics via `GET /claude_agent/help?topic=<slug>` for the `read_help_topic` tool.
- **Frontend**: executor registry (`src/agent/executors.ts`) mapping each tool to an existing Vuex action, the agent-loop store (`src/store/aiPanel.ts`), and the panel UI (`src/components/AiPanel.vue`) with transcript cards, Run/Cancel gating, stop, and per-turn "revert view changes". Conversation persisted per-user to IndexedDB (`src/agent/conversationStore.ts`). Gated behind login + a `VITE_AI_PANEL_ENABLED` build flag.

### Absorbing Nimbus Chat (this branch)
The standalone chat (`ChatComponent.vue`, `chat.ts`, `ChatAPI.ts`, the `claude_chat` endpoint, `system_prompt_2.txt`) is **removed** — the agent is a capability superset. To carry over chat's value:
- **Hybrid knowledge (reorganized, not compressed):** the 1,487-line chat system prompt was split near-verbatim (~97% of words retained) into a small concepts core (folded into the agent's cached system prompt) + **14 `help/*.md` topic files**, served on demand by a new **`read_help_topic`** tool. Knowledge stays out of the always-on prompt; the agent pulls a topic only when it needs to explain a workflow or how to do something by hand.
- **Persistence:** the panel conversation (display transcript **and** Anthropic wire history) persists to browser IndexedDB — a single record tagged with the user id, restored on same-user reload, wiped when a different user logs in. Screenshots pruned before write; snapshot/revert state not persisted.
- `getToolSuggestions` relocated to `ToolSuggestionsAPI` so the auto tool-suggestion flow survives chat's deletion. Chat toolbar button + its tour anchor/step removed.

### Tool surface (~29 tools)
- **Read/grounding:** `get_interface_state`, `capture_screenshot`, `get_annotation_summary`, `list_annotations` (paged, **hard-capped at 200 rows/call**, steers toward summaries), `list_tools`, `list_workers`, `get_worker_interface`, `list_properties`, `get_property_values` (summary stats), **`read_help_topic`**.
- **View / settings:** `set_location`, `set_camera` (+ `fit` annotations/selection/full), `set_layer_mode`, `update_layer` (+ contrast `view`/`configuration` scope), `set_layer_visibility`, `set_display_options`, `set_view_mode` (2D/3D), `set_scale` (gated), `set_annotation_filter` (tags, current-frame, **property-value ranges**), `select_annotations`, `select_tool`, `undo`/`redo`.
- **Mutating (gated):** `color_annotations`, `tag_annotations`, `run_worker`, `create_tool` (manual/worker), `create_property`, `compute_property`.

Deliberately deferred: bulk `create_annotations`/`delete_annotations`, `navigate` between datasets/collections.

### Scaling (keeps large datasets out of context)
The surface is designed so raw annotation/property data never enters the model's context, even at millions of objects: mutations act on **queries** (`color_annotations` etc. return only an `affectedCount`), reads return **aggregates** (`get_annotation_summary`, `get_property_values` stats), and `list_annotations` — the only raw-row path — is **hard-capped at 200 rows/call**. The real ceiling is the pre-existing in-memory annotation set, not the agent. Server-side query/mutation resolution and managed-agent code execution (Option B) are the documented deferred path for datasets that outgrow memory or actions that can't be expressed as a query (see `AI_PANEL_SPEC.md` §4.1).

### Safety / correctness
- Annotation edit targets validated (no silent "match-all"); tool execution + revert bound to the originating dataset and re-checked before each tool in a batch; conversation cleared when the authenticated user changes; shared-config mutations (incl. `set_scale`) gated.
- **Three code-review rounds addressed** (see `codebaseDocumentation/AI_PANEL_REVIEW.md`), including a persistence boot-sequence bug where storage was wiped on the initial `null` user watcher fire (before the async user fetch resolved) — which would have erased every returning user's conversation on reload — plus unknown-help-topic error surfacing, `compute_property`/`create_property` validation gaps, and a `list_annotations` RangeError on large property sets.
- `syncConfiguration` swallowed-error behavior tracked in #1239.

### Verification
- `pnpm tsc` + `lint:ci` clean; **2360 frontend tests**; backend `tox` 18/18 + flake8.
- **Live browser pass** across the whole feature: multi-turn loop, camera-fit (found + fixed a GeoJS y-axis bug), scale, property-value filtering, 2D/3D, `create_tool` gating; and the chat-merge specifically — chat button gone, `read_help_topic` answering a "how do I count RNA spots?" question end-to-end against the real API, and the conversation restored after a full page reload.

Per-feature design docs under `docs/superpowers/specs/2026-07-10-*.md` (incl. the chat-merge spec + implementation plan). Builds on PR #1224 (tool suggestions) and the `girder-claude-chat` plugin.

https://claude.ai/code/session_012V4i6BV27W4iPwUmUP2rya


pchoisel (commented):
LGTM

PR #1234, opened 2026-07-07
Fix JSON import of connections via new server-side import endpoint

arjunrajlab:
## Problem

Importing annotations from a JSON file was broken when the file contained connections. The JSON export endpoint (`GET /export/json`) dumps raw Mongo documents whose identifier field is `_id`, but the frontend import built its old→new annotation ID map from `annotation.id` — which doesn't exist in exported files. Every connection lookup missed and the import threw `"Can't find the parent or the child of the connection to create"`. Property-value remapping failed the same way. The TypeScript types declared `id` and the test fixtures used `id`, so neither the compiler nor the tests caught it.

## Solution

Instead of patching the client-side remap, move the import to the backend:

### New endpoint: `POST /annotation_import`

- `server/api/dataImport.py` — REST resource: permission check (WRITE on the target dataset), input conversion at the API boundary, single undo-history record via `@recordable`.
- `server/helpers/dataImport.py` — orchestration: sanitizes imported annotation docs to the schema fields, bulk-creates them, builds the old→new ID map (accepting old ids under `_id` or legacy `id`), remaps connection `parentId`/`childId` and property-value annotation IDs, and remaps property IDs through a caller-supplied `propertyIdMap`. Values referencing property IDs missing from the map are skipped (exports can legitimately contain values for properties outside the exported property list). On any failure, everything created so far is explicitly rolled back (annotations, connections, property values) before re-raising.

### Frontend

- `importAnnotationsFromData` now sends the parsed export file through in a single request — no client-side ID remapping remains, so the `_id`/`id` mismatch can't recur. Properties are still created client-side (creating a property also attaches it to the current configuration) and their old→new ID map is passed to the endpoint.
- Failure path: only client-created properties are cleaned up (the backend rolls back its own work); overwrite deletions happen only after a successful import.
- `ISerializedData` is now typed as the real exported document shape (`_id`, optional `id`), making this class of bug visible to the compiler.

## Tests

- **Backend** (`test/test_import.py`): remapping with `_id` keys (server-export shape) and legacy `id` keys, `propertyIdMap` remapping incl. silent skip of unmapped property IDs, rollback when a connection references an unknown annotation (400 + dataset left empty), 403 for users without WRITE access.
- **Frontend** (`src/utils/annotationImport.test.ts`, 10 tests): payload pass-through (including `_id`-only annotations), `propertyIdMap` construction, option gating, failure cleanup, and overwrite semantics.
- `pnpm tsc`, `pnpm lint`, and the full frontend suite (2123 tests) pass; backend files are flake8-clean. Backend pytest requires MongoDB and should run via `tox` / CI.

https://claude.ai/code/session_01Cryuo9WTWBFgX8zR6Xwu6K

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cryuo9WTWBFgX8zR6Xwu6K)_


arjunrajlab:
Pushed `8b8a3340`: merged `master` in and added input-type validation to the `annotation_import` endpoint — malformed body, invalid `datasetId` (bson `InvalidId`), and wrong-typed `annotations`/`connections`/`propertyValues`/`propertyIdMap` now return **400** at the API layer instead of an uncaught 500. Added 6 backend tests; verified live (curl) and via the full plugin suite (200 passed).

@codex review

arjunrajlab:
Thanks — both flagged issues fixed in `a4bb8f20`.

**1. Non-string `datasetId`** (`server/api/dataImport.py`): `ObjectId()` raises `TypeError` for a non-string value (JSON number/bool), not `InvalidId`, so it now catches `(InvalidId, TypeError)` → 400.

**2. Nested property-value map** (`server/helpers/dataImport.py`): each `propertyValues` entry is now `isinstance(..., dict)`-checked → `ValidationException` (400) before `.items()`.

Audited the branch for the same underlying pattern — *iterating/accessing a caller-supplied element without first checking it's an object* — and found two more instances that weren't flagged, fixed identically:

- annotation list entries (`_createAnnotations`): `{"annotations": [123]}` would `.get()` on a non-dict → 500
- connection list entries (`_createConnections`): `{"connections": [123]}` → same

All four now raise `ValidationException` → 400 with the existing rollback intact.

Added 4 tests (non-string `datasetId`; non-object entry in `annotations` / `connections` / `propertyValues`), each asserting 400 + rollback. Full plugin suite: **204 passed**, and verified live against the running endpoint.

@codex review

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/dataImport.py
```diff
@@ -0,0 +1,225 @@
...
+            for field in ANNOTATION_IMPORT_FIELDS
+            if annotation.get(field) is not None
+        }
+        doc["datasetId"] = datasetId
+        docs.append(doc)
+
+    # saveMany() inserts via insert_many(), which preserves list order,
+    # so zipping the pre-creation old ids with the created docs lines
+    # each new document up with the old id it came from.
+    createdDocs = AnnotationModel().createMultiple(docs)
```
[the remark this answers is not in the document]
arjunrajlab: Thanks — I looked into this closely and believe it's a false positive: the import writes **are** recorded, and undo reverts them.

The premise is that the helper's `AnnotationModel()` / `ConnectionModel()` / `PropertyValuesModel()` are *fresh* instances that miss `proxiedModel.startRecording`. But Girder's `Model` base uses the `_ModelSingleton` metaclass (`girder/models/model_base.py`), so `AnnotationModel()` returns the cached singleton — the same instance bound to the `startRecording`/`stopRecording` handlers, with `is_recording=True` during the `@recordable` window. The helper's writes therefore go through the recording path. This is the same mechanism the existing `@recordable` batch endpoints (e.g. `annotation.py::createMultiple`) rely on.

Verified end-to-end with a new regression test (`testImportedDataIsRecordedAndUndoable`, `4aa70464`): import 2 annotations + 1 connection + 1 property value through `/annotation_import`, then `PUT /history/undo` — all three revert to empty. If the writes weren't recorded, undo would be a no-op and the test would fail.

So no code change beyond adding that regression test. Happy to reconsider if there's a scenario I'm missing (e.g. a model that isn't a singleton), but the singleton behavior is exactly what the recording mechanism depends on.

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/dataImport.py
```diff
@@ -0,0 +1,121 @@
...
+from girder.models.folder import Folder
+
+from ..helpers.dataImport import importAnnotationData
+from ..helpers.proxiedModel import recordable, memoizeBodyJson
+
+
+# Helper function to get dataset ID for the recordable endpoint
+
+
+def getDatasetIdFromImportBody(self: "DataImport", *args, **kwargs):
```
pchoisel: `self` is not used and this function is not a class method, you can remove it

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/dataImport.py
```diff
@@ -0,0 +1,121 @@
...
+            propertyIdMap; values for a property id missing from that
+            map are silently skipped, since an export can legitimately
+            include values for properties outside the exported
+            property list.
+        """)
+        .param("body", "The annotation data to import", paramType="body")
+        .errorResponse("Missing datasetId.", 400)
+        .errorResponse("Write access was denied for the dataset.", 403)
+    )
+    @memoizeBodyJson
```
pchoisel, line 69: Is this decorator really useful here ? Usually an import request will only be done once with a given JSON file

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/dataImport.py
```diff
@@ -0,0 +1,121 @@
...
+                "connections must be a JSON array.", code=400
+            )
+        if not isinstance(propertyValues, dict):
+            raise RestException(
+                "propertyValues must be a JSON object.", code=400
+            )
+        if not isinstance(propertyIdMap, dict):
+            raise RestException(
+                "propertyIdMap must be a JSON object.", code=400
+            )
```
pchoisel: Please factorize this with a for loop

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/dataImport.py
```diff
@@ -0,0 +1,225 @@
...
+    Server-generated exports key the old id under "_id"; some legacy
+    frontend exports use "id" instead. Returns None if neither key is
+    present, in which case the annotation is still imported but has
+    no old id to remap connections/property values against.
+    """
+    oldId = annotation.get("_id", annotation.get("id"))
+    return str(oldId) if oldId is not None else None
+
+
+def _createAnnotations(datasetId, annotations):
```
pchoisel, line 39: Shouldn't you delete the old annotations if they exist ?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/dataImport.py
arjunrajlab: Removed in cd2e8cb7. The `recordable` wrapper passes the resource instance positionally (`self.findDatasetIdFn(*args, **kwargs)`), but the function never used it, and the `"DataImport"` annotation misleadingly implied a method — it is now `def getDatasetIdFromImportBody(*args, **kwargs)`.

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/dataImport.py
arjunrajlab: Keeping it, intentionally. The value is not that the request is repeated — it is that `@recordable` turns the whole import into a single undo-history entry. A bulk import can create thousands of annotations/connections/property values, and one-step undo is exactly what you want when someone imports the wrong file or into the wrong dataset. It is consistent with the other recordable batch endpoints, and there is a regression test (`testImportedDataIsRecordedAndUndoable`) confirming that `PUT /history/undo` reverts the entire import.

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/api/dataImport.py
arjunrajlab: Done in cd2e8cb7 — the four near-identical `isinstance` guards now iterate over `(name, value, expectedType, typeLabel)` tuples. Error messages are unchanged.

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/helpers/dataImport.py
arjunrajlab: By design, no — replace-on-import is an opt-in handled on the frontend rather than in this helper. When `overwriteAnnotations` is set, `importAnnotationsFromData` (`src/utils/annotationImport.ts`) snapshots the existing annotation ids and deletes them only **after** the backend import succeeds, so a failed import never destroys the user's existing data. The default import is additive, so this helper is intentionally create-only.

pchoisel:
LGTM

PR #1236, opened 2026-07-08
[queue-split 3/3, merge LAST — NOT backward compatible alone] Route worker jobs to cpu/gpu queues by isGPUWorker label

arjunrajlab:
Part **3 of 3** of the cpu/gpu Celery queue split (CytoPixel/AWSDeploy#83 has the infra + full runbook; arjunrajlaboratory/ImageAnalysisProject#153 has the labels). **Opened as a draft on purpose — this is the only PR of the three that changes behavior, and it must merge last.** Production topology and cutover specifics live in the private deployment repo's `doc/CPU_GPU_Queue_Split.md`, not here.

## What this PR does

Backend-only (Girder plugin — no frontend code):

- New `server/helpers/workerQueues.py`: `getQueueForImage(image)` reads the image's `isGPUWorker` Docker label from the docker daemon Girder already uses for tool discovery. `true/1/yes` → `gpu`; `false/0/no` → `cpu`; missing/garbage/docker-error → `gpu` fail-safe + warning log. Only definitive label reads are cached, so a relabeled + re-pulled image takes effect without a Girder restart; a 5s client timeout keeps a wedged daemon from stalling the interactive dispatch path.
- `getQueueForRequest(image, requestType)`: **interface requests always route to the `cpu` queue** regardless of the worker's GPU-ness — they just emit the tool's parameter schema, no GPU compute — so opening a tool's parameter form never occupies the GPU fleet. Compute/preview route by the label as above.
- `runJobRequest` passes `queue=getQueueForRequest(image, requestType)` to `docker_run.apply_async` — all four dispatch call sites (compute, property, preview, interface) inherit routing.
- Local dev: the compose `worker` service now consumes `-Q celery,cpu,gpu`, so local development is self-contained regardless of label state.
- 21 unit tests for the routing helpers.

## ⚠️ Coupling / merge order — read before merging

**Not backward compatible on its own.** Once a backend runs this code, every job is published to the `cpu` or `gpu` queue — if nothing consumes those queues yet, jobs sit in RabbitMQ unconsumed (a stall, not a graceful fallback).

**HARD DEPENDENCY (interface routing):** interface jobs run the actual worker container with `--request interface` under `pull_image=False`, so whatever consumes the `cpu` queue **must have every worker image on disk** — otherwise GPU workers' interface calls (their parameter forms) land on a box that lacks the image and fail.

Merge only after the pre-steps in the deployment repo's runbook (CytoPixel/AWSDeploy#83, `doc/CPU_GPU_Queue_Split.md`) are complete: labels live on all worker images, the `cpu`-queue consumer up and stocked with every worker image, the `gpu` fleet and autoscaler switched over — then deploy the backend per that runbook.

## How to test locally (safe now)

`docker compose up` on this branch, run a CPU tool (e.g. blob metrics) and a GPU-class tool; both run on the dev worker (it consumes all three queues). Unlabeled local images log the fail-safe warning and still run. Opening any tool's parameter form dispatches its interface job to the `cpu` queue.

https://claude.ai/code/session_01VjwUGizxeWLUxYs4yGa3me


arjunrajlab:
@codex review

arjunrajlab:
@codex review

PR #1240, opened 2026-07-12
Auto-suggest tools once per collection, on first open (persisted)

arjunrajlab:
## Problem

The AI tool-suggestion flow fired on **every** `layers-ready` for any toolless collection. Its only cross-reload guard was an in-memory `seenConfigurationIds` set, which a page reload wipes — so **reopening or refreshing an established, toolless collection re-ran the full screenshot + LLM suggestion request every time**. This is both noisy (a "Suggested tools" panel pops up on every reload) and wasteful (a real `claude_suggest_tools` LLM call).

## Fix

Auto-suggest now fires **once per collection, the first time it is opened with no tools**, and remembers which collections it has already suggested for in a **persisted set** (localStorage via the existing `Persister`, one key). Reopening or reloading in any later session never re-triggers it.

The set is **capped at the 500 most-recently-suggested ids** (~14 KB) and evicts oldest, so it can't grow unbounded for power users — 500 is larger than any realistic single-user collection count, so eviction (which would let an evicted, still-toolless collection re-suggest) effectively never happens.

The existing `tools.length` and in-session `seenConfigurationIds` guards are unchanged, and the **manual (lightbulb) trigger still works on demand** for older collections.

### Why not gate on the collection's creation time?
An earlier revision of this PR gated on a 5-minute creation-age window. [Codex correctly flagged](https://github.com/arjunrajlaboratory/NimbusImage/pull/1240) that this **expires before the first open** in batch-upload flows: `NewDataset.vue` creates the collection while processing the *first* dataset but only navigates to the viewer (first `layers-ready`) after the *last* dataset finishes uploading + transcoding — easily >5 minutes for a multi-dataset batch. A persisted "already suggested" set doesn't use time at all, so it's immune to that delay.

### Why not the recent-datasets list / `lastViewed`?
- The recent-datasets list is **capped at 20** (`MAX_NUMBER_OF_RECENT_DATASET_VIEWS`) — a recency window, not a complete "have I opened this" history.
- `lastViewed` is captured but **overwritten on every open**, and the `dataset_view` is created *during* the open flow, so it can't distinguish first open from later opens without a new backend field.

**Known tradeoff:** the persisted set is per-browser (not cross-device). First personal open of a shared toolless collection will suggest once — intended.

## Changes
- `toolSuggestions.ts` — localStorage-backed suggested-id helpers (`getSuggestedConfigurationIds` / `rememberSuggestedConfigurationId`, capped + evict-oldest); guard skips collections already in the persisted set; record on successful completion.
- `toolSuggestions.test.ts` — persisted-set skip, persist-on-success, no-persist-on-error, and cap/evict-oldest tests.

## Verification
- **In-browser (first open):** toolless collection not yet persisted → `POST claude_suggest_tools` (200), `status: done`, id written to the persisted set.
- **In-browser (reopen/reload):** same collection → **no request**, `status: idle`, still toolless — suppressed by the persisted guard.
- `pnpm tsc` clean; full `pnpm test` green (2293 tests, 135 files); eslint clean on changed files.

### DIFF
--- a/src/store/GirderAPI.ts
+++ b/src/store/GirderAPI.ts
@@ -1258,6 +1258,7 @@ export function setBaseCollectionValues(
     id: item._id,
     name: item.name,
     description: item.description,
+    created: item.created,
   };
   for (const key of configurationBaseKeys) {
     config[key] =
--- a/src/store/GirderAPI.ts
+++ b/src/store/GirderAPI.ts
@@ -1258,7 +1258,6 @@ export function setBaseCollectionValues(
     id: item._id,
     name: item.name,
     description: item.description,
-    created: item.created,
   };
   for (const key of configurationBaseKeys) {
     config[key] =
--- a/src/store/model.ts
+++ b/src/store/model.ts
@@ -519,6 +519,11 @@ export interface IDatasetConfiguration extends IDatasetConfigurationBase {
   readonly id: string;
   readonly name: string;
   readonly description: string;
+  // ISO creation timestamp from the underlying Girder document. Surfaced so
+  // consumers can tell a freshly created collection from an existing one being
+  // reopened (see toolSuggestions auto-suggest gating). Optional because not
+  // every construction path has the raw document in hand.
+  readonly created?: string;
 }
 
 export interface IDatasetViewBase {
--- a/src/store/model.ts
+++ b/src/store/model.ts
@@ -519,11 +519,6 @@ export interface IDatasetConfiguration extends IDatasetConfigurationBase {
   readonly id: string;
   readonly name: string;
   readonly description: string;
-  // ISO creation timestamp from the underlying Girder document. Surfaced so
-  // consumers can tell a freshly created collection from an existing one being
-  // reopened (see toolSuggestions auto-suggest gating). Optional because not
-  // every construction path has the raw document in hand.
-  readonly created?: string;
 }
 
 export interface IDatasetViewBase {
--- a/src/store/toolSuggestions.test.ts
+++ b/src/store/toolSuggestions.test.ts
@@ -328,7 +328,7 @@ describe("toolSuggestions store", () => {
     });
 
     it("caps the persisted suggested set and evicts the oldest ids", async () => {
-      // Seed one more than the cap; the mutation should keep the newest 500.
+      // Seed one more than the cap; remembering another keeps the newest 500.
       const seeded = Array.from({ length: 501 }, (_, i) => `cfg-${i}`);
       localStorage.setItem(
         "toolSuggestions.suggestedConfigIds",
--- a/src/store/toolSuggestions.test.ts
+++ b/src/store/toolSuggestions.test.ts
@@ -120,6 +120,9 @@ function makeConfiguration(
     id: "config-1",
     name: "Config",
     description: "",
+    // Default to a just-created collection so the recency-window guard treats
+    // configs as fresh unless a test overrides `created` with an older value.
+    created: new Date().toISOString(),
     compatibility: {
       xyDimensions: "one",
       zDimensions: "one",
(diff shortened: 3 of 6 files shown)


arjunrajlab:
**Good catch — fixed by dropping the age-based gate entirely.**

Confirmed the batch-upload race: `NewDataset.vue` calls `createCollection()` while processing the *first* dataset (`NewDataset.vue:1116`), but `navigateToCollection()` — which opens the viewer and triggers the first `layers-ready` → suggestion — only runs after the *last* dataset finishes uploading + transcoding (`NewDataset.vue:913`). For a multi-dataset batch that gap easily exceeds five minutes, so the creation-age window would suppress suggestions on the genuine first open.

**Fix:** replaced the `created`-age window with a persisted "already suggested" set (localStorage via `Persister`, one key). Auto-suggest now fires the first time a toolless collection is opened and records its id on completion, so reopening/reloading in any later session never re-triggers it. Because it doesn't use time at all, it's immune to the creation→first-open delay however long processing takes. The set is capped at the 500 most-recent ids (evict-oldest) so it can't grow unbounded. The now-unneeded `created` plumbing was reverted, so the net diff is self-contained in `toolSuggestions.ts`.

**Pattern audit:** the underlying issue is gating a user action on document *creation* time when the action can be delayed arbitrarily from creation. Swept the branch — this was the only such gate; no other instances remain.

Verified in-browser: first open of a toolless collection fires `claude_suggest_tools` and persists its id; reload no longer fires (`status` stays `idle`) while the collection is still toolless. Full suite green (2293), tsc + eslint clean.

@codex review

arjunrajlab:
**Correct — fixed by not running/persisting until templates are loaded.**

Confirmed: `toolTemplateList` is populated by `App.vue`'s async `fetchConfig()` (`config/templates.json`), which races the viewer's first `layers-ready`. If templates lose that race, `buildToolConfiguration()` returns `null` for every suggestion, the run still resolves to `status: "done"` with zero tools, and the old code persisted the collection id — making the transient miss permanent.

**Fix:** `maybeSuggestForCurrentConfiguration` now bails when `main.toolTemplateList.length === 0`, *before* marking seen or persisting, so a later open (templates cached) retries. `templates.json` always defines the `create`/`segmentation` templates, so a non-empty list means they're available. (Chose "don't run until templates loaded" over "treat all-unresolvable as retryable" because it targets the actual precondition and avoids any re-ask loop when a suggestion is genuinely unresolvable with templates present.)

**Pattern audit:** the underlying issue is persisting permanent dedup state on a `done` that a startup race can reach with an unloaded dependency. Swept the flow — the only other async dependency, `workerImageList`, is already `await`ed and a missing one just shrinks the catalog (manual blob still resolves), so it can't produce a spurious empty-then-persisted run. This was the only instance.

Verified in-browser: with templates loaded (`toolTemplateList.length === 11`) a fresh toolless collection still fires `claude_suggest_tools` and persists; a unit test covers the empty-templates bail (neither marked nor persisted). Full suite green (2294), tsc + eslint clean.

@codex review

arjunrajlab:
**Valid — and my round-2 audit was too narrow, thanks for catching it.**

I'd dismissed the worker-catalog path as safe because it "can't produce an empty run" (the manual blob always resolves). But that missed the real issue you're pointing at: the run isn't *empty*, it's *incomplete* — computed against a catalog with no worker tools — and persisting it permanently blocks the fuller retry.

Confirmed the mechanism: `fetchWorkerImageList()` early-returns when `main.isLoggedIn` is false (`properties.ts:720`). The `claude_suggest_tools` endpoint is `@access.user`, so a truly anonymous view 401s and errors out (not persisted) — but the concrete case you describe, a direct `datasetView` load with a stored token, *authenticates* the request before `main.initialize()` flips `isLoggedIn`, so the run succeeds against a blob-only catalog and gets persisted, suppressing Cellpose/Piscis for that collection.

**Fix:** this is the same class as the templates race (round 2), so I folded both readiness preconditions into one gate in `maybeSuggestForCurrentConfiguration` — bail (without marking or persisting, so a later open retries) unless `main.isLoggedIn && main.toolTemplateList.length > 0`. Those are the only async inputs a *complete* run needs: a missing screenshot already errors out (not persisted), and dataset/layers are implied by `layers-ready`. Also consistent with the manual AI trigger, which is already gated on `isLoggedIn`.

Verified in-browser: logged in with templates loaded (`toolTemplateList.length === 11`), a fresh toolless collection still fires `claude_suggest_tools` and persists; unit tests now cover both the not-logged-in and no-templates bails (neither marked nor persisted). Full suite green (2295), tsc + eslint clean.

@codex review

arjunrajlab:
**Right — the guards deferred without a re-trigger. Fixed with a readiness watcher.**

Confirmed the gap: the readiness bails (login + templates) don't mark or persist, but `Viewer.vue`'s `handleLayersReady` is the only production trigger. If prerequisites lose the startup race with the first `layers-ready` (e.g. a direct `datasetView` load with a stored token), that trigger has already passed and nothing re-invokes it — so the first-open suggestion was silently skipped, not merely deferred.

**Fix (`Viewer.vue`):** track whether the current configuration's image has rendered (`layersHaveRendered`, reset on config change) and watch a `suggestPrerequisitesReady` computed (`isLoggedIn && toolTemplateList.length > 0`). When readiness flips true *after* the image already rendered, re-attempt. `maybeSuggestForCurrentConfiguration` is self-guarding and idempotent, so the extra call is a no-op once already suggested/persisted, and it never fires before there's a rendered image to screenshot.

**Verified in-browser** by staging the exact race: on an already-rendered toolless collection, clear the persisted set and toggle `toolTemplateList` empty→full — the watcher fires the flow, `claude_suggest_tools` POSTs (200), `status` reaches `done`, and the collection is persisted. Unit tests cover `retrySuggestWhenReady` (fires only when ready **and** rendered), `layersHaveRendered` set/reset, and the readiness computed. Full suite green (2300), tsc + eslint clean.

@codex review

Issue #1243, opened 2026-07-12
Orphaned annotation_property_values left behind after (bulk) annotation deletion


arjunrajlab:
## Summary

Deleting annotations can leave behind orphaned `annotation_property_values`
documents — computed property values whose annotation no longer exists. These
accumulate in the database and inflate any per-annotation property analysis.

## Evidence

Observed on a real dataset (`38_km gdnf 2_3`, dataset id
`69f4eb65aaba948c2d7b9b24`):

| Count | |
|---|---|
| Live annotations tagged `nucleus` | 2,618 |
| `annotation_property_values` docs carrying an `Area` value | **5,237** |
| …whose annotation still exists | 2,618 |
| …**orphaned** (no matching annotation) | **2,619** |

All `annotationId` fields are `ObjectId`-typed (no string/ObjectId mix in the
stored data itself). So ~2,619 property-value documents reference annotations
that have been deleted.

## Likely root cause

There *is* a cleanup handler, so this is a bug in it (or in a delete path that
bypasses it), not a missing feature:

- `server/models/propertyValues.py` → `annotationsRemovedEvent` binds to
  `model.upenn_annotation.removeStringIds` and runs
  `removeWithQuery({"annotationId": {"$in": annotationStringIds}})`.
- `server/models/annotation.py` triggers `removeStringIds` from two paths:
  - single delete → `annotationRemovedEvent` passes `[event.info["_id"]]`
    (an **ObjectId**) → the `$in` matches the ObjectId-typed `annotationId`
    → values are cleaned up correctly.
  - bulk delete → `multipleAnnotationsRemovedEvent` passes `event.info`
    (the list is literally named `annotationStringIds`) → if those are
    **string** ids, the `$in` against an ObjectId field matches nothing
    → the property values are left orphaned.

That single-vs-bulk asymmetry is consistent with the observed partial
orphaning. Worth confirming by tracing what type `deleteMultiple` /
`removeWithQuery` actually pass into the event, and whether any delete path
(e.g. `removeWithQuery`, dataset re-import) skips the event entirely.

## Impact

- Property analysis over-counts: stats, histograms, and plots that read
  `annotation_property_values` see values for deleted annotations. (The AI
  analysis panel now intersects with the live annotation set to avoid this
  client-side — see NimbusImage PR for the AI-panel data-analysis branch — but
  the underlying data is still wrong for any other consumer, e.g. export.)
- Storage bloat over a dataset's editing lifetime.
[...]

PR #1221, opened 2026-07-03
Add claude_analysis endpoint: agentic property-value analysis backend

arjunrajlab:
New POST /api/v1/claude_analysis endpoint in the girder-claude-chat plugin
runs a Claude (claude-sonnet-5) agent loop with dataset-bound tools over
computed annotation property values: statistics, histograms, samples,
tag/shape summaries, and Plotly plot builders (scatter, histogram, box).
Dataset READ access is checked once at the API boundary; tools never
receive a caller-chosen datasetId. Plot data arrays are assembled
server-side from MongoDB so raw values stay out of the model context.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01HrBgREFZ2T7zYzMZbHjh1p

PR #1244, opened 2026-07-12
Agentic data analysis in the AI panel (builds on #1231, supersedes #1221)

arjunrajlab:
## Agentic data analysis in the AI panel

Builds the agentic property-value analysis of #1221 **on top of the AI panel
(#1231)** instead of as a separate server-side agent loop. This PR is based on
the #1231 branch so the diff shows only the data-analysis work; it should be
retargeted to `master` (or merged after) once #1231 lands. **Supersedes #1221**
— that PR can be closed.

Full design + verification: `codebaseDocumentation/AI_PANEL_DATA_ANALYSIS_SPEC.md`.

### Approach
The AI-panel agent loop already runs in the browser and the frontend already
holds every property value in memory, so the analysis tools are **frontend
executors on the existing `/claude_agent` surface** — **no backend Python code
changes** (only the plugin's tool-definition JSON + system-prompt data files).
The #1221 invariant is preserved: plots build their Plotly traces client-side,
park them in a panel-local registry, and return only `{plotId, title, count}`
to the model, so raw per-annotation data never enters the model context.

### What's added
- **Six tools**: `get_property_values` extended with std/median/p25/p75, plus
  new `get_property_histogram`, `get_sample_values`, `create_scatter_plot`,
  `create_histogram_plot`, `create_box_plot` (all read-only, ungated).
- `src/agent/analysis.ts` (pure stat helpers) and `src/agent/plotRegistry.ts`
  (plot data lives outside Vuex; avoids a circular import).
- `kind:"plot"` transcript items rendered by `AiPanelPlot.vue`, which
  **lazy-loads `plotly.js-dist-min`** — the production build code-splits it into
  its own ~4.6 MB chunk, never the main bundle.
- Plots persist to IndexedDB (≤12, size-capped) and re-render across reload.

### Analysis reflects live annotations
`collectPathValues`/`get_sample_values` intersect with the live annotation set
rather than iterating raw `propertyValues` keys, so deleted annotations' values
don't inflate stats or add a spurious "untagged" group to tag-grouped plots.
The underlying backend cleanup gap is filed separately as #1243.

### Also in this PR
- Fixes a **pre-existing race** in `aiPanel.ts` (from #1231): clearing the
  conversation during the post-reload IndexedDB hydration could strand the
  `hydrating` send-guard and silently block all sends. Fixed with a dedicated
  `hydrationGeneration` counter + regression test. (This also repaired the
  `aiPanel.test.ts` conversationStore mock, which #1231 phase 3 had broken —
  only `src/agent` had been run in CI-equivalent checks.)

### Verification
- `pnpm tsc`, `pnpm lint:ci` clean; **189 `src/agent` + `src/store` tests pass**.
- Production build confirms Plotly is a separate lazy chunk.
- **Live-tested** against a real dataset (2,618 nuclei with computed intensity +
  morphology): histogram / scatter-by-tag / box-by-tag / query restriction all
  render inline; reported stats match MongoDB ground truth (Area mean 367.9,
  std 106.6, range 25–975); the raw-data invariant holds on the wire (plot
  tool_results ~23 KB, no value arrays); conversation + plots persist and
  re-render across reload; a mixed analysis→`set_annotation_filter` turn reverts
  cleanly; and the live-set fix drops the unrestricted scatter from 5,237 to
  the 2,618 real nuclei.

### DIFF
--- a/codebaseDocumentation/AI_PANEL_DATA_ANALYSIS_SPEC.md
+++ b/codebaseDocumentation/AI_PANEL_DATA_ANALYSIS_SPEC.md
@@ -13,18 +13,29 @@
 > `src/agent/executors.ts`, the tool schemas + "Analyzing data" prompt
 > section in the `girder-claude-chat` data files, `kind:"plot"` transcript
 > items with `AiPanelPlot.vue` (lazy Plotly), and IndexedDB plot
-> persistence. `pnpm tsc` / `lint:ci` clean; 125 `src/agent` tests pass.
+> persistence. `pnpm tsc` / `lint:ci` clean; 189 `src/agent` + `src/store`
+> tests pass.
 >
-> **Live-verified** against dataset `69f4eb65aaba948c2d7b9b24` (2,618 nuclei
-> with computed intensity + morphology properties): histogram, scatter
-> (colorByTag → per-tag WebGL traces), box (groupByTag), and query
-> restriction all render inline and read back correct stats (matched
-> MongoDB ground truth: Area n=5237, mean 367.9, std 106.6, range 25-975;
-> query→nucleus dropped n to 2618). Confirmed the raw-data invariant on the
-> wire (plot tool_results ~23 KB, no raw value arrays), conversation +
-> full plot data persisted and re-rendered across reload, a mixed
-> analysis→`set_annotation_filter` turn with working revert, and that the
-> production build code-splits Plotly into its own 4.6 MB lazy chunk.
+> **Follow-up fixes on this branch** (from live testing):
+> - **Live-set intersection** (§2, §5.1) — analysis excludes property values
+>   orphaned by deleted annotations; backend cleanup gap filed as
+>   [issue #1243](https://github.com/arjunrajlaboratory/NimbusImage/issues/1243).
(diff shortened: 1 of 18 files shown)


arjunrajlab:
Thanks — both addressed in `27782ed0`.

1. **Dataset re-check after the fetch**: added `assertDatasetUnchanged(context)` after `fetchPropertyValues()` in all six analysis executors (`get_property_values`, `get_property_histogram`, `get_sample_values`, `create_scatter_plot`, `create_histogram_plot`, `create_box_plot`) — the same post-await guard `run_worker` already uses — so a mid-fetch dataset switch aborts instead of producing a stat/plot for the wrong dataset.

2. **Box-plot downsampling**: no longer downsamples. At/below `MAX_BOX_POINTS` it passes the raw points (exact, with outliers); above it, it passes exact precomputed quartiles from the full data (`q1`/`median`/`q3` + min/max fences), so the box is always accurate rather than a silent every-kth approximation. Audited the other downsamplers on the branch: scatter already flags `(downsampled)`, histogram bins the full data, and `get_sample_values` is an explicit N-row sample.

Regression tests added for both (identity-abort throws; the >cap box carries exact quartiles from the full data with no raw `y`).

PR #1247, opened 2026-07-17
fix(annotations): restore click-to-navigate/highlight in the annotation list

arjunrajlab:
Clicking an annotation in the viewer stopped highlighting and scrolling to
its row in the annotation list. Two regressions from the stub-annotation
work combined to break it:

1. setHoveredAnnotationFromCoordinates resolved the clicked feature only via
   getAnnotationFromId, which returns undefined for unhydrated non-point
   stubs. In stub-only mode (large datasets) every stub-rendered annotation
   was therefore silently unclickable. Resolve to the stub and hit-test the
   rendered dot (shouldSelectStub), mirroring the point-click selection path.
   Connection features are now skipped explicitly.

2. The AnnotationList hover watch returned early for the entire body in
   server mode, so it never scrolled to the hovered row. Only the client-side
   page-jump depends on the client filtered set; guard just that computation
   and keep the scroll-to-row behavior in both modes.

Add regression tests covering click-to-hover for hydrated point annotations,
stub-rendered annotations, empty-space clears, and the tool-selected no-op.

## Stub-mode page navigation (anchorId)

Fixing (2) restored the scroll-to-row, but in server/stub mode the list only
holds the current page of ~10 rows, so clicking an annotation that lives on
another page flashed and then showed the first page again — the browser
cannot know the annotation's page from what it holds. That position depends
on the filtered, sorted order computed in MongoDB, so it is resolved
server-side:

- New optional `anchorId` on `POST /upenn_annotation/list`. When supplied,
  the server locates the annotation under the same filters/sort and returns
  the page-aligned `offset` (null when the annotation is filtered out).
- `Annotation.listPosition` reuses the exact match/lookup/sort builders as
  `listPage`, so its ordering cannot drift from the list. The `_id`/default
  sort uses an indexed range count; non-`_id` sorts fetch the anchor's
  normalized sort value then count rows sorting strictly before it with a
  single `$expr` match (no full-set sort or window walk — bounded on 700K+
  datasets).
- Frontend `annotationListServer.fetchPageContaining` jumps to the returned
  page; a navigation sequence guard lets a later hover cancel an in-flight
  lookup, and a pre-navigation options snapshot lets a genuine user
  pagination click cancel the navigation while Vuetify's stale options echo
  is dropped.

## Verification

- Backend: flake8 clean; 305 backend tests (new anchor + self-consistency
  cases covering desc sorts, `_id` ties, missing `name` values, property
  sorts with/without values, and property filters).
- Frontend: `pnpm tsc`, `pnpm lint:ci`, and 2651 vitest tests green.
- Live against a 708,983-annotation dataset in stub mode: a real annotation
  click navigated the list to the containing page with the row highlighted
  and scrolled into view; normal pagination and hover-clear behave correctly.

See `codebaseDocumentation/ANNOTATION-LIST-CLICK-NAV-REVIEW.md` for the
review-finding tracker and live-verification notes.

### DIFF
--- a/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
+++ b/devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
@@ -366,10 +366,7 @@ class Annotation(AccessControlMixin, ProxiedModel):
             cursor = self._aggregate(self._pvModel.collection, pipeline)
             return [str(doc["annotationId"]) for doc in cursor]
 
-        pipeline = self._buildListMatchStages(datasetId, filters)
-        if filters.get("propertyFilters"):
-            pipeline += self._lookupStages()
-            pipeline += self._propertyFilterStages(filters)
+        pipeline = self._annotationDrivenStages(datasetId, filters)
         pipeline.append({"$project": {"_id": 1}})
         cursor = self._aggregate(self.collection, pipeline)
         return [str(doc["_id"]) for doc in cursor]
@@ -627,10 +624,7 @@ class Annotation(AccessControlMixin, ProxiedModel):
             result = list(self._aggregate(self._pvModel.collection, pipeline))
             return result[0]["n"] if result else 0
 
-        pipeline = self._buildListMatchStages(datasetId, filters)
-        if filters.get("propertyFilters"):
-            pipeline += self._lookupStages()
-            pipeline += self._propertyFilterStages(filters)
+        pipeline = self._annotationDrivenStages(datasetId, filters)
         pipeline.append({"$count": "n"})
         result = list(self._aggregate(self.collection, pipeline))
         return result[0]["n"] if result else 0
@@ -644,12 +638,7 @@ class Annotation(AccessControlMixin, ProxiedModel):
         transfer of the full matching id set to the browser.
         """
         def basePipeline():
-            pipeline = self._buildListMatchStages(datasetId, filters)
-            if self._needsPropertyBeforePage(filters, sort):
-                pipeline += self._lookupStages()
-                pipeline += self._propertyFilterStages(filters)
-                pipeline += self._propertySortAddFields(sort)
-            return pipeline
+            return self._annotationDrivenStages(datasetId, filters, sort)
 
         # The default list order is _id ascending, and explicit _id sorting
         # is also common. Resolve that position with an indexed range count
@@ -687,7 +676,8 @@ class Annotation(AccessControlMixin, ProxiedModel):
         # replicates the page order (_hasSortValue desc, value asc/desc, _id
         # asc) as an expression. $ifNull maps a missing field to null exactly
         # like $sort does, so missing and null order identically here and in
-        # listPage.
+        # listPage. `sort` is never None here: the isIdSort branch above
+        # returns for a missing sort.
         descending = sort.get("order") == "desc"
         if sort.get("type") == "property":
             valueRef = {"$ifNull": ["$_sortValue", None]}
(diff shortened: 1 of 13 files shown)


src/components/AnnotationViewer.vue
```diff
@@ -2961,27 +2961,37 @@ function handleInteractionModeChange(evt: any) {
...
+          [gcsCoordinates],
+          candidate,
+          geoAnnotation.style(),
+          unitsPerPixel,
+        )
+      : shouldSelectStub(
+          gcsCoordinates,
+          candidate,
+          geoAnnotation.style(),
+          unitsPerPixel,
```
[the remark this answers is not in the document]
arjunrajlab: Fixed in 48643a02. `shouldSelectStub` no longer routes through `pointNearPoint` (which multiplies the radius by `unitsPerPixel`). Since the stub dot renders world-locked and its `style.radius` is `estimatedRadius` in world units, the hit-test now compares distance directly against the radius in world units and converts only the display-pixel stroke width by `unitsPerPixel`. This also corrects the drag/click-select path that shares the helper. Added regression tests that fail against the old scaling.

---

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
```diff
@@ -635,6 +635,118 @@ def listCount(self, datasetId, filters):
...
 
+    def listPosition(self, datasetId, filters, sort, annotationId):
+        """Zero-based position of an annotation in a filtered/sorted list.
+
+        This intentionally uses the annotation-driven form for every query.
+        It has the same filter and ordering semantics as listPage, including
+        property filters/sorts and the stable _id tie-break, while avoiding
+        transfer of the full matching id set to the browser.
+        """
+        def basePipeline():
```
pchoisel, line 640: Can this function be used in other places in this file ?

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
```diff
@@ -635,6 +635,118 @@ def listCount(self, datasetId, filters):
...
+
+        # Non-_id sorts: fetch the anchor's sort value inside the filtered
+        # set, then COUNT the rows that sort strictly before it -- the same
+        # match/lookup stages, but no full-set $sort or window walk. "Before"
+        # replicates the page order (_hasSortValue desc, value asc/desc, _id
+        # asc) as an expression. $ifNull maps a missing field to null exactly
+        # like $sort does, so missing and null order identically here and in
+        # listPage.
+        descending = sort.get("order") == "desc"
+        if sort.get("type") == "property":
```
pchoisel, line 682: Can't `sort` be `None` ?

pchoisel (commented):
LGTM otherwise

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
arjunrajlab: Extracted it as `_annotationDrivenStages` in eae91bf4 and reused it in `listIds`, `listCount`, `listPosition`'s `basePipeline`, and `listPage` — so the match + conditional property-join stages aren't duplicated across four call sites anymore.

devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/server/models/annotation.py
arjunrajlab: No — `sort` is always populated by this point: the `isIdSort` check earlier in the function (`not sort or (sort.type == "field" and sort.key == "_id")`) returns before we get here whenever sort is missing, so every code path reaching this line has a non-`_id` sort dict. Added a comment in eae91bf4 to make that explicit.

Issue #1258, opened 2026-07-22
Expose an "active users" / usage metric (e.g. daily active users) for admins


arjunrajlab:
### Problem / motivation

There's currently no straightforward way to answer a basic operational question: **how many distinct users used the service today (or over a given period)?**

The AWS infrastructure telemetry we have (CloudWatch, CloudTrail, EC2/ASG) is purely infrastructure-level — it shows machine load and autoscaling behavior but has **zero per-user information**. The only ways to get a distinct-user count today are indirect and require privileged access:

- **Girder's MongoDB** (authoritative): count distinct `userId` in the `token` collection created within a window (or a `lastLogin` field if one exists). But the Atlas cluster is peering-only, so this needs in-VPC access (SSH to a backend + query through the container) or an Atlas Admin API key.
- **HAProxy / nginx access logs**: distinct client IPs as a rough proxy — inaccurate (NAT / shared IPs) and requires SSHing the load balancer.
- **Outseta**: external dashboard, not integrated with the app.

None of these is self-serve, and none is both accurate and convenient.

### Proposed feature

Expose a lightweight usage metric that an admin can read **without SSH or direct DB access**. Minimum viable version:

- **Active users today** — distinct authenticated users in the last 24h — surfaced as either:
  - an admin-only REST endpoint (e.g. `GET /api/v1/system/active_users?window=1d`), and/or
  - a panel on an admin dashboard.

Nice-to-haves (later):
- Configurable window (today / 7d / 30d).
- A simple daily distinct-user series for trend visibility.
- Optional breakdown by activity type (login vs. active API use vs. job submission).

### Implementation sketch

Girder already persists auth tokens with `userId` + `created`, so a distinct-count aggregation over a time window is cheap and yields an accurate "authenticated users in window" number. If a stricter definition of *active* is wanted (users who actually did something, not merely held a session), key it off recent activity — e.g. a lightweight `lastSeen` touch on requests or job submissions.

### Acceptance criteria

- An admin can retrieve the count of distinct users active in a given window without SSH / DB access.
- The count is deduplicated per user (not request or IP counts).
- The endpoint and the definition of "active" are documented.

### Context

Filed from an ops session (2026-07-21) while assessing infrastructure usage. The infra metrics showed platform load clearly, but there was no way to tie that to a distinct-user count — hence this request. Infra-side investigation lives in the AWSDeploy operational notes.

PR #1260, opened 2026-07-22
Add admin-only authenticated-users usage metric endpoint

arjunrajlab:
Related to #1258 (does not fully resolve it — see "Scope & what this does not do" below).

## Summary

Exposes a lightweight usage metric so a site admin can answer "how many distinct users authenticated in the last N?" **without SSH or direct database access**.

Adds `GET /api/v1/system/authenticated_users?window=1d`, registered on Girder's core `system` route via `addSystemEndpoints` in `system.py` (alongside the existing `item/query` and `folder/query` additions). It counts **distinct** users who obtained an authentication token within the look-back window, by grouping on `userId` over the `token` collection.

## What this measures

A user is counted for the window if the `token` collection holds a token whose `created` timestamp falls inside it — i.e. they authenticated (logged in / refreshed a session) during the window. Girder persists tokens with `userId` + `created`, so a distinct-count aggregation is cheap and accurate. The count is deduplicated per user (`$group` on `userId`), so it is a user count, not a request or token count.

Because Girder deletes tokens once they expire (default ~180 days), counts for windows longer than the token lifetime are bounded by token retention. This is documented on the endpoint (`.notes()`) and in `CLAUDE.md`.

## Scope & what this does *not* do

The endpoint was originally named `active_users` / `activeUsers`, which overclaimed. It has been renamed to **`authenticated_users` / `authenticatedUsers`** to describe exactly what it measures.

It is **not** a general "used the app" activity metric. The frontend persists the auth token in `localStorage` and reuses it across reloads, minting a new token only on a **fresh login** (or after the token expires). So a user who logged in three weeks ago and uses the app every day generates no new token and will **not** appear in a short window. Read this as **distinct authentications in the window**, not daily-active usage.

Because of that gap, this PR is intentionally **not** marked as closing #1258 — a fuller "active = used the app" definition (e.g. a `lastSeen` touched on request/job activity) remains open there as a follow-up.

## Details

- `@access.admin` — gated to site administrators (anonymous → 401, non-admin → 403).
- `window` accepts an integer + unit (`s`/`m`/`h`/`d`/`w`), defaults to `1d`, and is capped at `366d` to bound the scanned token range. It is normalized once at the API boundary (strip/lower) so the echoed `window` matches the parsed `windowSeconds`. Malformed values return a clean `400`.
- Uses `collection.aggregate` — the sanctioned exception to the "no raw PyMongo" rule, since Girder's `find()` doesn't support aggregation pipelines.
- Response: `{ window, windowSeconds, start, end, authenticatedUsers }`.

## Acceptance criteria

- [x] An admin can retrieve the count of distinct authenticated users in a window without SSH / DB access.
- [x] The count is deduplicated per user (not request or IP counts).
- [x] The endpoint name and its definition/limitations are documented (Swagger `.notes()` + `CLAUDE.md`).

## Testing

Verified against the merged-up-to-date branch with the plugin rebuilt into the running Girder container:

- **Backend pytest** (`tox`): `test/test_authenticated_users.py` — **11 passing**. Covers distinct-user dedup, window boundary inclusion/exclusion, the `1d` default, normalized-echo of whitespace/case in `window`, admin gating (403), unauthenticated rejection (401), and invalid-window rejection (400).
- **Live HTTP** against `http://localhost:8080`: all window units (`s`/`m`/`h`/`d`/`w`), default window, access control (403 non-admin, 401 anonymous/bogus token), and the full invalid-window matrix (`abc`, `0d`, `-1d`, `5x`, `9999d`, `367d`, …) → all correct.
- **Correctness cross-check**: endpoint count matched an independent MongoDB aggregation over the `token` collection; anonymous (`userId: null`) tokens correctly excluded.
- **Dedup / window boundary** proven with controlled data: 5 tokens for one new user raised the distinct count by exactly 1; backdating those tokens to 100 days ago dropped the user out of a 90d window and back in at 200d.
- `flake8` clean on the changed backend files.

## Follow-ups (out of scope)

- A stricter `lastSeen`-based definition (touch on request/job activity) so users on a still-valid token are counted — this is what would let the endpoint answer #1258's "active users" intent in full.
- An admin-dashboard panel surfacing this metric.
- A daily distinct-user time series for trend visibility.

https://claude.ai/code/session_01N92Zx6tbPbCFczVMQJEmXR

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,8 +155,11 @@ Plugin endpoints are registered in `__init__.py` (lines 159-173). Endpoint names
 | `/api/v1/annotation_import` | `server/api/dataImport.py` | Server-side JSON import (annotations, connections, property values) |
 | `/api/v1/project` | `server/api/project.py` | Project management |
 | `/api/v1/resource` | `server/api/resource.py` | Custom resource search |
+| `/api/v1/system/active_users` | `system.py` | Admin-only usage metric: count of distinct users who obtained an auth token within a look-back `window` (default `1d`). Added to Girder's core `system` route via `addSystemEndpoints`, not a plugin resource. |
 
 All source files under `devops/girder/plugins/AnnotationPlugin/upenncontrast_annotation/`. Swagger UI at `http://localhost:8080/api/v1#!/`.
+
+**Active-users metric — "active" definition:** a user counts as active in the window if the `token` collection holds a token whose `created` timestamp falls inside it, i.e. they authenticated (logged in / refreshed a session) during the window. The count is deduplicated per user (`$group` on `userId`). Because Girder deletes tokens once they expire (default ~180 days), counts for windows longer than the token lifetime are bounded by token retention.
 
 ### Image Rendering Pipeline
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,11 +155,11 @@ Plugin endpoints are registered in `__init__.py` (lines 159-173). Endpoint names
 | `/api/v1/annotation_import` | `server/api/dataImport.py` | Server-side JSON import (annotations, connections, property values) |
 | `/api/v1/project` | `server/api/project.py` | Project management |
 | `/api/v1/resource` | `server/api/resource.py` | Custom resource search |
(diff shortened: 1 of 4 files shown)


pchoisel:
LGTM

PR #1263, opened 2026-07-22
Raise full-fetch caps to 100M and paginate large dataset fetches

arjunrajlab:
## Summary

Large datasets (millions of annotations) hit inadvertent ceilings on the full-fetch paths that silently truncated results rather than erroring. This was surfaced by a dataset with ~2.7M annotations. This PR raises the relevant caps to 100M and switches the worker-client "fetch all" methods to cursor pagination so they can never silently truncate.

## Changes

- **Worker client `getAnnotationsByDatasetId`** (`annotation_client/annotations.py`): replaced the hardcoded `limit=1_000_000` single-request default with `_id` cursor pagination that walks the entire dataset (100k/page, up to a 100M runaway guard). An explicit `limit` still fetches a single bounded page. Also fixes a `&offset{offset}` typo that emitted `offset0`, silently ignoring the offset.
- **Worker client `getPropertyValuesForDataset`**: paginate the same way. It previously passed no limit and so returned only the endpoint's default first page for the whole dataset.
- **Frontend `fetchAllPages`** (`src/utils/fetch.ts`): raise the cursor safety limit from 100 to 1000 requests, lifting the effective ceiling from 10M to 100M rows (page size unchanged at 100k). This path already paginates; only the ceiling changed.
- **Backend validation** (`server/helpers/validation.py`): raise the `MAX_ANNOTATION_IDS` and `MAX_UNCOMPUTED_PROPERTIES` request-size guards from 10M to 100M.

## Deliberately not changed

- `MAX_LIST_LIMIT = 10_000` — a per-page abuse guard on the public `/list` endpoint, not a total-fetch cap. The list view pages properly, so it never truncates a full load; raising it would let a single request serialize 100M full rows (the OOM risk it exists to prevent).
- `get_annotation_list_by_shape` default `limit=50` and `getAnnotationConnections` default `limit=50` — search helpers whose callers pass explicit limits, not silent-truncation caps on a full fetch.

## Notes

The "loading forever" symptom in the browser was not itself a 1M truncation — that path already capped at 10M (now 100M) and streams all stubs for large datasets. The genuine 1M truncation was on the worker/compute path (e.g. property computation), which is fixed here.

## Testing

- `flake8` clean on the changed Python files.
- No existing tests pin the changed cap values (backend tests monkeypatch these constants to small values).

---
_Generated by [Claude Code](https://claude.ai/code/session_019ymoRPZ9wMUAQNGztfwPjv)_


pchoisel:
LGTM

PR #1266, opened 2026-07-23
Persist annotation-browser property selection in configuration

arjunrajlab:
## Problem

The property columns shown in `AnnotationList` and the property filter rows in the annotation browser (including each filter's range and enabled state) were only held in memory. Refreshing the page or switching to another dataset and back discarded them — the user had to re-select columns and re-add filters every time.

## Approach

Persist this interface state in the **configuration metadata**, mirroring the existing `visibilityConfig` pattern (frontend-only; configuration metadata round-trips generically and the collection JSON schema does not reject unknown keys, so **no backend change** is required).

Rather than a global watcher, the three store methods that change this state — `togglePropertyPathVisibility`, `togglePropertyPathFiltering`, `updatePropertyFilter` — become actions that mutate as before and then schedule a debounced save. This automatically covers every existing call site (PropertyPicker, AnnotationList, PropertyCreation, PropertyFilterHistogram, and the chat agent) without touching any of them, and it avoids saving on the resets/prunes/unmount-writes that a deep watcher would have to filter out.

## Changes

- **`model.ts`** — `IAnnotationBrowserConfig` (displayed columns, filter paths, property filters) added as an optional key on `IDatasetConfigurationBase`; `resolveAnnotationBrowserConfig()` drops malformed entries and paths referencing properties no longer attached to the configuration.
- **`properties.ts` / `filters.ts`** — toggle/update mutations split into private `*Impl` mutations wrapped by persisting actions; `hydrate*` actions restore state via raw mutations (never scheduling a save); `updateDisplayedFromComputedProperties` no longer prunes hydrated columns before any property values have loaded.
- **`index.ts`** — debounced `scheduleAnnotationBrowserSave` (500 ms, so dragging a filter slider doesn't spam the backend), `flushAnnotationBrowserSave`, and `hydrateAnnotationBrowserState`, wired into `setConfiguration` and `setSelectedDataset`. A "hydrated for this configuration" guard blocks saves during transitions — notably the disabled filter that `PropertyFilterHistogram` re-adds on unmount, which would otherwise write into the wrong configuration.

## Testing

- `pnpm tsc`, `pnpm lint:ci`, and `pnpm test` (172 files / 2862 tests) all green. New `annotationBrowserConfig.test.ts` covers save-scheduling, hydration-does-not-save, malformed-data resilience, and the prune guard.
- **In-browser (HCR dataset):** selected a column + set a filter range → confirmed the debounced save wrote the full shape to the backend config; full reload restored all three; switched datasets → no cross-dataset bleed; reloaded HCR fresh → state intact (switch-away did not corrupt its config). No console errors; test data cleaned up afterward.

https://claude.ai/code/session_01N92Zx6tbPbCFczVMQJEmXR


arjunrajlab:
Both P2 findings were real — fixed in a50d6523 (and follow-up hardening in d7d763b1).

**1. Disabled filter rows re-enabled on hydration** (`filters.ts` / `PropertyFilterHistogram.vue`)
Confirmed: `PropertyFilterHistogram`'s `onMounted` force-enabled any disabled filter, so a persisted `enabled: false` was flipped back on the mount that follows hydration. Fixed by moving the "re-adding a removed row reactivates it" behavior into the `togglePropertyPathFiltering` action (fires only on add) and removing the `onMounted` force-enable. Hydration no longer goes through that path, so a deliberately-disabled filter now survives.
Verified live: disabled the Area filter, reloaded, opened the filter panel so the histogram component actually mounts — checkbox stays unchecked and the store keeps `enabled: false`; separately confirmed remove-then-re-add still comes back enabled.

**2. Pending save dropped on same-dataset config switch** (`index.ts`)
Confirmed: the debounced save was only flushed in `setSelectedDataset`; a configuration switch within the 500 ms window cleared the timer (via `hydrateAnnotationBrowserState`) without saving. Fixed by flushing *before* the configuration flips — flushing inside hydrate would be too late, since `this.configuration` already points at the new config there. Both transition entry points now go through a single `beginAnnotationBrowserTransition` (flush old + block saves) so the invariant holds regardless of the order the two concurrent actions resolve in.

**Branch-wide audit of both patterns:**
- *Component lifecycle clobbering hydrated state:* the `onMounted` force-enable was the only instance — `AnnotationList`/`AnnotationViewer`/`AnnotationCSVDialog` only read this state on mount; `PropertyCreation` writes it via a user action, not a mount hook.
- *Debounced config-write droppable across transitions:* `scheduleAnnotationBrowserSave` was the only one, now flushed at both transition entry points. (The pre-existing `@Debounce` `datasetView` writers are the same shape but out of scope for this PR.)

A self-review also caught that `saveAnnotationBrowserConfig` persisted disabled orphan filters (paths no longer in `filterPaths`), which would accumulate unboundedly across add/remove cycles — now it persists only filters backing a visible row (extracted to a unit-tested `buildAnnotationBrowserConfig`, with the load side dropping orphans too). Two lower-priority review findings were filed as #1267 (read-only viewers hitting failing config saves — systemic to `syncConfiguration`) and #1268 (splitting up `index.ts`). `tsc`/`lint`/`test` (2868) green.

@codex review

PR #1275, opened 2026-07-25
Show an active-filter count badge on the Filters button

arjunrajlab:
## What

Filters keep narrowing the visible object set after the Filters palette is closed, and nothing in the app bar says so — users can be left wondering why objects are missing. This adds a count badge to the app-bar Filters button.

## Changes

**`src/store/filters.ts`** — new `activeFilterCount` getter. It counts:

- each enabled tag / property / region / object-id filter
- the selection filter when enabled
- the two boolean toggles when set away from their permissive default: `onlyCurrentFrame` on, and `showAnnotationsFromHiddenLayers` off (it defaults to `true`)

`emptyROIFilter` — a region still being drawn — filters nothing yet and is excluded.

**`src/App.vue`** — a `.palette-ibtn-badge` span inside the Filters button, shown only when the count is non-zero and capped at `9+` so it stays inside the 32px icon button. The tooltip gains the count too (`… (3 active filters)`), and the `aria-label` becomes `Filters (3 active)` — it stays terse to match the sibling palette buttons, but carries the count because `aria-label` overrides the button's rendered content and would otherwise hide the badge from assistive tech.

The badge's separating ring and the palette cluster's background now share a `$palette-cluster-tone` SCSS variable instead of each hardcoding the same value.

## Testing

- `src/store/__tests__/activeFilterCount.test.ts` (new, 11 cases) covers each filter kind, disabled-but-populated filters, a region mid-draw, mixed sums, and the reset-on-dataset-switch path.
- `src/App.test.ts` covers the tooltip/aria-label strings and asserts the badge markup against real rendered output — this needed pass-through stubs for `v-app` / `v-app-bar` / `v-tooltip`, since `shallowMount` stubs the root `v-app` and App.vue's own app-bar markup never renders otherwise.
- `pnpm test` 2964 passed (178 files), `pnpm tsc` clean, `pnpm lint:ci` clean, `pnpm build` succeeds and the compiled CSS was checked to confirm the SCSS variable resolves to the same tone as before.

Not verified in a live browser: this environment has no Docker, so no Girder backend or dataset to load the app bar against. The badge markup and CSS are covered by the render test and a bundle-output check instead.

https://claude.ai/code/session_01JDmkc7um3o911eUYyXgzgj

---
_Generated by [Claude Code](https://claude.ai/code/session_01JDmkc7um3o911eUYyXgzgj)_

### DIFF
--- a/src/App.test.ts
+++ b/src/App.test.ts
@@ -28,6 +28,12 @@ vi.mock("@/store/properties", () => ({
   },
 }));
 
+vi.mock("@/store/filters", () => ({
+  default: {
+    activeFilterCount: 0,
+  },
+}));
+
 vi.mock("axios", () => ({
   default: {
     get: vi
@@ -52,6 +58,7 @@ import { routeProvider, routerProvider } from "@/test/helpers";
 import App from "./App.vue";
 import store from "@/store";
 import propertyStore from "@/store/properties";
+import filterStore from "@/store/filters";
 import axios from "axios";
 import { logError } from "@/utils/log";
 
@@ -64,7 +71,10 @@ const mockRouter = {
   push: vi.fn(),
 };
 
-function mountComponent(routeOverrides: Record<string, any> = {}) {
+function mountComponent(
+  routeOverrides: Record<string, any> = {},
+  extraStubs: Record<string, any> = {},
+) {
   return shallowMount(App, {
     global: {
       mocks: {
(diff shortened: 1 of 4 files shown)


arjunrajlab:
@codex review

(No automated review arrived when this PR was opened via the API, so triggering one manually.)

---

PR #1276, opened 2026-07-25
Label unrolled frames and make the labels navigate (#151)

### LINKED ISSUE #151: Unroll navigation features
It would be nice if the unroll would have labels on the individual unrolled frames, and that clicking on the labels would navigate back to that particular frame without unroll. For instance, if you have a bunch of XY coordinates, then unroll could show "XY=1", "XY=2", … in the upper left corner. If you were to click that label, then it would go back to the un-unrolled (i.e., rolled) view, but with that XY coordinate. So if you clicked XY=4, then it would go back to XY=4.

arjunrajlab:
Closes #151.

## What

When a dimension is unrolled, each cell of the grid now carries a label in its upper-left corner, and the label is a control: clicking it rolls the grid back up at that frame.

- **Label text** mirrors the navigator sliders' 1-based display, and includes the dataset's dimension label when it has one: `XY 3 (19263, -6626)`, or plain `XY 3` / `Z 2` / `Time 7` without one. With several dimensions unrolled at once it joins them — `XY 3 · Z 2`. A dimension with a single value is left out of the label (no `XY 3 · Time 1` on a one-timepoint dataset) but is still pinned on navigation. The parenthetical is gated per axis on the existing **Show XY / Z / Time labels** viewer settings — the same switches the navigator sliders use — so one toggle controls both places.
- **Clicking a label** sets `xy`/`z`/`time` to that cell's frame for the unrolled dimensions, leaves the others alone, then clears the unroll flags. Clearing the flags is what triggers the dataset reload — `NavigatorPanel` watches all three, the same path snapshot restore (`Snapshots.vue`) and the AI panel (`agent/executors.ts`) already rely on. Calling `refreshDataset()` here as well would load the dataset twice.
- Labels are `<button>` elements, so they work from the keyboard as well as the mouse. Only the label is clickable; the rest of each cell behaves exactly as before, so annotation tools still work in unroll mode.
- Labels exist only while unrolled — no new setting, no new persisted state.

## How

`src/utils/unroll.ts` (new, pure, unit-tested) turns the grid's frames into `{ index, label, location }` per cell. Two subtleties it encapsulates:

- `store.xy`/`.z`/`.time` are **indices into `dataset.xy`/`.z`/`.time`**, which `parseTiles` builds as the sorted distinct axis values. So the index to navigate to is a frame's *rank* among those values, not its raw `IndexXY`/`IndexZ` (`z` can fall back to a `PositionZ` float, and nothing guarantees dense indices). Ranking also keeps the label (`rank + 1`) consistent with where the slider lands.
- Those ranks have to come from **every frame of the dataset** (`dataset.allImages`), not from the drawn layer's frames. A channel imaged only at XY 0 and 5, in a dataset whose frames cover XY 0, 2 and 5, would otherwise label its second cell `XY 2` and navigate to XY 2 — where that channel has no image.

`ImageViewer.vue` renders each label as a GeoJS `dom` widget on the map's ui layer, positioned with map coordinates so GeoJS tracks it through pan, zoom and container resize (its `zoom()` ends in a `pan()` and `size()` ends in a `center()`, both firing the `geo_event.pan` that widgets bind). A `domWidget` stops `mousedown` from reaching the GeoJS interactor, which is why a label click cannot start an annotation. Cells are derived **per map** from that map's own `mapLayerList` group, so the multi-map `layerMode: "unroll"` view labels each grid from its own layer. Widgets are rebuilt only when a signature (labels + grid width + cell size) changes, and a watcher on the cells covers the inputs that change label text without causing a redraw (the viewer settings). Grids past 400 cells go unlabelled — one repositioned element per cell stops being reasonable and the text is unreadable at that density — and that decision is logged rather than silent.

## Review rounds

Two Codex rounds plus a `/branch-review`, tracked in `codebaseDocumentation/UNROLL-FRAME-LABELS-REVIEW.md`:

| Finding | Fix |
|---|---|
| Ranks from one layer's frames, and one layer's cells reused for every map (Medium) | `axisImages` (`dataset.allImages`) separate from `cellImages`; cells derived per map |
| Clickable `<div>` had no keyboard access or button semantics (Low) | `<button type="button">` with `aria-label` and a focus ring, via GeoJS's `el` option |
| Label settings didn't update existing widgets until the next redraw (Low) | `watch(unrollCellsByMap, …)` rebuilds them |
| `/branch-review`: impure computed (logged and mutated state in its getter), test identified widgets by call order, one naming slip | Cap and warning moved to the draw path; test picks labels out by class; name restored |

Two pre-existing bugs in `ImageViewer.test.ts` surfaced while testing the watcher and are fixed here: the file never unmounted its wrappers, so every earlier test's component stayed alive and reacted to the shared `reactive()` store mock (one store change ran ~90 components' watchers); and the label assertions read the widget *creation log* rather than the widgets a layer still has, which `draw()`'s map-recreation in the harness made misleading.

## Verification

`pnpm tsc`, `pnpm lint:ci`, and the full suite (178 files / 2973 tests) are green. New tests: 16 unit tests for the ranking and label rules (sparse coverage, dataset-wide ranks, `PositionZ` fallback, grid order, multi-axis, single-valued dimensions, dimension labels on/off/missing), and 11 `ImageViewer` tests for widget creation and positions, the click → navigate path, button semantics, per-map cells, the settings watcher, no-unroll, signature reuse, teardown, and the 400-cell cap.

Driven in the browser on the `normmedia_8well_col2_livecellgfp` dataset (6 XY × 7 Z × 4 T), with real clicks and keystrokes:

| Case | Result |
|---|---|
| Unroll XY | 6 labels `XY 1 (19263, -6626)` … at the cell corners of the 3×2 grid |
| Click `XY 5` | `unrollXY=false`, `xy=4`, `z`/`time` untouched, labels gone, slider reads the 5th XY's dimension label |
| Unroll Z + Time | 28 labels `Z 1 · Time 1` … `Z 7 · Time 4`, 6 per row, matching the tile layout |
| Click `Z 1 · Time 2` | both flags cleared, `z=0`, `time=1`, `xy` preserved |
| Focus `XY 4` + Enter | same navigation as a click — keyboard path works |
| `layerMode: "unroll"` + XY unroll | 12 labels across 2 maps, each from its own layer; clicking the second map's `XY 6` navigates |
| Toggle **Show XY labels** off/on while unrolled | labels rewrite to `XY 1` and back to `XY 1 (19263, -6626)` with no redraw |
| Click a label with the "YFP blob" tool armed | navigated; no annotation created, no pending annotation |

No console errors (only pre-existing Vuetify `dense` deprecation warnings). Note for anyone reproducing with screenshots: the tiles are WebGL canvases that come out black in captured screenshots even though the page renders them — `map.screenshot()` compositing confirms the pixels are there.

https://claude.ai/code/session_01XzqX5knspQjXKBJGUPVMpo

### DIFF
--- a/.agents/skills/in-browser-testing/SKILL.md
+++ b/.agents/skills/in-browser-testing/SKILL.md
@@ -9,12 +9,18 @@ description: "Use when verifying NimbusImage changes in a live browser tab (loca
 
 Multiple real bugs in this repo passed tsc, lint, all unit tests, and code reasoning — and only broke in the live app (pointer-events layering, deep-watch over-firing, HMR-corrupted store, background-tab false-passes). For any user-facing change, in-browser verification is a required gate, and it must be done in a way that can actually fail. The traps below each produced a false "it works" at least once.
 
-## The four false-pass traps
+## The five false-pass traps
 
 1. **Synthetic clicks lie.** `element.dispatchEvent(new MouseEvent('click'))` bypasses hit-testing, `pointer-events`, z-index, and overlays. A synthetic-click walkthrough once reported a tour "working" while real user clicks did nothing. Verify clickability with `document.elementFromPoint(cx, cy)` (is the intended element the topmost hit target?), then click with the real `computer` tool. Reserve synthetic dispatch for *driving* flows fast only after clickability is independently confirmed.
 2. **Background tabs lie.** A hidden tab pauses `requestAnimationFrame`, throttles `setTimeout` to ~1 Hz, and never composites — perf probes report 0 long tasks and event loops stall/time out. The tab must be foreground/visible for any timing or rendering claim. If a paced loop must survive a background tab, use busy-wait gaps (`while (performance.now() < end) {}`), not setTimeout.
 3. **A stale store lies.** Editing any `src/store/*.ts` while `pnpm run dev` runs breaks HMR (`[vuex] duplicate getter key`, annotations stuck at 0). Hard-reload the page after every store-module edit before trusting anything. (Component `.vue` edits HMR fine.)
 4. **A stale backend lies.** After backend plugin edits, `docker compose restart girder` does NOT load the new code (plugin is baked into the image) — new routes 404 while tox passes. Rebuild: `docker compose build girder && docker compose up -d girder`.
+5. **Screenshots lie about the image itself.** The tile layers are WebGL canvases without `preserveDrawingBuffer`, so a captured screenshot can show the image area **solid black** while the page renders it normally for the user — DOM overlays in the same shot look fine, which makes it read as "the image failed to load". Don't debug that; confirm pixels through GeoJS's own compositing instead:
+   ```js
+   const dataUrl = await map.screenshot(undefined, 'image/png');  // re-renders into a 2D canvas
+   // draw to an offscreen canvas, then getImageData → report max/mean channel value
+   ```
+   A non-zero max means the tiles are there. Note that a fluorescence dataset is *legitimately* near-black at fit-to-view zoom (sparse bright cells, `hist.max` pinned to 65535 by a few saturated pixels), so a low mean is not evidence of a bug either. Verify tile health from `statusCode`s in `read_network_requests` (`/tiles/zxy/` → 200) plus the layers' `idle`/`_imageUrls`, and verify overlays from live DOM state — not from the picture.
 
 ## Console handles (javascript_tool)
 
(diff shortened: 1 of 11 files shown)


arjunrajlab:
@codex review

Areas worth a close look: the cell → store-index mapping in `src/utils/unroll.ts` (ranks among sorted distinct axis values rather than raw `IndexXY`/`IndexZ`, because `store.xy`/`.z`/`.time` index into `dataset.xy`/`.z`/`.time`), and the widget lifecycle in `ImageViewer.vue` (`updateUnrollLabels` / `clearUnrollLabels`, keyed per map in a `WeakMap`).

PR #1279, opened 2026-07-25
Connection list view with selectable connections (#554)

### LINKED ISSUE #554: Probably need some sort of list view for connections
I think it would be good to have a list view for connections, just to see what is connected to what and to allow for precise deletion of unwanted connections.

Further along, it might be nice to do property calculations on connections as well.

arjunrajlab:
Closes #554.

Connections were write-mostly: tools created them, the viewer drew them, and the only removal path was a bulk-delete dialog operating on whole categories. There was no way to see what is connected to what, or to delete a single link.

## What this adds

- **A Connections tab in the Object Browser** — flat or grouped by track, with a scope selector (all / current location / selected objects / passing filters). A connection is in scope when *either* endpoint qualifies, matching the existing bulk-delete dialog.
- **Deletion** per-row, bulk, and per-track — all single batched requests.
- **Connect selected** — chains the objects selected in the Objects tab into connections by ascending time.
- **Clickable connection lines in the viewer**, in both normal and timelapse mode. Clicking a track segment selects exactly that link, so a bad track can be cut where you see it.

Frontend only — every endpoint already existed and is `@recordable`, so undo/redo works unchanged.

## Two decisions worth pinning down

**`parentId` is always the earlier annotation.** This matches what `createTimelapseConnection` already normalizes to; inverting it would produce links backwards relative to all existing data.

**Within a frame the data carries no direction**, so ties fall back to selection order. That is only meaningful when objects are clicked one at a time — under drag-select it is effectively arbitrary — so the toolbar says so rather than hiding it.

## A pre-existing bug this fixes

`drawNewConnections` null-checked its endpoints via `getAnnotationFromId`, which returns `undefined` for every unhydrated non-point annotation in stub-only mode. The function only ever used the endpoints' ids to look up centroids, so that check gated drawing on hydration for no reason.

On the 709K-object Xenium dataset this meant **only 1 of 11 connection lines was drawn** even though all 12 endpoints were on screen and all 12 centroids were present. Zooming in hydrated the rest and the lines appeared, which made it read as a zoom bug. This affects any lazily-loaded dataset with connections, independently of this feature.

## Incidental cleanups

- `findConnectedComponents` moves to `src/utils/connections.ts`, shared by the list and the timelapse draw path.
- The `"Time lapse connection"` magic string (three call sites) becomes `TIMELAPSE_CONNECTION_TAG`.
- Stub-aware navigation moves to `src/utils/annotationNavigation.ts` so both tabs use it.
- Connection hit-testing deliberately does *not* reuse `pointNearLine`, which compares a squared distance against an unsquared width — correct enough for its existing callers, but it would make lines unclickable when zoomed out.

## Verification

`tsc` clean, `lint:ci` zero warnings, **3,058 tests pass** (+63). Both review-round regressions and the stub-mode draw bug were each confirmed to fail without their fix.

Browser-verified on three datasets, since green tests do not mean the UI works:

| Check | Result |
|---|---|
| Click a connection line (normal + timelapse mode) | selects that link, highlights cyan |
| Annotations win / lasso never grabs lines / empty space clears | 2-0, 48-0, cleared |
| Connect selected → delete round trip | 54 → 55 → 54, confirmed in MongoDB |
| Xenium (709K objects, stub-only): fresh load | 11/11 lines with 4/12 endpoints hydrated |
| Xenium: zoom sweep 0 → 6 → 0, budget 18,040 ↔ 5,000 | 11/11 at every step, no drift |
| 4,983 connections: row / track build time | 6.7 ms / 6.1 ms |

Two guards came out of the self-review: **Connect selected is capped** at 500 (a select-all would otherwise POST tens of thousands of connections in one request — the backend has no cap either), and **changing the scope clears the connection selection** (otherwise selecting all 4,983, narrowing to a 3-row scope, and pressing delete removes connections the user cannot see).

## Note on real data

The list immediately surfaced genuine orphaned data: the timelapse dataset `38_km gdnf 2_3` has **2,616 of 5,230 connection endpoints pointing at annotations that no longer exist** (verified identically from the client and MongoDB). Those render as `⚠ missing` and stay deletable. Expect them to be common in older datasets — they are not a rendering bug.

Design notes and the full verification tables are in `codebaseDocumentation/CONNECTION_LIST.md`.

https://claude.ai/code/session_01XzqX5knspQjXKBJGUPVMpo


arjunrajlab:
Thanks — all three were legitimate. Fixed in 470d4fd8. In each case I generalized the finding and audited the branch for the same pattern rather than patching only the flagged line.

### P1 — Clear selections when scoped inputs change

You're right, and this was a real gap in my own earlier fix: clearing in `setScope` only covers an explicit scope *switch*, not a change to the **inputs** a dynamic scope reads. Scrubbing XY/Z/Time under `location`, changing the object selection under `selected`, or editing filters under `filtered` all silently replace the visible rows with the stale selection still armed.

Rather than chase every input with watchers, I made the invariant structural. The list's bulk delete now derives from a new `selectedInScopeConnectionIds` getter — the intersection of the selection with the rows currently in scope — so *"you can only bulk-delete rows the list is showing"* holds by construction regardless of which input moved. The button's count shows the same number, so it never promises more than it will do.

One distinction worth flagging, since intersecting *globally* would have been wrong: the viewer's action panel keeps using the **raw** selection (`deleteSelectedConnections` vs. the new `deleteSelectedInScopeConnections`). Clicking a connection line on the canvas can legitimately select something outside the list's current scope, and deleting the link you just clicked is the intent there.

Verified live on the 709K-object Xenium dataset in exactly your scenario — with scope `selected`, shrinking the object selection leaves the raw selection at 11 while Delete selected correctly acts on 1.

### P2 — Reset saving state after failed batch creation

Correct. Wrapped in `try/finally`. Worth noting *why* this one bit: the action is `rawError: true`, so a rejection genuinely propagates instead of being swallowed. Audited the branch — this was the only `setSaving` pair it introduced. The other 13 sites in `annotation.ts` are pre-existing bare `@Action`s whose API helpers catch and log, so they don't propagate; I left them alone rather than widen the diff.

### P2 — Reveal viewer-selected connections in the list

Good catch, and this was a documentation overclaim rather than a code bug — I wrote that behavior into `CONNECTION_LIST.md` and never implemented it.

Implemented rather than weakened: `revealConnection` pages the flat list to the row (or expands the containing track in track mode), then scrolls to it via a `data-connection-id` attribute, triggered whenever the selection becomes a single connection — which is what a viewer click produces.

It deliberately does **not** open the Object Browser or activate the tab, which your comment also mentioned: a click on the canvas shouldn't throw a palette over the image. The doc now states that limit explicitly instead of implying full sync.

### Verification

`tsc` clean, `lint:ci` zero warnings, 3,064 tests pass. All three P1 scope-input cases and the reveal behavior have tests that were each confirmed to fail without their fix.

@codex review

arjunrajlab:
Round 2: all five confirmed against the code and reproduced on the live 709K-object Xenium dataset before fixing. Pushed in 05054dd5.

### Retain stub-backed connection lines using their centroids

Correct, and this one was my miss: `clearOldAnnotations` carried the *identical* `getAnnotationFromId` coupling I had just removed from `drawNewConnections`. I fixed one half of a pair and didn't audit for the other. Retention now uses the same displayed-id + centroid criteria as the draw.

Reproduced from a fresh load at 4/12 endpoints hydrated: `clearOldAnnotations` removed **10 of 11** lines, which the next draw rebuilt. Worth recording that my *first* probe showed no churn at all — a false negative, because by the time I panned, every endpoint had hydrated and the predicate passed. Only a fresh load exposes it.

### Apply selected styling when constructing connection lines

Confirmed. `restyleAnnotations` only runs on selection/hover *changes*, so a selected line torn down and rebuilt came back default-blue and stayed. Now styled via `getConnectionStyle` at construction. Verified: after wiping the layer and redrawing, the selected line comes back `#00e5ff` immediately.

### Draw or reject same-time links created by the chain action

Confirmed — `otherAnnotation.location.Time >= annotation.location.Time` skips an equal-time link from *both* endpoints, so it is never drawn.

Of your two options I took "draw it once" rather than "prevent creation", because same-frame chaining is a deliberate product decision here (direction is unknowable within a frame, so it falls back to selection order and the UI says so). Rejecting the link would contradict that. Ties now break on id so exactly one traversal draws the segment.

### Prune selections for every connection deletion path

Correct — `DeleteConnections.vue::submit` and `deleteAllTimelapseConnections` both delete straight through the annotation store.

Rather than react to each deletion path, I derived existence: `selectedExistingConnectionIds` intersects the selection with the live connection collection, and the viewer panel's visibility, its count, and its delete all key off it. That stays correct no matter who deletes, and needs no watcher.

### Reveal the selection when the tab becomes visible

Confirmed on both counts — the component mounts lazily on first activation, and `scrollIntoView` on a hidden `VWindowItem` row is a no-op. The tab now passes `is-active` and the reveal retries when it becomes visible, covering the never-opened case via an `immediate` watch.

### Verification

`tsc` clean, `lint:ci` zero warnings, 3,068 tests. Each of the five has a test confirmed to fail without its fix. Live re-check on Xenium at 4/12 endpoints hydrated: 11/11 lines survive `clearOldAnnotations`, a rebuilt selected line is cyan, and a nonexistent selected id is excluded from the effective selection.

@codex review

arjunrajlab:
Round 3: both confirmed and fixed in 03ca2d72. The first was a bug my own round-2 fix introduced.

### Skip connection features in the annotation restyle pass

Correct, and the mechanism is exactly as described. Connection lines carry a `girderId`, so they land in `drawnGeoJSAnnotations`; the retained-feature loop treats every entry as an object annotation; connections never set `isHovered`/`isSelected`, and `undefined != false` is true, so the branch fired and replaced the cyan with `getAnnotationStyle(connectionId, …)`. The construction-time styling I added in round 2 was therefore undone by the very next redraw. The loop now `continue`s on `isConnection`.

Reverting the guard turns the regression test's expected `#00e5ff` into the layer's `red`, so it fails for the right reason.

### Preserve per-record selection for duplicate timelapse links

Confirmed, and not hypothetical — the HCR dataset in this repo has **21 duplicate pairs** across 54 connections.

I took the "choose the selected duplicate as the rendered representative" option rather than keying `drawnLines` by connection id, since the latter would stack N exactly-overlapping segments and still leave clicks resolving to whichever came first. One segment per pair is retained; a selected duplicate wins the representative slot.

### On verification

Rather than test against whatever data happened to exist, I built a fixture on a genuinely 4-timepoint dataset: a 4-frame track, a same-frame pair, and an off-location pair. Live result: 4 segments, 3/3 track links, the same-frame link drawn, no duplicate segments, and the `location` scope correctly excluding the off-location link.

That exercise paid for itself twice:

- It showed the round-2 equal-time fix **not** working live while its unit test passed. Cause was not the code — an interrupted revert-verify-restore command had silently reverted that hunk in my working tree. Restored, and re-verified.
- It exposed that the round-2 equal-time change never had a regression test. Added, along with one for self-connections (zero-length segments, correctly not drawn).

An earlier probe of yours-would-have-been-round-4 territory also turned out to be a false negative: my first "is there churn?" check passed only because everything had hydrated by the time it ran. Stub-only behavior has to be reproduced from a fresh load.

### Verification

`tsc` clean, `lint:ci` zero warnings, 3,072 tests. Each fix has a test confirmed to fail without it.

@codex review

arjunrajlab:
Round 4 addressed in 4c0be794. Both were real, and the P1 was the most dangerous finding of this cycle.

### Ignore object-delete shortcut modifiers

Confirmed and fixed. `mod+backspace` is bound to "delete selected objects" (`ImageViewer.vue:546`), and the panel's handler matched bare `Backspace` without inspecting modifiers. The state you describe is not an edge case — clicking a connection row selects both endpoint objects *and* the connection — so ⌘+Backspace deleted two annotations and a connection from one keystroke. The handler now ignores meta/ctrl/alt.

`ConnectionActionPanel` had no tests at all; it now has five, including this regression (confirmed to fail without the guard), the typing-in-a-field guard, existence-based counting, and listener teardown on unmount.

### Apply hover styling to timelapse segments

Confirmed. Worth noting the waste as well as the inconsistency: the selection/hover watcher rebuilds the *entire* timelapse layer when `hoveredConnectionId` changes, so before this the redraw happened and produced no visible difference. Segments now honour hover.

Verified live in real timelapse mode on a purpose-built 4-timepoint dataset: segment width 3 base, 5 hovered, 6 selected in cyan, `stroke: true` throughout.

### Two fixes you have not seen yet

This round targeted `03ca2d72`, so it predates `5e4ff9ea`, which fixed a bug reported from the Xenium dataset — clicking a connection in the list while zoomed in showed nothing. Two independent causes:

1. **The construction-time styling from round 3 made every connection line invisible at every zoom.** `options("style", …)` replaces rather than merges, so `{strokeColor, strokeWidth, strokeOpacity}` dropped GeoJS's default `stroke: true`. The lines were present in `layer.annotations()`, correctly positioned, the right colour in `options().style`, and never painted. The round-3 test asserted colour only; it now asserts `style.stroke` too.

2. **Navigating to a connection recentered on one endpoint.** A connection draws only when *both* endpoints are displayed, so at max zoom on Xenium — viewport 51 world units, endpoints 139 apart — the line was skipped. Added `frameCameraInfo` (recenter, zoom out only as far as needed, rotation preserved) and `goToConnection`, which frames both endpoints when they share a frame and degrades to a plain navigate when they don't, since endpoints on different timepoints can never both be displayed in normal mode.

Both are worth a look in this pass.

### Verification

`tsc` clean, `lint:ci` zero warnings, 3,082 tests. The `sync` CI job is also green again — the skill edits in this branch had updated only `.claude/skills/` and not the `.agents/skills/` mirror; that requirement was documented nowhere, so `CLAUDE.md` now states it.

@codex review

arjunrajlab:
Round 5 triaged: one fixed here, one filed as a follow-up.

### Compute framing scale in the rotated camera basis — fixed in b6443c19

Correct, and not theoretical: `ImageViewer` applies `cameraInfo.rotate` through `map.rotation()`, so a rotated viewport is reachable. `frameCameraInfo` now projects the required span onto the two viewport edge directions and compares against those edge lengths, which reduces to the previous formula at zero rotation.

The test uses a 45-degree diamond viewport and pins both sides of the boundary: a `(14, 14)` span whose endpoints fall outside the diamond must zoom out — the axis-aligned formula compared 14 against a 20×20 bounding box and wrongly concluded it fit — while a 20-wide horizontal span, which reaches exactly the left/right vertices, must **not**. Confirmed to fail against the old math.

### Frame connections using unrolled coordinates — filed as #1280 (bug, high priority)

You're right, and I verified it with numbers rather than by inspection. On a 4-timepoint dataset with `unrollT` on, tile width 1024:

- the annotation's raw centroid is `x = 260`
- it actually renders on the unrolled grid at `x = 1284`
- navigation centres the camera on `x = 260` — **off by exactly one tile**

I also confirmed your related point: with unrolling active, a Time 0 → Time 1 connection **is** drawn, as a line between two grid tiles. So the `sameFrame` guard's premise ("endpoints on different frames can never both be displayed") is false while unrolling, and should become "both endpoints displayed" once unrolled coordinates are reachable.

Not fixed in this PR because it is **pre-existing rather than a regression**: the same measurement against `goToAnnotationLocation` — the Objects tab path that shipped long before this branch — also centred on `x = 260`. Clicking any annotation row while unrolled already pans to the wrong tile.

Fixing it properly is not local. `unrolledCoordinates` depends on `props.unrollW`, a ref in `ImageViewer.vue` passed down as a prop, so it is unreachable from a util. It is derivable, so the clean fix is to expose the unroll grid on the store, extract the offset into a shared util, and route both `goToAnnotationLocation` and `goToConnection` through it — which changes Objects-tab behaviour and deserves its own review. That's all captured in #1280 with the measurements and a suggested approach.

### State

`tsc` clean, `lint:ci` zero warnings, 3,083 tests, and both CI jobs green — the earlier `sync` failure was skill edits that updated only `.claude/skills/` and not the `.agents/skills/` mirror, a requirement that was documented nowhere and is now in `CLAUDE.md`.

@codex review

arjunrajlab:
Round 6 addressed in 1f69d2d4. All four were real; the first is one my own test had already caught and I wrongly dismissed.

### Mount a single connection action panel across maps

Confirmed and fixed. `ImageViewer` renders one `AnnotationViewer` per layer group in unroll mode, so the panel appeared N times and registered N global keydown listeners — one Delete press fired N concurrent deletes with duplicate batch DELETEs for the same ids. The panel reads only store state, so it now mounts **once** in `ImageViewer`, outside the per-map loop.

Worth recording how this shipped: when I wrote the panel's tests, the delete spy fired twice and I attributed it to a leaked mount between tests. It *was* that — but the same leak is reachable in production whenever more than one viewer mounts, and I didn't ask that second question. Both skills now say so explicitly.

The mount also needed a `key`: an unkeyed conditional sibling next to a keyed `v-for` tripped Vue's `patchKeyedChildren`, which surfaced during user testing as `Cannot read properties of null (reading 'emitsOptions')`.

### Preserve API failures from Connect selected

Correct. `AnnotationsAPI.createMultipleConnections` catches HTTP errors and returns `null`, which becomes `[]`, so a permissions or server failure was reported as successful deduplication.

Rather than change the API layer's error contract, dedupe is now determined from the chain being empty **before** the request — `chainAnnotationsByTime` returning no bases is the only thing that actually means "nothing to do". An empty result with a non-empty chain is now reported as a failure. Both branches have tests.

### Retain the endpoint vector sign when framing

Correct, and your worked example is now the test: on a 20×10 viewport at 45°, `(-23, 11)` needs ~2.40× while `(23, 11)` needs ~1.20×.

One correction to the framing of the finding: `frameCameraInfo` already projected whatever vector it was handed, so the defect was purely in the **caller** passing `Math.abs()` of each delta. That means a camera-level test cannot catch it — I confirmed the new rotated-viewport test still passes against the old code — so the regression guard is a `goToConnection` test asserting the signed delta reaches `frameCameraInfo`.

### Prefer the hovered duplicate as the track representative

Correct — the representative preferred a *selected* duplicate but not a *hovered* one, so hovering a later duplicate's row triggered a full layer rebuild that neither widened the segment nor retagged it. Now selected, then hovered, then first.

### Also from user testing

A plain click highlighted a connection but never brought its row into view: reveal was wired to selection changes only, and a plain click sets hover. It now reveals on hover too, which is safe because a row hovered *in the list* is already on the current page, so both the paging and the scroll are no-ops there.

### Verification

`tsc` clean, `lint:ci` zero warnings, 3,092 tests, both CI jobs green. Verified live on the 4,983-connection dataset: exactly one action panel in the DOM across panel/tab churn, no console errors, and hovering a connection at row index 140 pages the list to page 3 and scrolls to it.

@codex review

arjunrajlab:
Round 7 addressed in adcde8a0, along with a systematic pass at the recurrence problem.

### Resolve the closest connection hit instead of the first

Confirmed and fixed. `findConnectionIdAtPoint` returned the first line within tolerance, so with parallel or dense tracks a click could select a far segment over a near one and some links were unreachable from the canvas entirely. `pointNearConnectionLine` became `connectionLineHitDistance`, returning the squared distance instead of a boolean, and the caller takes the minimum. Layer order still decides *between* layers, because in timelapse mode the track lines are the ones the user can see.

### On the recurrence pattern

Several findings in this review have been regressions introduced while fixing an earlier one — construction-time styling that the next redraw undid, a retention predicate keeping a coupling the draw path had just dropped, a per-viewer mount resurfacing as duplicate DELETEs. That is a testing gap, not bad luck, so I audited every fix in the branch for a corresponding test and added the six that had none:

| Gap | Test |
|---|---|
| closest-hit resolution | *selects the closest connection when several are within tolerance* |
| panel must not mount per viewer | *does not mount the connection action panel per viewer* |
| hovered duplicate as representative, **and** hover changing the rendered width | *renders the hovered duplicate as the representative* |
| reveal on hover, selection winning | *reveals on hover…*, *prefers the selection over the hover…* |
| dedupe vs. request failure | *flags dedupe only when the chain was empty before the request* |
| saving state on a rejected create | *clears the saving state when the create request rejects* |

Each was confirmed to fail without its fix.

`codebaseDocumentation/CONNECTION_LIST.md` now carries a **Regression checklist** — 21 invariants grouped by drawing / interaction / destructive actions, each naming the test that holds it, so touching this code means re-checking a list rather than rediscovering it. It also records the two process rules that would have prevented most of the churn: verify from a *fresh page load* on a dataset that actually has the property under test, and use `git stash` rather than a `cp` round-trip when checking that a test fails without its fix (an interrupt between revert and restore silently dropped a fix earlier in this branch, and the live check disagreeing with a passing unit test is what exposed it).

### Verification

`tsc` clean, `lint:ci` zero warnings, **3,099 tests**, skill parity green.

@codex review

arjunrajlab:
Round 8 addressed in 23cd672a, plus the timelapse issues from user testing in d230ca19.

### Ignore deleted selections when revealing connections

Correct, and a good catch — it is an interaction between two earlier fixes in this branch rather than either being wrong alone. Selection deliberately retains ids for connections deleted through other paths (round 6: existence is *derived*, not pruned), and reveal was given hover as a trigger with selection taking priority. Together, one externally deleted selected connection left `revealCurrentSelection` permanently taking the selection branch, calling `revealConnection` with an id no longer in the list, and returning early — so every later plain click highlighted a row the list would never page to.

Priority now reads `selectedExistingConnectionIds`. Written test-first and confirmed to fail without the change (*"ignores a deleted selection when revealing on hover"*).

### Two timelapse issues from user testing (d230ca19)

Both reported on the 4,983-connection timelapse dataset; regular mode was unaffected, which turned out to be the clue.

**Clicking a track segment did nothing.** Objects win a click so a line crossing an object cannot steal it — but in timelapse mode the track segments *are* the visual and the annotation-layer dots sit underneath, so a segment almost always crosses one and the connection never got hovered. Confirmed live: `findConnectionIdAtPoint` returned the right id while the handler resolved to an object. Timelapse now prefers the connection and falls back to the object; normal mode is unchanged, and both directions are pinned by tests.

**Hover was slow.** The watcher rebuilt the whole timelapse layer — one line feature per connection, 5,217 on that dataset — and hover changes continuously as the pointer moves down the list. Rebuilds are now selection-only: measured 0.02 ms per hover change afterwards, with no rebuild. This reverses your round-4 request in the other direction; it was right while the rebuild happened anyway, but not worth this cost. A hovered track segment no longer widens until something else redraws.

### On recurrence

Several findings in this review were regressions introduced while fixing earlier ones, so beyond per-fix tests the branch now carries a **Regression checklist** in `codebaseDocumentation/CONNECTION_LIST.md` — 27 invariants grouped by drawing / interaction / cost / destructive actions, each naming the test that holds it.

Getting the timelapse precedence test to fail *correctly* took three attempts, each defeated by a shared mock returning a fixed value (`pointDistance` is a bare `vi.fn()` returning `undefined`, so an object can never be hit by default). That trap is now recorded in the `nimbus-geojs` and `nimbus-frontend` skills, since a test that passes before the fix is worse than no test.

### Not from this branch

Two console warnings surfaced during testing that this branch does not touch: the Vuetify `dense` deprecation (`PipelineList`, `PipelineEditor`, `PipelineStepEditor`, `PropertyCreation`, `SharingStatusIcon` — 16 occurrences) and a `<Transition>` non-element-root warning from `LargeImageDropdown`. Split out to #1281 rather than widening this PR.

### Verification

`tsc` clean, `lint:ci` zero warnings, 3,107 tests, both CI jobs green.

@codex review

arjunrajlab:
Round 9 addressed in fa7e584d. Both were real.

### Prefer track connections during timelapse selection

Correct, and the more useful framing is that I applied a rule to one of two paths. The timelapse inversion went into `setHoveredAnnotationFromCoordinates` (plain-click highlighting) and not into `selectAnnotations`, which is what shift+click and the select tool use — so a track segment overlapping a dot could be highlighted but not selected, which is most of them.

Both paths now share the rule: connections win when a connection is hit and either nothing else was hit **or** timelapse mode is on; objects keep winning outside timelapse. Tests pin both directions on both paths.

Worth noting this is the third instance of the same shape in this branch — draw vs. retention gating on different criteria, construction styling vs. the redraw loop, and now hover vs. selection precedence. The checklist in `codebaseDocumentation/CONNECTION_LIST.md` now has an explicit item that both click paths must share every precedence rule, since "fixed it in the path I was looking at" is clearly my recurring failure mode here.

### Gate scoped counts while the tab is hidden

Correct — my visibility gate was incomplete. Gating the row getters left `scopedCount` and `selectedInScopeCount` still consuming `scopedConnections`, and `scopeAnnotationIds` scans `annotationsForIteration` for the dynamic scopes, so a hidden tab kept doing a full-dataset scan on every XY/Z/Time scrub. Both counts are gated on `isActive` now, with a test asserting the scope getters are not read while hidden.

### Verification

Both written test-first and confirmed to fail without their fix. `tsc` clean, `lint:ci` zero warnings, 3,110 tests, both CI jobs green.

I also re-verified that nothing earlier in the branch has regressed — each of the 27 prior fixes checked individually against the tree rather than assumed, after an interrupted revert silently dropped one earlier in this review cycle.

@codex review

arjunrajlab:
Round 10 addressed in 1f265bf4, plus a branch-wide sweep for the two shapes this review keeps surfacing (6433141b).

### Avoid scanning every annotation for location scope

Correct. Scoping is now a **per-connection predicate** (`connectionInScope`) rather than a set of qualifying annotation ids: connections are the far smaller collection, so it resolves their two endpoints instead of enumerating every object. That removes both the `annotationsForIteration` array materialization and the full scan from every XY/Z/Time scrub. The filtered scope reuses the filters module's existing `filteredAnnotationIdToIdx` rather than rebuilding a set of its own.

The test leaves `annotationsForIteration` **empty** while keeping the endpoints resolvable, so any implementation that scans it finds nothing and fails — it cannot pass for the wrong reason.

### Short-circuit oversized selections before resolving them

Correct, and a fair characterisation: the cap was materializing exactly the work it exists to prevent. `canConnectSelected` and `connectSelectedExceedsMax` now check the raw `selectedAnnotationIds.size` first and only resolve within the cap; `connectSelectedTimeTies` returns `[]` above it rather than traversing a selection Connect selected has already refused. The test counts resolve calls and asserts zero for an oversized selection.

### Branch-wide sweep for both shapes

Ten rounds have produced roughly twenty findings that reduce to **two shapes**, so rather than wait for the next instance I swept the branch for both and found one more of each:

- **A rule applied to one of two symmetric paths** (four prior instances: draw vs retention gating, construction styling vs the redraw loop, hover vs selection precedence). The sweep found a fifth: `deleteConnectionsById` pruned `selectedConnectionIds` and left `hoveredConnectionId` pointing at a deleted connection.
- **Cost before the guard** (your two findings above). The sweep found a third: `selectedExistingConnectionIds` rebuilt a `Set` of every connection id on each selection change, and it is read from `ImageViewer` and `ConnectionActionPanel` — neither gated by the Connections tab. The set now lives in its own getter, cached against the connections; a test asserts its identity survives a selection change so the allocation cannot quietly move back inside.

Both shapes, with their concrete instances and a "what is the other one?" table, are now in the `fixing-review-findings` skill and in `CLAUDE.md`, so they carry to other features rather than just this one. The feature doc carries a 31-item regression checklist naming the test that holds each invariant.

### Verification

`tsc` clean, `lint:ci` zero warnings, 3,116 tests, both CI jobs green.

Re-verified in the browser after the scoping refactor, since it replaced a whole getter: all four scopes correct (all 5/5, location 3 correctly excluding an off-location link, selected 1, filtered 5), hover cleared when the hovered connection is deleted, cap short-circuit, track grouping, click-to-select and plain-click-to-highlight, and both tab badges.

@codex review

arjunrajlab:
### Timelapse hover highlight (round 12)

Reported: highlighting a connection line works in normal mode but not in timelapse mode.

**Cause.** The timelapse layer bakes styling in at draw time, so a state change is reflected there only if something rebuilds it. Hover was deliberately excluded from that (rebuilding ~2,500 line features per row made the list sluggish) and only selection rebuilt — but a row click *highlights via hover* rather than selecting, so the feature's main gesture had no visible effect there. Same for a plain click on a track segment in the viewer.

**Fix.** Restyle the drawn segments in place rather than rebuilding them. Each segment now carries `timelapseBaseStyle` (its appearance minus the highlight) and `connectionIds` (every document sharing its endpoint pair, so a hovered duplicate still lights the pair's one segment). Unchanged segments are skipped and the layer isn't drawn at all when the hovered connection is off-layer. Selection still rebuilds — it also decides which duplicate represents a pair, which is a draw-time choice.

**Cost.** 2,364 segments: 0.8 ms to scan, 6.6 ms median redraw, throttled at 100 ms. The "hover must not rebuild the layer" test still guards the property this exists to preserve.

**Verified live** on a 145-timepoint dataset (4,982 connections): a real click on a list row and a real click on a track segment each leave exactly one line at the hovered width in the aggregated webgl feature's *resolved render style* (`{"3":1312,"5":1,"6":1175}`), and none once hover clears. Stashing the fix reproduces the bug — no widened segment at all.

**Tests.** Five new `AnnotationViewer` cases; four fail without the fix. In-place widen with no rebuild · base-style restore for a time-jump segment (red/dashed/0.7 opacity) · non-representative duplicate · no redraw when the connection is off-layer · no work outside timelapse mode.

The shape is recorded in the feature checklist and in the `nimbus-geojs` / `fixing-review-findings` skills: skipping the reflection of a state change for performance is a correctness decision unless nothing user-facing depends on that state.

@codex review

arjunrajlab:
### Round 12 — fixed, plus the sibling it pointed at

**P2, `AnnotationViewer.vue:1959` — cancel the timelapse restyle throttle during teardown.** Correct, fixed in `bfcf69b9`.

Sweeping the file for the shape turned up a **second, pre-existing** instance: `handleLineScanMouseMove` has never been cancelled either, and its trailing call runs `updateLineScanLine` — writing to the line-scan store and the interaction layer right after the teardown block cleared them. Cancelled both.

The more useful finding is why nothing caught it. The guard test named the five throttles that existed when it was written, so adding a sixth and forgetting its `cancel()` left the suite green — and the seventh had been missing all along. It now **discovers** them from the exposed surface (a function carrying both `.cancel` and `.flush`) and asserts a count floor so the discovery itself can't silently break:

```
AssertionError: expected [ …(2) ] to deeply equal []
+   "restyleTimelapseConnectionsThrottled",
+   "handleLineScanMouseMove",
```

That's how both were confirmed — removing either cancel now fails with the offending name.

Recorded in the feature checklist and the `fixing-review-findings` / `nimbus-frontend` skills: where the invariant is "every X does Y", a test that enumerates today's Xs cannot catch tomorrow's.

3,121 tests green, tsc and lint clean.

@codex review

arjunrajlab:
### Round 13 — fixed in `afaff9c5`

**P2, `AnnotationViewer.test.ts:5915` — discover throttles without relying on `defineExpose`.** Right, and it's the same defect as the one it replaced. Scanning `wrapper.vm` didn't remove the hand-maintained list, it moved it from the test body to `defineExpose`: a throttle nobody exposed is invisible to the scan, and the seven already-exposed ones keep the count floor satisfied. The test would have read as comprehensive while covering exactly the cases that were never at risk.

Now recorded where they're **made** — a `lodash` mock delegating to `importOriginal` (so every fake-timer test that advances through a throttle behaves identically) pushes each wrapper into an array as it's constructed. `<script setup>` runs per instance, so setup-scope throttles are created during mount and land in the recording.

Verified against exactly the case described — un-exposing **and** un-cancelling one throttle. The previous version passed; this one fails with:

```
AssertionError: expected [ 'unexposed#4' ] to deeply equal []
```

Failures carry the exposed name where there is one and `unexposed#N` otherwise, which is what makes the red test actionable. Residual gap recorded in the test comment: a wrapper built lazily inside a handler isn't recorded until that handler runs — all seven today are created in setup.

The meta-lesson went into `fixing-review-findings`: when a test discovers what it checks, ask what feeds the discovery, and whether that feed is itself a list someone has to remember to update.

3,121 tests green, tsc and lint clean.

@codex review

PR #1282, opened 2026-07-26
Replace deprecated Vuetify 4 `dense` prop (#1281)

### LINKED ISSUE #1281: Vuetify 4: replace deprecated `dense` prop (16 occurrences, console warning on every render)
## Summary

Vuetify 4 removed the boolean `dense` prop. Every component still using it logs on render:

```
[Vue warn]: [Vuetify UPGRADE] 'dense' is deprecated, use 'density="comfortable"' instead.
  at <VRow class="my-0" dense="" >
  at <PipelineList key=0 onOpen=fn<openEditor> >
  ...
```

This is pure console noise — the prop is ignored, so the components render at default density rather than the compact spacing originally intended. Opening the Pipelines dialog produces several warnings at once.

This is already the documented convention in the repo; it just hasn't been applied everywhere:

- `codebaseDocumentation/VUETIFY4_MIGRATION.md`
- the `nimbus-frontend` skill: *"`dense` prop is deprecated. Use `density="comfortable"`"*

## All occurrences (16 across 5 files)

| File | Count | Element(s) |
|---|---|---|
| `src/components/pipelines/PipelineEditor.vue` | 8 | 7× `<v-row>`, 1× standalone `dense` attribute on its own line |
| `src/components/pipelines/PipelineStepEditor.vue` | 3 | `<v-row class="my-0" dense>` ×3 |
| `src/components/pipelines/PipelineList.vue` | 2 | line 3 `<v-row class="my-0" dense>`, line 109 `<v-row v-else class="my-2" dense>` |
| `src/components/AnnotationBrowser/AnnotationProperties/PropertyCreation.vue` | 2 | `<v-list-subheader dense>` ×2 |
| `src/components/SharingStatusIcon.vue` | 1 | `<v-card dense class="pa-2" max-width="300">` |

Find them all with:

```bash
grep -rn "dense" src --include="*.vue" | grep -v density | grep -E "<v-[a-z-]+[^>]*\sdense"
```

That regex deliberately matches only a boolean `dense` on a component tag, so it skips the several prose comments in `ImageViewer.vue`, `AnnotationViewer.vue`, `ValueSlider.vue` and `Home.vue` that contain the word "dense" and must not be touched.

## Fix

Replace the boolean with `density="comfortable"` — the repo's standard substitution, which preserves roughly the Vuetify 2/3 `dense` spacing:

```diff
-    <v-row class="my-0" dense>
+    <v-row class="my-0" density="comfortable">
```

A scripted sweep works, but run prettier afterwards: the longer attribute pushes two tags over the line-length limit and `pnpm lint:ci` fails on the formatting (`PipelineEditor.vue:42` and `PropertyCreation.vue:13` specifically).

```bash
python3 - <<'PY'
import re, pathlib
for t in ["src/components/pipelines/PipelineEditor.vue",
          "src/components/pipelines/PipelineStepEditor.vue",
          "src/components/pipelines/PipelineList.vue",
          "src/components/AnnotationBrowser/AnnotationProperties/PropertyCreation.vue",
          "src/components/SharingStatusIcon.vue"]:
    p = pathlib.Path(t)
    s, n = re.subn(r'(<v-(?:row|card|list-subheader)\b[^>]*?)\sdense(?=[\s>])',
                   r'\1 density="comfortable"', p.read_text())
    p.write_text(s); print(t, n)
PY
pnpm lint:fix && pnpm lint:ci && pnpm tsc && pnpm test
```

Note the standalone `dense` on its own line in `PipelineEditor.vue` is not matched by that regex — handle it by hand.

## Verification

Open the Pipelines dialog and the sharing status popover with the console open; there should be no `[Vuetify UPGRADE]` warnings. Visually confirm row spacing in the pipeline editor and list, since these components were written expecting the tighter spacing.

## Not included

A second, unrelated console warning is also outstanding and would need its own investigation:

```
[Vue warn]: Component inside <Transition> renders non-element root node that cannot be animated.
  at <VOverlay key=0 absolute="" opacity="0.2" ...>
  at <FadeTransition>
  at <LargeImageDropdown>
  at <NavigatorPanel>
```

_Split out of #1279, where these warnings surfaced during testing but were unrelated to that branch._

arjunrajlab:
Fixes #1281.

## Verified, with two corrections to the issue

**`dense` was not being ignored.** The issue says the components "render at default density rather than the compact spacing originally intended." `VRow` maps `density === 'comfortable' || props.dense` to the same `v-row--density-comfortable` class, so `dense` still worked — only the console warning was new. Confirmed live: pre-fix the rows rendered `v-row--density-comfortable`, and post-fix all 6 rows in the dialog do too, with zero `v-row--density-default`. **This PR is a pure warning cleanup with no visual change.**

**There are 19 occurrences, not 16.** `PropertyCreation.vue` has 5, not 2. The issue's regex was scoped to `<v-[a-z-]+`, so it missed `dense` on three *custom* components — which happen to be the ones needing the opposite treatment.

## Why the 6 non-`VRow` cases are deleted, not converted

`VRow` is the **only** component in Vuetify 4 that calls `deprecate('dense', …)` (`VRow.js:167`). Everything else was silent *and* dead: `VCard`, `VListSubheader`, `tag-picker`, `docker-image-select` and `property-worker-menu` declare no `dense` prop, so Vue passed it through to the DOM as a stray attribute. Measured in the live DOM: **3 stray `[dense]` attributes pre-fix, 0 after.**

The issue's scripted sweep would have been wrong on all of these:

| Target | `dense` prop? | `density` prop? | Correct fix |
|---|---|---|---|
| `v-row` × 13 | yes (deprecated) | yes | `density="comfortable"` |
| `v-card` × 1 | no | **yes** | delete — a swap would newly tighten title/subtitle/text padding |
| `v-list-subheader` × 2 | no | no | delete — a swap is just a different dead attribute |
| 3 custom components | no | no | delete — and `TagPicker`/`DockerImageSelect` already set `density="compact"` internally |

Note Vuetify contradicts itself here: `makeVRowProps` is annotated `@deprecated use density="compact"` while the runtime warning and the class mapping both say `comfortable`. `comfortable` is the behaviour-preserving one.

## Verification

`pnpm tsc` clean, `pnpm lint:ci` clean, 3123 tests across 186 files pass. (Two vitest stderr `Could not parse CSS stylesheet` errors are pre-existing jsdom-vs-`@layer` noise in files this PR doesn't touch — confirmed on stashed-clean `master`.)

In-browser A/B, since the whole issue is a console warning:

- **Pre-fix** (`git stash`): 2 `[Vuetify UPGRADE]` warnings at page load, trace matching the issue exactly — `<VRow class="my-0" dense="">` at `<PipelineList>`, with `modelValue=false`.
- **Post-fix**: zero, with `[vite] connected` still captured to prove the reader was live.

This took three attempts because the first two probes *couldn't fail*: `PipelineDialog` sets `eager`, so its content mounts at app boot and the one-shot warning fires before any console hook can be installed — and toggling the dialog re-runs no `setup()`. That trap is now written into the `in-browser-testing` skill.

**Not verified live:** `SharingStatusIcon` renders only `v-if="getConfigSharingInfo(...)"` and no dataset on hand was shared, so its popover never mounted. Its change rests on `VCard`'s prop table plus the 3→0 stray-attribute count — the same mechanism verified on `v-list-subheader`.

## Durable artifacts

Per `CLAUDE.md` ("turn review findings into durable artifacts"):

- **`src/vuetifyDeprecations.test.ts`** (new) — scans every `.vue` template for boolean `dense` on any tag, with self-tests for the scanner itself. Failed with exactly 19 offenders before the fix.
- **`nimbus-frontend` skill** — why `VRow` is the only warner, the JSDoc/runtime contradiction, and why a blanket regex sweep is wrong in both directions.
- **`in-browser-testing` skill** — sixth false-pass trap: an empty console reads as "clean" when it means "not capturing".
- **`VUETIFY4_MIGRATION.md`** — corrected; it had listed the custom components under "What Was NOT Changed" as "app-level props", which was factually wrong (none ever declared the prop).

## Out of scope

The `<Transition>` non-element-root warning at `<LargeImageDropdown>`/`<NavigatorPanel>` is untouched, as the issue specifies.

https://claude.ai/code/session_01UPECZgLuVqaRom51j1su9y

### DIFF
--- a/.agents/skills/in-browser-testing/SKILL.md
+++ b/.agents/skills/in-browser-testing/SKILL.md
@@ -9,7 +9,7 @@ description: "Use when verifying NimbusImage changes in a live browser tab (loca
 
 Multiple real bugs in this repo passed tsc, lint, all unit tests, and code reasoning — and only broke in the live app (pointer-events layering, deep-watch over-firing, HMR-corrupted store, background-tab false-passes). For any user-facing change, in-browser verification is a required gate, and it must be done in a way that can actually fail. The traps below each produced a false "it works" at least once.
 
-## The five false-pass traps
+## The six false-pass traps
 
 1. **Synthetic clicks lie.** `element.dispatchEvent(new MouseEvent('click'))` bypasses hit-testing, `pointer-events`, z-index, and overlays. A synthetic-click walkthrough once reported a tour "working" while real user clicks did nothing. Verify clickability with `document.elementFromPoint(cx, cy)` (is the intended element the topmost hit target?), then click with the real `computer` tool. Reserve synthetic dispatch for *driving* flows fast only after clickability is independently confirmed.
 2. **Background tabs lie.** A hidden tab pauses `requestAnimationFrame`, throttles `setTimeout` to ~1 Hz, and never composites — perf probes report 0 long tasks and event loops stall/time out. The tab must be foreground/visible for any timing or rendering claim. If a paced loop must survive a background tab, use busy-wait gaps (`while (performance.now() < end) {}`), not setTimeout.
@@ -21,6 +21,20 @@ Multiple real bugs in this repo passed tsc, lint, all unit tests, and code reaso
    // draw to an offscreen canvas, then getImageData → report max/mean channel value
    ```
    A non-zero max means the tiles are there. Note that a fluorescence dataset is *legitimately* near-black at fit-to-view zoom (sparse bright cells, `hist.max` pinned to 65535 by a few saturated pixels), so a low mean is not evidence of a bug either. Verify tile health from `statusCode`s in `read_network_requests` (`/tiles/zxy/` → 200) plus the layers' `idle`/`_imageUrls`, and verify overlays from live DOM state — not from the picture.
+6. **An empty console reads as "clean" when it means "not capturing."** `read_console_messages` only records from the moment it is **first called**, so a warning emitted during app boot is invisible unless you call it *then* reload *then* read. Worse, a `console.warn` wrapper installed via `javascript_tool` can never see boot-time output at all — the app has already mounted by the time you can install it. Chasing a missing `[Vuetify UPGRADE]` warning this way produced two false passes in a row (issue #1281) before the order was fixed.
+
+   Compounding it: **`eager` dialogs mount at app boot, and a mount-time warning fires exactly once.** `PipelineDialog.vue` sets `eager` so its content stays mounted while closed — so closing and reopening the dialog does *not* re-run any child's `setup()`, and any one-shot warning there (Vuetify deprecations, prop validators) never fires again for the life of the page. Toggling the dialog to "re-trigger" it silently measures nothing.
+
+   Working order for any boot-time console assertion:
+   ```
+   1. read_console_messages{clear:true}    // start tracking / empty the buffer
+   2. location.reload()                    // via javascript_tool; a #hash nav is NOT a reload
+   3. wait for boot (poll store state)
+   4. read_console_messages{pattern:…}     // the real read
+   5. read again with a broad pattern (e.g. "vite") to prove capture is live —
+      seeing "[vite] connected" is what makes an empty filtered result meaningful
+   ```
+   Then invert it: put the pre-fix code back (`git stash`) and confirm the same sequence *does* report the warning. Verify which code is actually being served rather than assuming the stash took effect — `fetch('/src/path/To.vue?t=' + Date.now())` returns Vite's current transform, and counting tokens in it (`dense` vs `comfortable`) settles it. Rendered DOM often can't: `dense` and `density="comfortable"` produce the *same* `v-row--density-comfortable` class, so the DOM looked identical in both states.
 
 ## Console handles (javascript_tool)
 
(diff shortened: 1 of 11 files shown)


arjunrajlab:
The P2 is legitimate and fixed in `6a81ccf4`. The tag pattern began with `[a-z]`, so `<VRow dense>` was skipped entirely — confirmed directly: the old pattern returns `[]` for both `<VRow dense>` and `<PipelineRow dense />`.

Generalizing the shape — *the guard's pattern structurally cannot see a valid spelling of what it guards* — turned up three more holes of the same kind, each now covered by a self-test:

- **PascalCase tags** (`<VRow dense>`) — your finding.
- **Bound forms** `:dense="x"` / `v-bind:dense="x"`, missed because the token pattern required whitespace before `dense`.
- **The inverse:** `class="a dense b"` was a false *positive*, since the token was tested against the blob with quoted values still in it. Attribute values are now blanked before the name check.

The scan is now scoped to the SFC `<template>` block, which is what lets the tag pattern accept PascalCase without TS generics in `<script>` (`ref<VForm>()`, `Ref<PaletteRefEl | undefined>`) registering as tags. A new test asserts >100 files and >1000 tags scanned, and that every `.vue` file has a template block — so the guard fails loudly rather than going vacuous if extraction ever breaks.

One correction: the supporting claim that "this repository already uses PascalCase component tags" doesn't hold — every PascalCase `<Ident` in `src` is a TypeScript generic in a script block, not a template tag. The fix stands on its own merits regardless, since PascalCase is valid Vue markup and Vuetify's own docs use it.

Verified the hardened scanner still catches the real thing: with `git checkout master -- src/components` it reports exactly 19 offenders. `tsc`, `lint:ci` and 3125 tests pass; CI green.

@codex review

PR #1288, opened 2026-07-28
Timelapse panel: move controls out of the Navigator, add track colouring and track actions

arjunrajlab:
## What

The timelapse controls were inline in the Navigator — mode checkbox, track-window slider, tag picker, labels checkbox and a delete-all button — which made the Navigator taller than the three sliders it exists for.

They now live in a **Timelapse palette** at `left: 444`, immediately right of the Navigator over empty image space, so turning the mode on doesn't push Layers and Tools down. (444, not 404: the left palettes aren't one width — Layers is 420, and 404 put the panel 32px on top of it.)

**Panel visibility *is* the mode.** Closing the palette turns timelapse off, so there's no "mode on, panel hidden" state to reason about. The Navigator keeps only the toggle. The `data-tour` anchors moved with their controls, so `testTimelapseTour.yaml` still resolves all three.

New controls in the panel:
- **Colouring toggle** — per-track hue vs. uniform white (easier to read when many tracks overlap), plus a shuffle that re-permutes the assignment.
- **`N tracks · M links`** readout with **Show tracks**, which opens the Object Browser directly into Connections → By track.

**Tracks in the Object Browser** reuse the existing By-track grouping rather than adding a third tab — one implementation over the same connected-components data. Each header gains a colour swatch matching the drawn line, and a **Select** action for the track's objects.

## Two track-colour bugs, both invisible to synthetic fixtures

**Luminance under the hash.** The old colour sliced the hash's own hex digits straight into `#rrggbb`, so lightness was whatever the hash happened to be — roughly a third of tracks rendered near-black or near-white and read as unhighlighted against the image. Saturation and lightness are now fixed; only the hue varies.

**Adjacent ObjectIds → adjacent hues.** Track ids are ObjectIds allocated in one batch, so neighbouring tracks differ in the last character. Under `hash % 360` that's a one-degree hue difference. On a real dataset the first five tracks came out `rgb(80,226,{162,218,215,213,211})` — five indistinguishable greens. The hue is now a golden-angle multiple, mapping hashes one apart to hues 222° apart.

Smallest neighbouring-id hue gap over 40 consecutive ObjectIds:

| | min adjacent gap |
|---|---|
| `hash % 360` (the bug) | 1.0° |
| `hashString` + `% 360` | 3.0° |
| `hashString` + golden angle | 9.2° |
| **this hash + golden angle** | **77.3°** |

That table is why this deliberately does **not** reuse `hashString` from `utils/annotation`, despite it being the stronger hash with a comment about breaking ObjectId correlation. Its murmur finalizer exists to *destroy* the sequential correlation the golden angle needs — the two are in direct tension, and swapping it in measures worse. I tried it, the test caught it, and the reason is now documented at the call site so the next reader doesn't repeat the attempt.

Separately, the viewer and the list now key colour off a shared `trackKey` (smallest member id) instead of `Set` insertion order, so a swatch always matches its line.

## Verification

Frontend-only. `pnpm tsc`, `pnpm lint:ci`, and **3145 tests across 187 files** pass.

Verified in a **foreground** browser tab on a 145-timepoint, 4,982-connection dataset with real clicks: panel placement with no overlap, uniform mode repainting all 1313 segments, shuffle changing all 1313, Show tracks landing on the right tab *and* grouping, Select objects selecting exactly the track's 18, close turning the mode off and clearing the layer, and all three tour anchors hit-testable. **Swatch and line colour agreed on all 123 visible tracks.**

Every new test was confirmed to **fail** against the pre-fix code before being accepted. One of them initially passed only on mock state leaked from an earlier test in the file, and another didn't discriminate because its fixture had insertion order equal to sort order — both are fixed and re-verified.

## Later rounds: track framing, panel collisions, and a second review

**Expanding a track frames it** in the viewport (`frameCameraInfoToExtent`), sized so the track occupies a fifth of the view. A new camera helper rather than `frameCameraInfo` because a track is a 2D box: projecting `(w, h)` as one signed vector fits only one of its two diagonals, so under rotation the other falls outside the viewport. The box support function `w·|û.x| + h·|û.y|` is sign-free. It also zooms **in**, which `frameCameraInfo` deliberately never does, and clamps to the live map's `zoomRange` — a track whose members share a centroid otherwise demands infinite zoom, and GeoJS clamps `map.zoom()` silently, desyncing `gcsBounds` from `zoom` and corrupting hydration. Framing happens on expand only; collapsing would drag the camera back every time you tidied the list.

XY/Z follow the track (it isn't drawn elsewhere). **Time doesn't** — it's the timelapse window's centre and you scrub it deliberately, so it only clamps to the nearest end when the current frame is outside the track's range.

### The panels then collided with two different overlays in turn

The selection action panels slide to `--nimbus-left-palette-clear-x` (446) when a left palette is open, which is where the Timelapse palette sits (444–744) — so they were drawn underneath it. Re-anchoring them to the right edge fixed that and created the same bug against a *different* overlay: `.ai-panel` is `z-index: 2001` against their 1000 and is mutually exclusive with neither timelapse mode nor the Object Browser. Measured at 1684×857, **6 of the two panels' 8 buttons failed `elementFromPoint`** — including `Deselect All`, the only non-destructive way to dismiss them, leaving two reachable buttons that both delete things.

The fix is one resolved offset instead of a class per overlay: `rightEdgeClearX()` takes the max over whatever is open, App.vue projects it as `--nimbus-right-edge-clear-x`, and each panel needs a single CSS rule. Adding a third right-edge overlay is now one term rather than a new class and 2ⁿ rules. Max, not sum — they share the edge rather than queueing along it.

| | annotation panel | connection panel | blocked buttons |
|---|---|---|---|
| under the Timelapse palette | x 446–626 | x 446–661 | — |
| under the AI panel | 171×138px covered | 206×103px covered | **6 of 8** |
| after the fix | x 953–1128 | x 918–1128 | **0** |

### A `/branch-review` of the branch found five more

1. **Track swatches showed with timelapse mode off**, where `trackColor` is never reached — 248 swatches in 248 hues against **zero** drawn connection features, since a timelapse link's endpoints sit on different timepoints and normal mode never co-displays them. And unturnoffable, because the colouring toggle lives in the palette that *is* the mode. This was the pair already written into `CONNECTION_LIST.md` ("a track's colour in the viewer | its swatch in the Connections tab"), missed on the mode gate.
2. **`Delete all timelapse connections` was disabled on the total connection count** while the action deletes only the tagged subset, so on a dataset of hand-made connections it was enabled and silently did nothing. The old Navigator button had no guard at all — the guard was new, and counted the wrong set.
3. **Palette geometry had three encodings of one number**, with two different gaps (16+420+8=444 in App.vue vs a hardcoded 446 in `style.scss`). Both copies had already gone wrong: a stale Layers width drew the palette 32px over Layers, and 444 landing 2px from 446 is what caused the collision above. Now one source in `@/utils/paletteGeometry`, feeding the `:width` props and projected onto `<v-app>`; `style.scss` declares neither.
4. A duplicate `watch([selectedAnnotationIds, hoveredAnnotationId])` folded into the identical watcher 38 lines above it.
5. **Timelapse state extracted to `src/store/timelapse.ts`** — the item deferred twice below, now done. Six fields, seven mutations. Mechanical because none of it is persisted.

The extraction's real risk was the harness, not the source: `AnnotationViewer.test.ts` drives the timelapse draw path entirely through watchers, so a plain-object mock of the new module would let all ~48 timelapse assertions pass against a layer that was never rebuilt. Confirmed `reactive()` is load-bearing by removing it — 6 tests fail, including both colouring-watch tests, while 295 stay green.

Also fixed a cascade in `TimelapsePanel.test.ts`: its `Object.defineProperty` accessor was removed after the assertions rather than in a `finally`, so one failure left a getter-only `trackCount` on the shared mock and reported four extra failures for one cause.

**Verification for these rounds.** `tsc`, `lint:ci`, **3185 tests across 188 files**. Every new test confirmed to fail against the pre-fix code (380 vs 420, 16 vs 556, 544 vs 556, `'false'` vs `'true'`, `true` vs `false`). I strengthened the geometry consistency test after noticing it compared two values derived from the same constant it was meant to check, so it could not fail.

Live from a fresh page load, driving the new store module: mode off 0 points / 0 segments; on 1,966 / 1,714 in 119 colours; uniform the same counts in **1** colour; shuffle 120 colours with the sampled hue changed; window 30 → 2,728 / 2,488. Track framing measured at `fractionOfViewport = 0.200` exactly with track centre == viewport centre, and 0.00 camera movement on collapse.

One trap worth recording: reproducing the *before* state of the overlap needed the transform transition disabled first. A hidden tab never runs CSS transitions, so `translateX(430px)` had never applied and the panels measured at their untransformed `left: 16` — which reads as "no overlap, nothing to fix."

## Chatbot help topics

The AI panel's help topics still described the pre-refactor UI ("Adjust track window", "checkbox in variable navigation panel"). Updated `connections.md` (the Timelapse palette and all its controls, plus a new **Track View in the Object List** section), `workflows.md`, `interface-navigation.md`, `interacting-with-objects.md`, and added **track** to `concepts_core.md` — the only always-loaded file, and the agent gets asked how many tracks there are with no tool that answers it. These are baked into the girder image at `COPY` time, so they need `docker compose build girder`, not a restart.

## Not done

- **`Delete all timelapse connections` was not exercised.** It's wired and now correctly disabled unless tagged connections exist, but clicking it would have destroyed the test dataset's 4,982 real connections.

`CONNECTION_LIST.md`'s regression checklist gains a **Track colouring** section, a **Cost** entry for the new panel's read gating, and three rows in the symmetric-pairs table.

https://claude.ai/code/session_01XtohtBv8wj1GU77KX6gN9u


arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

PR #1289, opened 2026-07-29
Aim navigation at unrolled positions, not raw centroids (#1280)

### LINKED ISSUE #1280: Go-to-location navigation ignores unroll offsets and pans to the wrong tile
## Summary

When an axis is unrolled (`unrollXY` / `unrollZ` / `unrollT`), clicking a row in the Object Browser pans the viewer to the annotation's **raw** centroid rather than its **rendered** position on the unrolled grid. The camera lands one or more tile-widths away from the object the user clicked.

This affects the existing **Objects** tab, and the **Connections** tab added in #1279 inherits it.

## Evidence

Measured live on `normmedia_8well_col2_livecellgfp` (dataset `69f4d960aaba948c2d7b9803`, 4 timepoints) with `unrollT` enabled, tile width 1024:

| | value |
|---|---|
| Annotation raw centroid | `x = 260` |
| Where it actually renders on the unrolled grid | `x = 1284` |
| Where `goToAnnotationLocation` centred the camera | `x = 260` |
| Error | **1024 world units — exactly one tile** |

The object is on the second time tile, so the viewer scrolls to the corresponding spot on the *first* tile instead.

## Cause

`src/utils/annotationNavigation.ts` centres on the annotation's raw centroid:

```ts
const center = annotation?.coordinates
  ? simpleCentroid(annotation.coordinates)
  : stub?.centroid ?? annotationStore.annotationCentroids[annotationId];
store.setCameraInfo(recenterCameraInfo(store.cameraInfo, center));
```

The viewer draws annotations at *unrolled* coordinates instead, via `unrolledCoordinates()` in `AnnotationViewer.vue`, which offsets each point by the tile the annotation's location maps to:

```ts
x: tileW * tileX + point.x,
y: tileH * tileY + point.y,
```

Navigation and rendering therefore disagree whenever unrolling is active.

## Why it wasn't fixed in #1279

That PR was already large, and this is a pre-existing defect in shared navigation rather than something the connection list introduced — the Objects tab has behaved this way since before it. Fixing it well is not a local change:

`unrolledCoordinates` depends on `props.unrollW`, which lives as a `ref` in `ImageViewer.vue` and is passed down as a prop, so it is not reachable from a util. It is derivable — a pure function of the unroll cell count and image size (`ImageViewer.vue:1481`) — so the clean fix is probably:

1. expose the unroll grid (`unrollW` / `unrollH`) on the store where `ImageViewer` computes it,
2. extract the unroll offset into a shared util taking `(centroid, location, image, unrollState)`,
3. route `goToAnnotationLocation` **and** `goToConnection` through it.

That touches rendering state and changes Objects-tab behaviour, so it deserves its own review.

## Extra consideration for connections

`goToConnection` currently frames both endpoints only when they share a frame, on the grounds that endpoints on different frames can't both be displayed. **That is false while unrolling** — all frames are on screen at once, and cross-frame connections are genuinely drawn between grid tiles (verified: a Time 0 → Time 1 link renders as a line between two tiles). Once unrolled coordinates are available, the same-frame condition should be relaxed to "both endpoints displayed", and the framing should use rendered positions.

## Reproduce

1. Open a multi-timepoint dataset and enable Unroll over time.
2. In the Object Browser → Objects, click a row for an object on any tile after the first.
3. The viewer pans to the equivalent position on the first tile, not to the object.

_Found while reviewing #1279; raised by Codex review round 5._

arjunrajlab:
Fixes #1280.

Clicking an Object Browser row while an axis is unrolled panned the camera to the annotation's **raw** centroid, while the viewer had drawn it offset by its grid cell — so the camera landed one or more tile-widths away, on the equivalent spot of the *first* tile.

Measured live on `normmedia_8well_col2_livecellgfp` (4 timepoints, tile 1024, 2×2 grid), clicking the row for a Time-4 object:

| | camera | GeoJS drew it at |
|---|---|---|
| before | (380, 380) | (1404, 1402) |
| after | (1404, 1402) | (1404, 1402) |

## Why it needed structure rather than a one-line change

The offset math was inline in `AnnotationViewer.vue` and depended on `unrollW`, which lives as a `ref` in `ImageViewer.vue` and is only reachable by the children it's passed to as a prop. A util couldn't see it, which is exactly why the two paths drifted.

- `src/utils/unroll.ts` now holds the geometry, pure and store-free. `unrollGridSize` is the **single** copy of the layout formula.
- `ImageViewer.draw()` assigns its `unrollW`/`unrollH` refs from it.
- `store.unrollGrid` mirrors it off the same `layerStackImages` entry, so navigation gets the same grid *by construction*.
- The draw path and the navigation path both build an `IUnrollLayout` via `unrollLayoutFor` and call the same transform.

One deliberate asymmetry, documented at the call site: the viewer takes `unrollW` from its **prop**, not `store.unrollGrid`. That keeps drawing keyed to the grid the tiles were actually laid out on, and keeps `unrolledCentroidCoordinates` — a computed over *every* annotation — off `layerStackImages`, which rebuilds tile URLs on every contrast tweak.

## `goToConnection`: "same frame" was the wrong test

It only framed both endpoints when they shared a frame, on the grounds that two frames can't both be displayed. That's false while unrolling — every frame is on screen and a cross-time connection is genuinely drawn as a line between tiles. The gate is now per-axis `unrolled || indices match`, and framing uses drawn positions, so a cross-tile span comes out tile-widths wide instead of collapsing to the distance between two raw centroids. Verified: a Time 1 → Time 2 link now frames with both endpoints on screen.

## Verification

- `tsc`, `lint:ci`, and 3154 frontend tests pass.
- The 3 new navigation tests were confirmed to **fail** without the fix, reproducing the issue's own measured numbers (260 vs 1284).
- Verified in a live browser with real clicks, before and after, including a re-verification after the review refactor changed the draw path.

## Review round included in this PR

`/branch-review` turned up seven findings, including **a regression this branch introduced**: the layout was being constructed inside the per-annotation transform, allocating two objects per annotation — including on the un-unrolled path, which this branch's own regression checklist says must allocate nothing. Now hoisted into a computed, with a test that asserts the layout count doesn't scale with annotation count (40 layouts for 40 annotations before, 2 after). Details in the second commit.

## Out of scope, but worth knowing

The same coordinate-space bug survives in the **hit-testing family** — pre-existing, untouched here, and confirmed live: while unrolled, **clicking an object selects nothing, and clicking the empty equivalent spot on the first tile selects it.** The annotation spatial index is built from raw centroids while its queries come from map space, which also affects hover highlight, the right-click menu, the alt-drag ghost, the viewport hydration partition (can silently collapse the drawn set on stub-mode datasets), and the agent's `set_camera` fit. That's a different subsystem and deserves its own review — the same reason #1279 deferred this one. Recorded in `codebaseDocumentation/UNROLL_NAVIGATION.md` and the `nimbus-geojs` skill; happy to file it as a follow-up issue.

## Docs

- `codebaseDocumentation/UNROLL_NAVIGATION.md` — the three coordinate spaces, the structure, and a machine-checked regression checklist (every item names its test).
- `nimbus-geojs` skill gains a "unrolled coordinates are a third space" section, since this trap will recur in other features.

Two test fixtures had been hiding bugs and are fixed here: `mockedStore.unroll` was an independent field where the real store derives it from the three flags (two tests passed only because of that), and the nav test's mock of `@/utils/annotation` now spreads `importOriginal` so `unrollIndexFromImages` stays real — a fixed-value stub there would have made every cell-offset assertion pass for the wrong reason.

https://claude.ai/code/session_01XtohtBv8wj1GU77KX6gN9u


arjunrajlab:
@codex review

PR #1295, opened 2026-07-31
Name the worker interface job instead of titling it "unknown"

### LINKED ISSUE #1294: Worker interface requests create jobs titled "unknown" (and containers named unknown_None_<ts>)
Worker **interface** requests create Girder jobs titled `unknown`, and Docker containers named `unknown_None_<timestamp>`.

### What happens

`runJobRequest` derives both the Girder job title and the container name from `params["name"]`, falling back to the literal string `"unknown"`:

```python
# server/helpers/tasks.py:22
name = params.get("name", "unknown")
...
"name": "{}_{}_{}".format(name, datasetId, datetime.datetime.now().timestamp()),
"girder_job_title": name,
```

Of the four request types, only the interface request omits `name`:

| Caller | params | Resulting title |
|---|---|---|
| `models/annotation.py:947` (`compute`) | `tool` — has `name` | tool name |
| `models/property.py:74` (`compute`) | has `name` | property name |
| `models/workerPreviews.py:82` (`preview`) | has `name` | tool name |
| `models/workerInterfaces.py:81` (`interface`) | `{"image": image}` — **no `name`** | `unknown` |

So every interface fetch (which happens whenever a worker tool is added, or when its interface isn't cached — i.e. immediately *before* a real run) leaves a job called `unknown` in Settings → Jobs & Logs, and a container named `unknown_None_1753980000.123` (`datasetId` is `None` for interface requests).

### Why it's worth fixing

- Users see an unexplained `unknown` job appear right before their segmentation run and can't tell it apart from the run itself, or tell whether it failed.
- The container name is equally opaque when debugging with `docker ps`.
- The AI assistant can't explain it either: nothing in `agent_system_prompt.txt` or the help topics mentions that an interface job exists. (The assistant is otherwise correctly insulated from these jobs — they carry no `toolId`/`propertyId`, so they can't reach `jobIdForToolId`/`jobIdForPropertyId` or `wait_for_job`.)

### Suggested fix

Give the interface request a real name — e.g. pass `{"image": image, "name": f"Interface {image}"}` from `requestWorkerUpdate`, or default `girder_job_title` from `requestType` in `runJobRequest` when `name` is absent. Note `name` is filtered to `[a-zA-Z0-9_.-]` for the container name, so an image-derived name is safe but loses `/` and `:`; consider setting the container name and `girder_job_title` from separate values so the title can stay readable.

Worth also adding a line to the chat plugin's `help/troubleshooting.md` describing what the interface job is, so the assistant can answer "what is this job?".

arjunrajlab:
Fixes #1294.

## Problem

`runJobRequest` derived both the Girder job title and the docker container name from `params["name"]`, defaulting to the literal `"unknown"`. Interface requests are the only one of the four callers that has no name to offer, so every interface fetch — which happens whenever a worker tool is added, i.e. immediately *before* a real run — left a job titled `unknown` in Settings → Jobs & Logs and a container named `unknown_None_<ts>`. Users couldn't tell it apart from the run itself, and the assistant couldn't explain it either.

## Changes

Backend only; no frontend changes were needed. The frontend only *reads* `job.title` (`src/store/jobs.ts`, `JobsLogs.vue`) and sends just `image` for interface requests, so the better title propagates on its own.

- **`server/models/workerInterfaces.py`** — pass an explicit `jobTitle="Pulled worker interface for <image>"`. It sets `girder_job_title` only, so the `--parameters` payload the worker parses stays exactly `{"image": image}` (locked by a test).
- **`server/helpers/tasks.py`** — derive the container name and the job title separately. Container names allow only `[a-zA-Z0-9_.-]`, and sharing one sanitized value cost every title its spaces, `/` and `:`; compute and preview jobs now read `Cellpose Segmentation` instead of `CellposeSegmentation`. Also:
  - fall back to the **request type** rather than `"unknown"`, so a future caller that omits a name still gets an identifiable job — this generalizes the fix past the one caller that triggered the issue;
  - join only the container-name parts that exist, so an absent `datasetId` no longer appears as the literal `None`;
  - guard a non-string `name` — `re.findall` on an int was an uncaught `TypeError` (500).
- **`girder-claude-chat/.../help/troubleshooting.md`** — a new "Unfamiliar Jobs in Jobs & Logs" entry explaining what the interface job is, that it does no analysis, and what to check if it fails, so the assistant can answer "what is this job?".
- **`.claude`/`.agents` `nimbus-backend` skill** — records the rule (job titles are user-visible; never default one to a placeholder; derive container name and title separately; document jobs users didn't start), both trees synced via `sync_skills.py`.

## Verification

- **New unit tests** (`test/test_tasks.py`, 10 cases): interface title/container name, the worker payload staying unchanged, named requests keeping spaces in the title, and unusable names (missing, empty, whitespace, int, dict, all-invalid-characters) falling back to the request type. Confirmed **9 of the 10 fail** against the unfixed source with the changes stashed; the 10th is the payload invariant, which must hold both before and after.
- **Full backend suite:** `tox` → 368 passed. `flake8` clean on all changed files.
- **End-to-end against a rebuilt local backend** — real `POST worker_interface/request` for `annotations/connect_to_nearest:latest`:
  - title: `Pulled worker interface for annotations/connect_to_nearest:latest`
  - container name: `interface_1785493338.471857` (was `unknown_None_<ts>`)
  - job status `3` (SUCCESS), so the interface fetch itself still works

The container name no longer embeds the image, but `docker ps` shows the image in its own column, so `interface_<ts>` plus that column is unambiguous.

## Noted, not fixed

The three `runJobRequest` callers that read a JSON request body (`property.compute`, `worker_preview/request`, annotation `compute`) would still 500 rather than 400 on a non-object body — a separate input-validation gap for `server/helpers/validation.py`, out of scope here.

https://claude.ai/code/session_01LE4Kf3kUrbZt3MfMqacuba

### DIFF
--- a/.agents/skills/nimbus-backend/SKILL.md
+++ b/.agents/skills/nimbus-backend/SKILL.md
@@ -305,6 +305,12 @@ Match each kind of request-data access to its helper:
 
 Rules: validation and `RestException` live in the API layer, never in models. Validate NESTED elements, not just the top-level container — each list entry (`[123]`) and nested map (`propertyValues: {"a1": 5}`) is caller-supplied; `.get()`/`.items()` on a non-dict entry → 500. Add a backend test per malformed-input case (malformed body → 400, not 500); `test/test_validation.py` unit-tests the helpers directly, and endpoint tests assert the 400. When you fix one endpoint, sweep the other endpoints in the same file for the identical gap — reviewers flag one instance per round.
 
+**`assertStatus(resp, 400)` alone is not a regression test for input validation.** These endpoints have *other* 400 paths — a missing `datasetId`, an unknown dataset id, a failed schema validation — so a malformed-body test can pass while the body is never validated at all. Observed for real: a `/upenn_annotation/compute` test using a syntactically valid but nonexistent `datasetId` passed **before** its fix, because the model's dataset lookup rejected the request first.
+
+Two habits close it:
+- **Assert the message, not just the status** (`assert "must be a JSON object" in resp.json["message"]`), and set up the request so the code actually reaches the validation — use a real `utilities.createFolder(...)` dataset when the handler looks one up before touching the body.
+- **`git stash push <source files>` and confirm the test fails**, leaving the new test file in place (untracked files aren't stashed). A malformed-input test that passes both ways is worse than none.
+
 ## Loading Plugin Changes Into the Running Backend
 
 The `girder` container bakes the plugin into its image (no source mount). After editing backend plugin code:
--- a/.agents/skills/nimbus-backend/SKILL.md
+++ b/.agents/skills/nimbus-backend/SKILL.md
@@ -363,6 +363,18 @@ job = JobModel().createLocalJob(
 JobModel().scheduleJob(job)
 ```
 
+### Every Job Title Reaches Users — Never Ship a Placeholder
+
+Job titles are user-visible: they are listed in Settings → **Jobs & Logs** and quoted in the frontend's job notifications (`src/store/jobs.ts`). A title that doesn't identify the work is a support burden — issue #1294 was a `girder_job_title` defaulting to the literal `"unknown"` for worker *interface* requests (containers named `unknown_None_<ts>`), which users saw appear right before their segmentation run with no way to tell the two apart.
+
+Rules when adding a job, or a helper that creates jobs:
+
+- **Default to something meaningful, not a placeholder.** If the title comes from caller-supplied data (`params.get("name")`), fall back to the request/job type, never to `"unknown"`. Check *every* caller — one caller omitting the field is how the placeholder reaches production.
+- **Derive the container name and the title separately.** Docker names allow only `[a-zA-Z0-9_.-]`, so sanitizing shared text costs the title its spaces, `/` and `:`. `runJobRequest` takes an explicit `jobTitle` for this reason (`server/helpers/tasks.py`).
+- **Don't interpolate `None` into a name.** Join only the parts that exist — `datasetId` is absent for interface requests.
+- **A caller-supplied name may not be a string.** `re.findall` on an int is a 500; guard with `isinstance(name, str)`.
+- **If users didn't start the job, document it** in `girder-claude-chat/girder_claude_chat/help/troubleshooting.md`, so the assistant can answer "what is this job?" instead of guessing.
+
 ### Progress Reporting via SSE
 
 Jobs report progress through `Job().updateJob()` which emits SSE events:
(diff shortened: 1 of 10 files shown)


arjunrajlab:
Second commit addresses a self-review of the branch. Four findings fixed, plus one found while writing the tests.

**Malformed input → 400, not 500.** All three worker-request endpoints handed their request body straight to `runJobRequest`, which calls `.get("name")` on it, so a JSON-array body was an uncaught `AttributeError`. Now validated with `requireObjectBody` at the API boundary:

| Endpoint | Before | After |
|---|---|---|
| `POST /upenn_annotation/compute` | 500 | 400 `Tool must be a JSON object` |
| `POST /annotation_property/:id/compute` | 500 | 400 `Parameters must be a JSON object` |
| `POST /worker_preview/request` | 500 | 400 `Parameters must be a JSON object` |
| `POST /worker_preview/request` (unknown dataset) | 500 | 400 `No such folder: …` |

That last row is the extra find: `models/workerPreviews.py` raised `RestException(code=500)` from inside a *model* for a bad dataset id, while its twin `models/annotation.py:939` raises `ValidationException` for the identical case. Now `Folder().load(..., exc=True)` and Girder maps it (400 / 403).

**Container name.** Docker requires `[a-zA-Z0-9]` as the first character, and `_`, `-`, `.` survive the character filter — so a tool named `_temp mask` yielded a name docker rejects. Now stripped, with the all-punctuation case falling back to the request type.

**Test-harness fix worth calling out.** My first annotation-compute test *passed without the fix*: I'd used a syntactically valid but nonexistent `datasetId`, so the model's dataset lookup rejected the request before the body was ever read. These endpoints have several 400 paths, so `assertStatus(resp, 400)` alone proves nothing. The tests now assert the message and use a real folder where the handler looks one up; all four fail with the source stashed. Logged in the `nimbus-backend` skill so the next malformed-input test doesn't repeat it.

Also: `None` instead of `mock.Mock()` as `self` in the unbound-method call (a Mock would absorb any future `self.*` access instead of failing), and the patched celery task no longer shadows the name `docker`.

**Verification.** Full suite `376 passed` (368 before, +8). `flake8` clean. Against a rebuilt girder container: all four cases above confirmed by curl, a real `worker_interface/request` still returns 200 with a SUCCESS job, and driving the frontend's own `requestWorkerInterface` action end-to-end returns a 9-parameter interface with the toast reading `Pulled worker interface for annotations/connect_to_nearest:latest has completed successfully.`

I checked the new body guard can't reject a legitimate caller: the frontend (`AnnotationsAPI.ts:314`, `PropertiesAPI.ts:109`, `:263`) and the `nimbusimage` package (`annotations.py:328`, `properties.py:209`) all send object bodies.

**Not changed, by decision:** interface jobs stay registered as normal tracked jobs, so they keep raising a completion toast and the annotation-panel badge. Now that the title is legible that was judged fine.

pchoisel:
LGTM

PR #1296, opened 2026-07-31
fix(dev): make the local worker actually consume the cpu/gpu queues

arjunrajlab:
## Problem

On a local dev install, running any worker from the frontend hangs forever on the interface load step. Girder returns `200` for `POST /api/v1/worker_interface/request`, but no worker container is ever launched.

#1236 split worker dispatch across `cpu`/`gpu` Celery queues and pointed the local worker at them via docker-compose:

```yaml
  worker:
    image: girder/girder_worker
    command: -Q celery,cpu,gpu
```

That never took effect. The `girder_worker` image's `/docker-entrypoint.sh` ends in:

```bash
$BIN -m girder_worker -l info
```

with no `"$@"`, so the container command is silently discarded and the worker subscribes only to the default `celery` queue. Every dispatched job then parks in a queue nobody is listening to:

```
name     messages_ready  consumers
celery   0               1
local    0               1
cpu      3               0     <-- interface requests, nobody listening
```

`compute` on a GPU-labelled worker stalls identically once a `gpu` queue gets declared.

## Fix

Override `entrypoint` rather than `command`, so the queue list actually reaches Celery. Bypassing the entrypoint script is safe: its only other responsibility is defaulting `RABBITMQ_USER`/`PASS`/`HOST`, all of which `worker.env` already provides.

## Verification

After `docker compose up -d worker`:

```
$ docker exec worker ps -eo args | grep girder_worker
python3 -m girder_worker -l info -Q celery,cpu,gpu

$ docker exec broker rabbitmqctl list_queues name messages_ready consumers
celery   0  1
cpu      0  1
gpu      0  1
local    0  1
```

`cpu` and `gpu` went from 0 consumers to 1. The interface requests that had been parked in `cpu` dispatched immediately, ran to **exit 0**, and POSTed their interface back to Girder (`200 OK`) — reproduced against `annotations/sam_automatic_mask_generator`.

## Scope

This compose service is the only `girder_worker` definition in the repo. The `girder` container's own `-Q local` Celery worker (line 14) is untouched, and deployment configs live outside this repo, so nothing there changes.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -47,7 +47,15 @@ services:
     image: girder/girder_worker
     # The dispatcher routes jobs to the "cpu"/"gpu" queues; locally one
     # worker consumes all of them, plus the legacy default queue.
-    command: -Q celery,cpu,gpu
+    #
+    # This has to override `entrypoint`, not `command`: the girder_worker image's
+    # /docker-entrypoint.sh ends in `$BIN -m girder_worker -l info` with no "$@",
+    # so anything passed as the command is silently discarded and the worker only
+    # ever consumes the default "celery" queue -- leaving every dispatched job
+    # parked in "cpu"/"gpu" with no consumer (worker interfaces never load).
+    # Bypassing that script is safe because its only other job is defaulting
+    # RABBITMQ_USER/PASS/HOST, which worker.env already provides.
+    entrypoint: ["python3", "-m", "girder_worker", "-l", "info", "-Q", "celery,cpu,gpu"]
     links:
       - girder
     volumes:


arjunrajlab:
Closing in favour of #1297, which documents the actual cause instead.

This change fixed the symptom, not the problem. The real issue was a **stale `girder/girder_worker` image**: builds before 2025-02-17 (upstream girder_worker `e9c975f9`, "Improve Docker entrypoint") have a `/docker-entrypoint.sh` that does not forward `"$@"`. Current images do:

```bash
celery -A girder_worker.app worker -l INFO $@
```

so master's existing `command: -Q celery,cpu,gpu` works fine on any reasonably recent image. Verified on the affected install: its cached image dated from **2022-11-17**, and `docker compose pull worker && docker compose up -d worker` fixed it with no change to `docker-compose.yaml` at all. The worker now runs `celery -A girder_worker.app worker -l INFO -Q celery,cpu,gpu` with a consumer on all four queues.

What made this hard to pin down: the worker image is *pulled*, not built. The service has `image:` and no `build:`, so `docker compose build` never updates it — not even `--no-cache` — and `docker compose up` reuses the cached copy. Rebuilding from scratch cannot fix it, which sends you looking at the wrong things.

Why not merge this anyway as defence in depth: it would have left the install on a 3.5-year-old worker image while looking healthy, hiding both the staleness and three years of upstream fixes. It also pins `python3 -m girder_worker` rather than upstream's `celery -A girder_worker.app worker`; that still works today, but it is not the maintained path.

For the record, the override itself was sound — I confirmed it subscribes to all three queues on both the old and current images. It was simply aimed one level too low.

Production was never affected: any deployment newer than 2025-02-17 already honours `command:`.

PR #1299, opened 2026-08-01
Fix empty-subset CSV exports

arjunrajlab:
## Summary

- define `annotationIds` as an exact optional CSV export subset: omitted exports all, a non-empty array exports those annotations, and an empty array exports none
- preserve that distinction in the frontend request serializer and backend iterator
- document the endpoint contract and cover it at both the client and live endpoint boundaries

## Root cause

The client converted an omitted ID filter to `[]`, while the backend treated both omitted and empty lists as “export all.” Any filtered export resolving to zero annotations could therefore silently download the whole dataset—the opposite of the requested scope.

## Impact

Existing “Export all” and non-empty subset exports are unchanged. An explicitly empty filtered subset now produces a header-only CSV instead of exporting every annotation.

## Validation

- `pnpm test src/store/ExportAPI.test.ts --run` — 11 passed
- backend export tests — 32 passed
- `pnpm tsc`
- `pnpm lint:ci`
- `flake8` on the two touched Python files
- full backend suite — 359 passed; one unrelated MongoDB setup error occurred in `testUpdateIgnoresUnknownFields`, and that test passed immediately when rerun alone

### DIFF
--- a/codebaseDocumentation/EMPTY_CSV_EXPORT-REVIEW.md
+++ b/codebaseDocumentation/EMPTY_CSV_EXPORT-REVIEW.md
@@ -1,6 +1,7 @@
 # Empty-subset CSV export — review findings (PR #1299)
 
-Codex round 1 on `codex/fix-empty-csv-export` (head d634f5db), 2026-08-01.
+Codex rounds on `codex/fix-empty-csv-export`, 2026-08-01.
+Round 1: head d634f5db (F1–F3). Round 2: head c1b4a05e (F4).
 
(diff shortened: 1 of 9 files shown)


arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

pchoisel (commented):
LGTM

PR #1071, opened 2026-03-09
Add Zenodo publication integration

arjunrajlab:
## Summary
- Adds server-side Zenodo upload workflow: streams project source images, annotation JSON exports, and collection configs directly from Girder to Zenodo's bucket API
- Zenodo API tokens stored encrypted at rest (Fernet) in user metadata
- Supports full lifecycle: upload draft → review → publish (mint DOI) → new versions
- Frontend card in ProjectInfo with token management, upload progress, publish confirmation, and draft discard

## Backend
- `zenodo_client.py` — Zenodo REST API wrapper (create/upload/publish/new version)
- `zenodo.py` — Upload/publish/status/discard endpoints with background thread uploads
- `zenodo_credentials.py` — Token CRUD with Fernet encryption
- Pre-flight validation enforces Zenodo limits (50GB total, 100 files)

## Frontend
- `ZenodoPublish.vue` — Main card with status chips, progress bar, action buttons
- `ZenodoTokenDialog.vue` — Token entry with sandbox toggle
- `ZenodoAPI.ts` — Extended with publish/credential methods
- Types added to `model.ts` (`IProjectZenodo`, `TZenodoStatus`)

## Test plan
- [x] Tested credential storage/retrieval/deletion via curl
- [x] Tested initial upload to Zenodo sandbox (sandbox.zenodo.org)
- [x] Tested publish flow (mints test DOI on sandbox)
- [x] Tested new version flow (re-upload after publish creates new version)
- [x] Verified `--` filename separator works (Zenodo rejects `/` in filenames)
- [x] Verified `application/octet-stream` content type for all uploads (Zenodo rejects other types)
- [x] Frontend UI testing in browser

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1300, opened 2026-08-01
branch-review skill: cross-cutting checks from the last three multi-round reviews

arjunrajlab:
Updates the `branch-review` skill with the recurring finding shapes from the last three multi-round reviews — #1298 (Analysis panel, four Codex rounds), #1288 (timelapse panel) and #1279 (connection list). Every shape below was flagged by review at least once, and several recurred across all three PRs; none was previously in the reviewer's checklist.

## New cross-cutting section (frontend + backend)

1. **Symmetric-path twin sweep** — CLAUDE.md calls this the repo's most repeated shape, but the review skill never told the reviewer to do it. Lists the twins by concept: create-styling ↔ update-restyling, draw ↔ retain/clear, register ↔ teardown, guarded call site ↔ siblings, client ↔ server end of a contract, mode branches.
2. **"Empty" is not a signal — three-state list contracts** — absent/null vs non-empty vs present-but-empty, checked at *both* ends of the wire (`ids || []` on the client and `if ids:` on the server each independently turned "export zero matches" into "export everything" in #1298). Includes failure-vs-empty: helpers returning `{}`/`[]` on error made a transient network failure hide an entire dataset.
3. **Review-fix commits are the riskiest part of the diff** — most late-round findings on #1298 were consequences of earlier fixes, not the original feature: fix interactions ("leave ids on failure" × "drop ids only on null"), call-site guards missing siblings (the invariant belonged at the request boundary), and half-fixes (palette eviction without stacking layout). Step 1 now pulls `git log` to spot these commits.

## New frontend checks (10–14)

- **Lossy change-identity signatures** — sampled hashes (first/middle/last), undelimited nested lists, and membership-only identities that miss content changes (tag edits, server-side recomputes). Three separate rounds of #1298 hit this one shape.
- **Async refresh actions** — claim the sequence-guard token before every early return (found twice in one file); invalidate derived state on every input change including replacement and downstream chain links; unresolved state contributes no constraint; one owner per refresh.
- **String-keyed references after renames** — stale `dispatch("...")` strings are invisible to tsc and masked by watchers.
- **Hidden-but-mounted work** — `v-show`-hidden palettes/tabs doing mount-time or per-scrub work; recurred in all three PRs (#1279 scoped counts, #1288 connection scan, #1298 invisible WebGL Plotly render).
- **Overlay/palette stacking combinations** — new palettes must join the shared clearance computation; the check is combinatorial (#1288, #1298).

Issue Categories and the Checklist Coverage example gain matching rows, so findings in these shapes are routable and the coverage table shows they were checked.

## New backend check 12 (added during review)

Distilled from 25 inline comments by pchoisel across PRs #1071–#1247 (new material only — his #1203 round was already check 11): Girder built-ins over hand-rolled plumbing (`getCurrentUser()` is already loaded, `getServerMode()`), factorization habits (repeated statements → for-loop; new helpers applied to existing call sites), Python idioms (`defaultdict`, `.copy()`, dataclass rows over parallel arrays), signature hygiene (`self` on non-methods, unjustified decorators), no in-band sentinels in user-data fields, the unhandled-input questions he reliably asks (None parameters, re-import semantics, dicts in CSV cells), and why-comments on non-obvious mechanisms.

## Verification

Application test: a planted diff containing four of the shapes (guard token claimed after a bail-out, `{}` returned on fetch failure, `[]`-collapse on both ends of an export contract, sampled id signature) was reviewed by a fresh agent given the updated skill. All four were flagged, with the client and server halves of the collapse correctly paired as twins and the sampled signature escalated for its selection-clearing/bulk-action consequence — framing that comes directly from the new text.

Mirror: `.agents/skills` regenerated with `sync_skills.py --write`; `--check` (what CI runs) passes.

Two forward references are called out by PR number in the text (`cyrb53` in `signatures.ts`, `dispatchedActions.test.ts`) — they land with #1298, which is still in review.

https://claude.ai/code/session_01LE4Kf3kUrbZt3MfMqacuba


arjunrajlab:
**Re: Remove truthy-array checks from empty-list warnings** — Confirmed and fixed in ac6471e3. The check no longer lists `if (ids)` as a collapse (arrays are truthy in JS, so it preserves the null-vs-`[]` distinction), and `|| []` is now described by its actual mechanics — an absent → empty merge — rather than the reverse. The rule is reframed as "any expression that merges two of the three states, in either direction," with per-language examples: `if ids:` in Python and `ids?.length` guards / length ternaries in TS (empty → absent); `|| []` / `?? []` defaults (absent → empty). The #1298 narrative now describes the composed failure correctly: the client merged null into `[]` before the request, and the server read `[]` back as "no constraint."

Branch-wide audit for the same shape (language-semantics claims in the added guidance — truthiness, null/undefined, length checks): no other instances. One near-miss examined and left alone: the failure-vs-empty bullet's "`null`-as-data" wording is correct in context — null as a *checked* failure sentinel is the prescribed fix there; the bug is a caller coercing it into data. Both skill copies regenerated via `sync_skills.py`; `--check` passes.

@codex review

arjunrajlab:
**Re: Describe both list-state merge directions accurately** — Confirmed and fixed in 5759928c. The opening now states the consequence per direction — empty → absent turns "act on zero matches" into "act on everything"; absent → empty turns "act on everything" into "act on nothing" when the receiver honors the contract — and presents the #1298 export regression as the composition of the two, rather than attributing one consequence to every merge.

Branch-wide audit for the same shape (a consequence claimed class-wide that belongs to one direction or subcase): one other instance found and fixed — the sampled-signature bullet attributed the refetch + selection-clearing consequence pair to every consumer; it now says watchers keyed on the signature skip their work, with the dual consequence attributed to #1298 specifically. No other instances. Mirror re-synced; `--check` passes.

@codex review

arjunrajlab:
**Re: Distinguish lazy VWindow tabs from v-show palettes** — Confirmed and fixed in e2eee94d, verified against the template comment in `AnnotationBrowser.vue` that documents the lazy-mount behavior. Check 13 now describes the two lifecycles separately with their distinct cost profiles: `FloatingPalette` slots mount eagerly and cost from dataset load even if never opened (#1298 Plotly render, #1288 connection scan), while `VWindowItem` tabs mount lazily on first activation and cost from the hidden tab only after they've been opened once (#1279 scoped counts). The Hidden-Mounted Work category row was tightened to match.

Branch-wide audit for the same shape (two mechanisms with different lifecycles lumped under one claim): no other instances — check 14 and the remaining bullets are FloatingPalette-specific. Mirror re-synced; `--check` passes.

Also new since the last review, in 0907f65c: backend check 12, distilling pchoisel's recurring review comments across PRs #1071–#1247 (Girder built-ins over hand-rolled plumbing, factorization habits, Python idioms, signature hygiene, in-band sentinels, and the unhandled-input questions he reliably asks). The PR description is updated to match.

@codex review

arjunrajlab:
**Re: Distinguish a mode field from an in-band sentinel** — Confirmed and fixed in 402ab9d4, verified against `codebaseDocumentation/PR1203-PAUL-REVIEW.md` (P3, verdict: by-design / already-separate). The bullet keeps the generic in-band-sentinel warning but now names the sibling-field shape as the correct resolution — `{values: string[], exclusive: boolean}` already keeps the mode out of band, and a tag literally named "exclusive" works fine — with the doc reference, and scopes flagging to values that live inside the data itself.

Branch-wide audit for the same shape (guidance examples drawn from review questions that were answered and closed rather than confirmed findings): every other verdict recorded in PR1203-PAUL-REVIEW.md is "fix" and matches checks 11/12, and the round-1–3 examples all correspond to fix commits in their PRs. No other instances. Mirror re-synced; `--check` passes.

@codex review

PR #1302, opened 2026-08-01
Server-side analysis gating: exact gates at any dataset size

arjunrajlab:
Stacked on #1298 — merge that first.

Makes the Analysis panel work above `MAX_ANALYSIS_PLOT_POINTS` (50,000), where it previously refused to plot or gate. Spec and enforced regression checklist: `codebaseDocumentation/SERVER_GATING.md`.

## The load-bearing idea

A gate becomes a **pure per-annotation predicate**: membership depends only on the gate definition and one annotation, never on the population it is evaluated against. Chaining is redundant for *filtering* (the intersection telescopes), so it only matters for display.

Purity is what makes server-side resolution stateless — and it removes the entire class of invalidation bugs that dominated #1298's twelve review rounds. The over-cap refresh signature is O(plots), not O(population): gate definitions plus two revision counters, with no population hash and no filter state, so no filter change can invalidate a server-resolved gate.

**One deliberate behavior change:** a category absent from a gate's pinned order is now **outside** the gate, below and above the cap. The old behavior appended unknown categories in population-encounter order, so membership depended on which other categories happened to exist — arbitrary, and inexpressible as a per-annotation predicate. Unknown categories still plot (appended, sorted for determinism); to gate them, redraw.

## Phases

1. **`POST /upenn_annotation/analysis/gate_ids`** — resolves polygons to ids server-side; the client routes over-cap resolution through it. Two projected scans regardless of plot count; annotation docs anchor existence so orphaned property-value docs can't produce ids.
2. **`POST /upenn_annotation/analysis/histogram2d`** — server-binned counts render as Plotly heatmaps above the cap, with gates drawn as closed shapes/rectangles instead of lassos. A banner names any active filter the distribution can't express (ROI, hidden-layer rule, oversized id lists); the picture may over-include, never under-include, and resolution is unaffected.
3. **Gate definitions as list filter terms** — page fetches, counts, and select-all carry polygons instead of id lists that can run to hundreds of thousands of entries.

## Cross-language parity is the keystone

Gating now has two implementations that must never disagree — a dataset crossing the cap must not change gate membership. TypeScript is the reference; `test/fixtures/analysis_gating_parity.json` is generated from it and the Python suite asserts **bit-exact** jitter and identical id sets against the same file. Two traps it pins: `JSON.stringify` doesn't escape non-ASCII, and JS sorts tags by UTF-16 code unit while Python sorts by code point (diverging only on astral characters).

## Performance, measured

On the 708,983-object Xenium dataset: gate resolution **~4.5 s**, histogram **~7.3 s**, gate-carrying list page **~4.8 s**. The first cut was 16.7 s / 33.4 s; two fixes closed the gap without changing a single result — memoizing category-key encoding per axis (700K `json.dumps` calls producing a handful of distinct strings) and batch-decoding id code units via `frombuffer` (1.9 s → 0.05 s).

## Verification

Backend verified end to end at 708K scale through the real REST API (`scripts/verify_server_gating.py`, committed and repeatable): a whole-dataset gate resolves to exactly 708,983; a narrow slice through the jittered strip resolves to 227,901 and is stable across requests; the histogram badge equals the `gate_ids` answer; `list/ids` and a filtered page carrying the same definition return the identical set.

Frontend: 3,497 vitest tests green, `tsc` and `lint:ci` clean. Backend: full suite green (one flaky *fixture* error in `test_uncomputed_counts` under 17-minute full-suite load — passes alone and alongside the new tests).

**Not yet clicked in a browser.** The heatmap rendering and shape-drawing paths are covered by unit tests but need a human pass: open the Xenium dataset → Analysis, confirm heatmaps replace the "narrow the filters" notice, draw a closed shape, and check the gate chip, viewer, and Objects tab all narrow.


arjunrajlab:
All three findings were real and are fixed. Each was reproduced with a failing test before any change, and I swept the branch for other instances of the same shapes rather than patching only what was flagged.

## P1 — Cap categorical histogram cell counts

Correct, and worse than described: the explosion is reachable **without a hostile request**. `derive_axis_categories` appends categories found in the data, so a dataset where every annotation carries a distinct tag yields one column per annotation — a 700K-wide axis from ordinary data.

Added `MAX_HISTOGRAM_CELLS = 512²`, deliberately set to the budget the numeric cap already permits, so no request can allocate more than a numeric plot already could. Checked in **two** places: at the boundary for request-supplied categories, and inside `histogram2d` after deriving them, which is the case the boundary cannot see. The helper raises `ValueError`; the endpoint maps it to a 400, per the layering rule.

Held by *"testHugeCategoricalGridIsRejected"* and *"testDataDerivedCategoriesAreAlsoBounded"*.

## P1 — Avoid materializing all gate matches into one `$in`

Correct, and it exposed an overclaim in my own spec, which said "exact at any dataset size". That holds for gate *resolution* (`analysis/gate_ids` streams ids, no bound) but not for the list path. The doc now states the real limit.

Two changes:

- **A majority gate ships as `$nin` of its complement.** Inside a pipeline already scoped to the dataset these are equivalent, and the complement is strictly smaller — a gate keeping 95% of a dataset now costs 5% of the ids instead of 95%. This halves the worst case (a gate matching exactly half).
- **`MAX_GATE_CONSTRAINT_IDS = 400_000`** across all gates in one request (~8 MB, half the command limit). Past it the request fails with a comprehensible 400 instead of an opaque BSON error from MongoDB.

I did not implement the full fix, and want to be explicit about that: this raises the ceiling roughly 2× rather than removing it. Eliminating it means not materializing ids for the list at all — pushing the gate's own predicate into the query (a property axis' bounding box is a `$gte`/`$lte` range on the PV collection; a categorical axis is an `$in` on tags/channel plus a jitter sub-range) and using exact point-in-polygon only to refine. That is a larger change; it is written up in `SERVER_GATING.md` under "Size bound".

Held by *"testMajorityGateUsesComplementNotAGiantIn"*, *"testMinorityGateStillUsesIn"*, *"testComplementAndInAgreeThroughTheEndpoint"*, *"testOversizedGateConstraintIs400NotAMongoFailure"*.

## P2 — Preserve unknown-category exclusion in `gateCount`

Correct, and your worked example was exactly right — one known plus one appended category resolved to 1 id while the badge reported 2. My coordinate-reuse optimization reused coordinates built from the *derived* categories while the gate only spans its *pinned* ones.

Fixed by applying the same pinned-count mask the client's `resolveGateIds` uses (jitter is bounded by ±0.28, so rounding recovers the category index exactly) — so the fast path and the resolver now enforce the invariant identically, which is the property that matters. The test asserts `gateCount` equals what `analysis/gate_ids` returns, not just a literal.

Held by *"testGateCountExcludesAppendedCategories"*.

## Branch-wide sweep

Generalizing the two shapes turned up two more instances, both fixed here:

1. **"Lenient cap on an expensive unit, public endpoint"** (shape of finding 1): `MAX_ANALYSIS_PLOTS` was 100. Each plot rebuilds coordinates for both axes over the whole dataset (~0.7 s per categorical axis at 700K), so one request could buy ~130 s of CPU. Lowered to 20 — a sequential gating strategy is a handful of plots.

2. **"Representation changed, sibling inspector missed"**: moving gate clauses out of `idConstraints` into `gateMatchClauses` (necessary — a `$nin` is not an id list) meant `_hasAnnotationFieldFilters` no longer saw them, so a gate **+ property filter** took the PV-driven pipeline where an `_id` clause is never applied and **the gate silently stopped filtering**. This is the "one of two symmetric paths" trap `AGENTS.md` warns about. Caught because it broke two existing tests, which is the coverage working as intended.

## Verification

- Backend: **446 passed**, full suite, no failures.
- Frontend: 3,507 passed; `tsc` and `lint:ci` clean.
- Live on the 708,983-object Xenium dataset and the 52,282-object HCR dataset (both above the plot cap): heatmaps render, drawn gates resolve, chaining and downstream invalidation behave, and a JS recomputation of the jitter over all 708,983 ids matches the server's resolved set with zero mismatches.

Also filed #1303 for a pre-existing display defect this work surfaced: `jitterFromId` is affine in its seed, so the X and Y salts differ by a constant and the jitter pairs land on a lattice (60 of 100 grid cells empty). Display-only — gating stays exact — and left unfixed here because gate polygons are stored in jittered coordinates, so changing the function would move every persisted gate's membership.

@codex review

arjunrajlab:
Round 2 fixed. All three were real, and — worth naming — **all three were consequences of the round-1 fixes**, which is exactly the pattern `ANALYSIS_PANEL.md`'s retrospective describes for this feature. Each was reproduced with a failing test first.

## P1 — Check the gate-ID budget before retaining each clause

Correct, and the irony is the point: the guard I added against exhausting memory could itself exhaust memory. The budget was checked after the loop, so every gate's ObjectIds were converted and retained first — 20 gates × ~350K on a 700K dataset is ~7M `ObjectId`s held on the way to returning the 400.

Now the budget accumulates per gate and is checked **before** conversion, and partial state is dropped from `filters` on the way out. The test asserts both that resolution stops early (fewer `resolve_gate_ids` calls than gates) and that nothing oversized is left retained.

Held by *"testBudgetIsCheckedBeforeRetainingClauses"*.

## P1 — Cap categorical-axis length independently of cell count

Correct. `MAX_HISTOGRAM_CELLS` is a product, so one axis could carry 262,144 categories whenever the other collapsed to a single bin — and your point about the downstream representation is the part that actually bites: every category is returned and `axisLayout` installs each as an explicit Plotly tick, so the browser is the victim, not the server.

Added `MAX_HISTOGRAM_AXIS_CATEGORIES = 512` (matching the numeric per-axis bin cap, already far past readable), enforced **per axis** at the boundary and again after deriving categories from the data. The test uses the exact shape you described — distinct tags against a constant property, so the grid is only 40 cells and the per-axis limit is the only thing that can catch it.

Held by *"testSingleCategoricalAxisIsCappedIndependently"*.

## P2 — Enforce the 20-gate limit in the client

Correct, and this is the "one of two symmetric paths" trap again: I lowered a backend cap without its frontend counterpart. Your failure trace is right — the 400 becomes `null`, the changed-input path has already cleared resolved ids, so every gate stops filtering and no retry can succeed.

Capped on the client at both places a plot can appear: creation (`addAnalysisPlot` refuses past the limit, and "Add plot" is disabled with a reason) and hydration (a stored configuration is sliced to the cap, since it isn't bound by the creation guard). I chose capping over batching because the same limit applies to `analysisGates` in list filters — batching `gate_ids` alone would still leave the list request rejected at 21 gates.

Held by *"stops adding plots at MAX_ANALYSIS_PLOTS"*, *"allows adding again after one is removed"*, *"disables Add plot at the cap"*.

## Also in this push: progress is now visible

Unrelated to the review, from user feedback. The busy indicator was a small line of text at the top of the panel — but the palette body scrolls, so it disappeared exactly when a user scrolled down to a plot to wait for it. Worse, the **histogram** requests (the slow path above the cap, seconds at 700K) had no indicator at all.

The bar is now sticky and covers both sources. Verified live at 708K: scrolled 834 px to the bottom and the bar stayed pinned and visible, while the non-sticky intro text scrolled out of view.

## Verification

- The two suites covering every changed path: **73 + 78 pass** (`test_server_list.py`, `test_analysis_gating.py`).
- Frontend: **3,513 pass**; `tsc` and `lint:ci` clean.
- Live at 708K: gate resolution, chaining, and the busy bar all re-checked after the changes; end-to-end timings unchanged.

One caveat I'd rather state than paper over: running the two backend files *together* intermittently produces MongoDB `index build ... Operation not permitted` errors during fixture setup. It lands on a different test each run and never reproduces in isolation (both files are 100% green individually, and a full-suite run after round 1 was 446/446 clean). I read it as local test-database churn, not these changes — a full run is going now and I'll follow up if it says otherwise.

Also filed from this work: #1303 (pre-existing correlated-jitter display defect) and #1304 (gate + property filter forces the slow annotation-driven pipeline, ~13× — measured, with the fix sketched).

@codex review

arjunrajlab:
Following up on the caveat in my last comment: the full backend suite is **448 passed, 1 error**, and the error is confirmed environmental.

It landed on `test_access_control.py::TestConnectionAccessControl::testCreateMultipleConnectionsDeniedWithoutAccess` — a connections access-control test with no relationship to anything in this PR. The cause is the same each time:

```
pymongo.errors.OperationFailure: Index build failed: ...
  Collection girder_test_<hash>.token :: caused by ::
  index build on empty collection failed :: 1: Operation not permitted
```

That is `pytest-girder` creating and dropping a fresh test database per class, with MongoDB intermittently refusing an index build on an empty collection. Evidence it is not this branch:

- It has landed on **four different unrelated tests** across runs (`test_uncomputed_counts`, `test_analysis_gating`, `test_server_list`, now `test_access_control`), and on the `token` collection here — not one this PR touches.
- Every one of them passes in isolation, including this one just now (4 passed).
- An earlier full run on this same branch was **446/446 clean**, so it is not deterministic in the code.

Net: 448 passing (up from 446 — the two new tests from round 2), with one flaky fixture error that predates and is orthogonal to these changes.

arjunrajlab:
Round 3 fixed. All three were real, and all three were the **same class as rounds 1 and 2** — resource bounds on the public analysis endpoints. Three consecutive rounds of one class means I was fixing instances instead of the class, so this push also writes the class into the `nimbus-backend` skill. More on that below.

## P1 — Bound the gate-resolution response size

Correct: `MAX_GATE_CONSTRAINT_IDS` covered the list query but nothing covered the resolution *response*. Added `MAX_GATE_RESPONSE_IDS = 2,000,000` across all plots — sized so a single gate matching most of a large dataset (708K ids, ~18 MB) still works, while the 20-plot multiple cannot.

Accumulated and checked **before each plot's ids are retained**, applying the lesson from your round-2 finding: a guard that validates after building the thing it guards against has already paid the cost.

Held by *"testGateResponseIdBudgetIsEnforced"*.

## P1 — Lower the polygon-work ceiling

Correct, and I measured it rather than taking the estimate. `points_in_polygon` does one full-length numpy pass per vertex, on 708,983 points:

| vertices | time per gate |
|---|---|
| 200 | 0.22 s |
| 1,000 | 1.02 s |
| 4,000 | 4.16 s |

So the real figure for the old cap is ~10 s per gate and ~3.5 min across a full request — less than the 140-billion-operation projection (numpy is faster than that implies), but a genuine slow-loris on a public endpoint that **no DB timeout covers**, which was your actual point.

A drawn lasso emits tens to a few hundred vertices — Plotly's `drawclosedpath` produced 19–20 in my live tests — so 10,000 was 30–100× real use and bought only CPU burn. Now `MAX_GATE_VERTICES = 1,000` plus `MAX_TOTAL_GATE_VERTICES = 10,000` per request, since many modest gates add up to the same cost.

Held by *"testPerGateVertexCapIsRealistic"*, *"testTotalVertexBudgetIsEnforced"*.

## P1 — Batch or queue histogram refreshes

Correct. Each request independently runs `_analysisData` and materializes the dataset, and the panel fired one per plot on open or on any signature change that invalidated them all.

Serialized through a queue: one in flight at a time, superseded requests skip the round trip entirely (guard checked before dispatch, not only after). I chose serializing over a batch endpoint because the sticky busy bar added in the last push already covers the whole queue, so the cost is visible latency rather than a silent stall — and a shared-scan batch endpoint is a bigger protocol change worth doing on its own.

Held by *"serializes histogram requests instead of firing one per plot"*.

## The durable fix

Per `AGENTS.md` ("turn review findings into durable artifacts, not one-off fixes"), the general rule is now in `.claude/skills/nimbus-backend` (mirrored to `.agents/`, parity check passes): enumerate the dimensions that **multiply** and bound each plus their product; check budgets **as** they accumulate; bound what the **data** can produce, not only what the request asks for; keep every backend limit paired with a client counterpart; and pick thresholds from **measurement** — with the `$in`/`$nin` case as the worked example of intuition being wrong (a "whichever is smaller" rule looked obviously right and lost time near the crossover).

## Verification

- Backend: **452 passed, 0 errors** — a completely clean full run.
- Frontend: **3,514 passed**; `tsc` and `lint:ci` clean.

That clean run also resolves the flaky `index build ... Operation not permitted` errors I flagged earlier: they only ever appeared when I had **two pytest processes running against the same MongoDB**. Run alone, the suite is green. My earlier "environmental" call was right about the cause but I should have identified the trigger — it was self-inflicted.

@codex review

arjunrajlab:
Round 4 fixed. Two of the three were introduced by my own round-3 fixes, which is worth stating plainly.

## P1 — Apply the total vertex budget to every endpoint

Correct, and a clean miss on my part: I added `_requireTotalVertexBudget` to **one of three** call sites. `histogram2d` validated each upstream gate individually and `validateListInputs` did the same for list-filter gates, so 20 gates x 1,000 vertices was an accepted public request.

This is the "one of N symmetric paths" trap that `AGENTS.md` warns about — made worse by the fact that I had just written a skill section saying *keep limits paired* and then applied a new limit to a single endpoint.

The budget now runs on all three, and on histogram2d it counts the upstream gates **plus the plot's own gate**, since all of them are resolved for one response.

Held by *"testHistogramUpstreamGatesShareTheVertexBudget"*, *"testListFilterGatesShareTheVertexBudget"*.

## P1 — Invalidate queued histograms when work disappears

Correct, and this one was a regression my round-3 serialization introduced. Your diagnosis is exact: the queue captures a guard **object**, so deleting the map entry (plot removed) or skipping the prune (panel closed) left the captured token current, and the callback ran a full-dataset scan for a plot nobody is looking at. That violates the panel's core invariant — no display work behind a closed palette.

Pending work is now tracked in a set; on panel close, on dropping below the cap, and on plot removal its guard is **advanced** rather than merely forgotten. Signatures of invalidated work are cleared so it refetches when relevant again, while completed signatures survive — which is what keeps a reopen from refetching everything.

Held by *"drops queued histogram work when the panel closes"*, *"drops queued histogram work for a removed plot"*.

### A test-harness bug this uncovered, worth more than the fix

Writing that second test, it kept passing against unfixed code. The cause: `setPlots` mutated the plots array **in place** while the real `applyAnalysisPlots` replaces it. Vue short-circuits a computed whose value is unchanged by identity, so `analysisPlots` never invalidated and **every watcher downstream of it silently never re-ran**.

It hid because an existing removal test appeared to cover this — its watcher happened to read `analysisPopulation`, which returns a fresh array each evaluation, so it re-fired for an unrelated reason.

Fixed in the harness and written into `nimbus-frontend`, including the general form: an intermediate computed returning an unchanged scalar (`overCap` staying `true`) stops propagation the same way.

## P2 — Update the documented gate limits

Correct — I appended rows to the Limits table instead of replacing them, so the spec advertised both `MAX_ANALYSIS_PLOTS = 100` / `MAX_GATE_VERTICES = 10,000` and their replacements. Stale values and prose are now gone, and `MAX_TOTAL_GATE_VERTICES` is documented alongside.

## Verification

- Backend: **454 passed, 0 errors** (full suite, clean).
- Frontend: **3,516 passed**; `tsc` and `lint:ci` clean.

@codex review

arjunrajlab:
Round 5 fixed — both real, and both deeper instances of the earlier shapes rather than new ground.

## P1 — Aggregate every histogram gate under one request budget

Correct, and investigating it turned up something worse than a doubled ceiling. `analysisHistogram` calls `listIds` **directly**, without `resolveListGateConstraints` — and the match builder looks for `gateMatchClauses`, which only that function populates. So `filters.analysisGates` on this endpoint was validated and then **silently ignored**: an accepted request whose returned picture quietly drops a filter the caller asked for. (So the "both groups are resolved" reading isn't what the current code does — the second group did nothing at all, which is the more dangerous failure.)

Rather than share a budget across two channels that shouldn't both exist here, I took your alternative and **rejected filter gates on this endpoint**. Gating travels through `upstreamGates`, which share one budget with the plot's own gate, and the client never sends the other form (`analysisHistogramFilterSpec` builds filters without gate definitions). Accept-then-ignore is now a 400 with a message pointing at `upstreamGates`.

Held by *"testHistogramRejectsFilterGates"*.

## P1 — Track pending histogram generations instead of plot IDs

Correct. A `Set<plotId>` collapses generations: when a superseded request's replacement sits behind another plot's, the older callback's cleanup removed the entry the **replacement** still needed, so invalidation couldn't see it and it dispatched after the panel closed.

Now counted per generation (`Map<plotId, count>`); advancing the guard still supersedes every queued generation for that plot at once, so invalidation stays one operation.

**Worth flagging about the test:** my first version passed against the unfixed code. It closed the panel before the old generation finished, so everything was invalidated together and the bug never showed. The failure requires the old callback to *complete* first — releasing it, letting p2 dispatch, and only then closing. Rewritten to that ordering and confirmed failing before the fix. A regression test that passes for the wrong reason is worse than none, and this one nearly shipped that way.

Held by *"still invalidates a requeued plot after an older generation finishes"*.

## Verification

- Frontend: **3,517 passed**; `tsc` and `lint:ci` clean.
- Backend: 84 gating tests pass; full suite running, and I'll follow up if it disagrees.

Not requesting another round automatically — five rounds in, the last two have been fixes-to-fixes and I'd rather not spend your cycles on diminishing returns. Happy to ask again if you'd like one more pass.

arjunrajlab:
Requesting another round. Since your last review there is **new surface you have not seen**, not just fixes: the AI panel now knows about analysis gating and can create gates.

## New in `2d3e9125` — AI panel integration

**Awareness.** `get_interface_state` now reports `analysisPlots` (axes, whether a gate is drawn and enabled, resolved count). This was the more important half: gates apply with the panel closed and are restored from the collection, so a shrunken object count with no visible filter is most likely a gate — and the model previously could not see that, so it would blame the tag or property filters. Also adds `help/analysis-panel-gating.md` (auto-indexed for `read_help_topic`) and a system-prompt block on gate vs. property filter.

**Two tools**, `create_analysis_plot` and `clear_analysis_plots`, in `src/agent/executors.ts` with schemas in `agent_tools.json`. Ranges build a **rectangular** gate — a model cannot sensibly hand-author a lasso, so freehand shapes and categorical axes deliberately stay a human affair (the tool creates the plot and says to draw it). Both are in `VIEW_STATE_TOOLS` so the turn-revert affordance covers them.

Worth a careful look, since agent tools are a new way to reach this machinery:

- The tools write through the same `filterStore` actions a user's clicks do, but nothing stops the model calling them repeatedly — each `create_analysis_plot` is a full gate resolution, and above the cap that is a server round trip over the whole dataset.
- `MAX_ANALYSIS_PLOTS` is enforced in the store getter (`canAddAnalysisPlot`) and surfaced as a tool error; I have not separately rate-limited tool calls within a turn.
- The unbounded side of a one-sided range is a finite stand-in (`1e12`), not `Infinity`, so the polygon maths cannot produce NaN. Whether that interacts badly with a property whose values legitimately exceed it is a question I have not tested.

## A bug live testing caught that the unit tests could not

`await refreshAnalysis()` is **not** sufficient to read a resulting count. It claims a stale-response guard token as its first statement, so a concurrent refresh — the Viewer watches the same inputs — supersedes it and the await resolves before any commit. The tool reported all 52,282 objects still passing while the gate had in fact resolved to 0.

It now waits for that plot's ids to appear and returns `null` counts with an explanatory note if they never do. The mocked store in unit tests commits synchronously, so no test at that layer could have found it — worth knowing if you see similar await-then-read patterns elsewhere in the executors.

## Also since your last review

Round 5's two fixes (`8ac4813d`): histogram2d rejects `filters.analysisGates` (it was a second gate channel that `analysisHistogram` then silently ignored, since it calls `listIds` without `resolveListGateConstraints`), and pending histogram work is counted per generation rather than per plot.

## Verification

- Backend **455 passed**, frontend **3,527 passed**, `tsc` and `lint:ci` clean.
- Live on the 52,282-object HCR dataset: asked the panel in plain language to gate Area >= 100 against Perimeter; it called the tool and reported 22,098 objects, with the store, badge and viewer all agreeing exactly.

@codex review

arjunrajlab:
## Self-review round (7550197b)

Ran two independent adversarial reviewers over the whole branch — one on the backend + gating maths, one on the frontend components + agent tools. Each finding below was reproduced against the code before being fixed, and each new test was confirmed to fail without its fix (by temporarily restoring the pre-fix behaviour, not by reasoning).

**Security — one High.** `gateMatchClauses` was a client-writable Mongo query channel. It's internal: `resolveListGateConstraints` writes it and `_buildListMatchStages` splices its contents straight into the aggregation `$match.$and`. But `validateListInputs` allowlisted the keys it knew without removing unknown ones, and the resolver *appends* to it (`setdefault`) rather than replacing it — so a client value survived into the query on `/list`, `/list/ids` and `/analysis/histogram2d`. A string gave `{"$and": ["x"]}` → uncaught 500; a real clause ANDed an arbitrary operator into the dataset-scoped match, including a catastrophic-backtracking `$regex` worth up to `AGGREGATION_MAX_TIME_MS` of Mongo CPU per unauthenticated request. Now stripped at the boundary.

**Correctness — four Highs.**

- **Pure ids have no dependent suffix**, but `dropAnalysisGateIdsFromPlot` dropped one regardless. Above the cap the refresh identity projects only `{id, xAxis, yAxis, gate}`, so toggling `gateEnabled` — or removing/re-axing an *ungated* plot that precedes a gated one — dropped downstream ids while leaving the identity untouched. Nothing then re-requested them: one click on the enable toggle turned **both** gates off, widened the viewer to the whole dataset, and zeroed the badge, with no way back but reopening the panel.
- **Histogram invalidation broke after close → reopen → close.** The pending count belongs to the outstanding callbacks; clearing it inside the invalidator let a superseded callback's `finally` consume a *live* generation's entry, making that generation invisible to the next invalidation. It then dispatched a full-dataset scan with the panel shut.
- **One throw poisoned the histogram queue.** No catch on the chain, so it settled rejected and every later `.then` skipped its callback — all histogram fetching dead for the session, and since the in-flight count is incremented *before* the `.then`, the busy bar stayed pinned. `toDisplay` was the reachable thrower (`categories !== null` passes for `undefined`).
- **The agent's "Revert view changes" was a memory-only lie.** `hydrateAnalysisPlots` deliberately doesn't save (it runs store-catches-up-to-config); the forward path does. So reverting undid the gates on screen while the shared configuration kept them, and whether a reload or the next debounced save won was a race.

**Also fixed:** open gate bounds read the annotation store, which is empty in exactly the stub-only regime this feature exists for (every "unbounded" side collapsed to ±1,001,000 — a *smaller* silent sentinel than the 1e12 constant round 6 replaced); `categorical_raw_identity` indexed location sub-keys `locationSchema` doesn't require, a 500 on `/list` for `"location": {}`; appended categories ordered by label on the client and by key on the server, so a categorical axis silently reordered across the cap; the histogram request identity omitted the dataset id its twin includes; `clear_analysis_plots` was ungated despite destroying shared hand-drawn work that can't be reconstructed; a refused gate resolution was console-only; `derive_axis_categories` bypassed the encoder memo and checked its cap after allocating; `analysisHistogramFilterSpec` re-read frame state unconditionally, re-hashing up to 50,000 ids per Z-scrub.

**Test quality — four tests couldn't fail.** The open-bound test set up the single state in which the code worked. `describeAgentToolCall`'s tool list was hand-maintained and never got the new tools, so the test that exists for exactly this bug class couldn't see it — it's now derived from the registry. The revert test stopped at "hydrate was called". `contentRevision` was a plain constant no test could drive, on the adjacent line to a `propertyValuesRevision` that was properly wired.

`SERVER_GATING.md`'s regression checklist gained an entry per invariant (citations validated by `regressionChecklist.test.ts`), plus a note on the one display limitation this review surfaced and did *not* fix: above the cap a categorical column is binned to a zero-width cell while the predicate still places points at `index ± jitter`, so a shape drawn inside the column takes an arbitrary fraction of it with no on-screen cue.

3,544 frontend and 463 backend tests pass; `tsc`, `eslint --max-warnings=0` and `flake8` clean.

arjunrajlab:
### Live verification of the self-review fixes

Backend rebuilt (`docker compose build girder`, not restarted — the plugin is baked into the image) and the page hard-reloaded, since `src/store/filters.ts` changed.

**HCR, 52,282 objects, property axes.** Two gated plots resolved server-side in 2.3 s (`pure=true`, 34,154 and 22,508; composed set 22,508). Then the exact bug the store fix targets: toggling plot 1's enable checkbox left plot 2's ids **intact** (22,508), dropped the active-gate count to 1, and the input signature was **unchanged** — confirming nothing would have re-requested those ids, which is why the old behaviour was permanent rather than transient. Re-toggling restored both gates. Gates survived a reload and re-resolved to identical counts.

**morphology, 708,983 objects, stub-only, no computed properties.** The regime where the old open-bound code read an empty store: `annotations.length === 0`, both axes categorical. The categorical heatmap rendered end-to-end (708,983 of 708,983 binned), which is what exercises the changed `derive_axis_categories` signature and the memo. The documented jitter-strip gate resolved to **352,994 — byte-for-byte the number recorded in SERVER_GATING.md before these changes**. That is the meaningful parity result here: `categorical_raw_identity` now reads every field defensively and `derive_axis_categories` encodes through the memo, so a coercion that disagreed across the two languages would have moved that number.

Also checked live: the gate keeps filtering with the panel closed (`activeGates=1`, 352,994, `panelOpen=false`); close → reopen → close issued exactly **one** `histogram2d` POST and none after the final close; the Analysis badge and its tooltip track the gate count, with the tooltip now reading "select the objects to keep"; no error banner appeared, and the console was clean on both datasets with capture confirmed live (`[vite] connected` present, error filter empty).

One console exception did show up initially — a `vuex-module-decorators` "reading 'running'" in `AiPanel.vue`. It was timestamped 95 minutes before the reload and immediately followed `[vite] hot updated: /src/store/aiPanel.ts`, i.e. the known HMR-corrupted-store artefact from editing a store module during dev, not a fault. It did not recur on any fresh load.

Test plots were removed afterwards and a Mongo scan across every `upenn_collection` confirms no leftover `analysisPlots` in any shared configuration.

arjunrajlab:
Requesting another round (`4ffecd94`). Since your last review I ran two independent adversarial reviewers over the whole branch and fixed what they found, so **most of the new surface is fixes rather than features** — and on this branch that has reliably been where the next round's findings come from. Rounds 2 and 4 were entirely consequences of the previous round's fixes; please weight `7550197b` accordingly.

## The one you should look at first

**`gateMatchClauses` was a client-writable Mongo query channel.** It is internal — `resolveListGateConstraints` writes it and `_buildListMatchStages` splices its contents straight into the aggregation `$match.$and` — but `validateListInputs` allowlisted the keys it knew and never removed unknown ones, and the resolver *appends* to the key (`setdefault`) rather than replacing it. So a client value survived into the query on `/list`, `/list/ids` and `/analysis/histogram2d`. A string gave `{"$and": ["x"]}` → uncaught 500; a list gave an arbitrary operator ANDed into the dataset-scoped match, including a catastrophic-backtracking `$regex` worth up to `AGGREGATION_MAX_TIME_MS` of Mongo CPU per unauthenticated request. It could only narrow within the dataset, so not a cross-dataset read, but it is the worst thing found on this branch.

It is now `filters.pop("gateMatchClauses", None)` at the top of the validator. **The generalizable question I would like you to check: are there other dicts in this plugin that are both client input and an internal write target?** The trap is that `setdefault` / `.get(x) or []` append to a client value rather than replacing it, so the reader looks innocuous and the bug lives in the fact that the dict has two authors.

## Other fixes in `7550197b`, each a place a fix can go wrong

- **Deleted a check.** `derive_axis_categories` now enforces `MAX_HISTOGRAM_AXIS_CATEGORIES` while the distinct set accumulates (it also now goes through the encoder memo — it was calling `json.dumps` once per annotation, ~1s per axis at 700K). The old post-hoc per-axis loop in `histogram2d` was removed as unreachable. Worth confirming every path that produces a categorical list is still bounded, including client-supplied pinned categories.
- **New boolean that must track other state exactly.** `analysisGateIdsArePure` decides whether `dropAnalysisGateIdsFromPlot` invalidates the downstream suffix. The bug it fixes: above the cap ids are pure, but the refresh identity projects only `{id, xAxis, yAxis, gate}`, so toggling `gateEnabled` — or removing/re-axing an *ungated* plot before a gated one — dropped downstream ids while leaving the identity untouched, and nothing re-requested them. One click turned every gate off for the session. Please check the flag cannot go stale relative to the ids it describes (cap crossings both ways, dataset switch, hydration, `resetFilterState`).
- **Changed a lifecycle invariant.** `invalidatePendingHistograms` no longer deletes the pending count — it belongs to the outstanding callbacks, and deleting it let a superseded callback's `finally` consume a *live* generation's entry, so close→reopen→close dispatched a full-dataset scan with the panel shut. Please check the entry cannot now leak instead.
- **Made a sync function async.** `openGateBound` derived the "unbounded" side from `propertyStore.propertyValues`, which is empty in exactly the stub-only regime this feature exists for — every open bound collapsed to ±1,001,000, a *smaller* silent sentinel than the `1e12` you flagged in round 6. It now reads the server property histogram. `rectangularGate` became async and takes a datasetId; I added one `assertDatasetUnchanged` after it.
- **Changed category indices**, which are gate polygon coordinates. Appended categories were ordered by display label on the client and by encoded key on the server, so a categorical axis silently reordered when a dataset crossed the cap. Aligned on key order (the only ordering the server can reproduce — `channel` labels need the dataset config and `localeCompare` is locale-dependent). Pinned orders still win, so persisted gates should be unaffected, but that is worth a second opinion.
- `categorical_raw_identity` indexed location sub-keys `locationSchema` does not require, so `"location": {}` was an uncaught 500 on `/list` — every page of the Objects tab, not just the panel. Both sides now coerce missing to `null`; `TAnalysisCategoryRaw` gained `null` as a member.
- The agent's "Revert view changes" restored analysis plots **in memory only** — `hydrateAnalysisPlots` deliberately does not save, while the forward path does — so the shared configuration kept the agent's gates and a reload brought them back. New `restoreAnalysisPlots` action persists.
- `clear_analysis_plots` is now gated (it writes the shared configuration and a hand-drawn polygon cannot be reconstructed) and the system prompt no longer steers the model toward it for diagnosing a low count.
- A refused gate resolution now says why in the panel; the id budgets are reachable with a few broad gates on 700K and the retry fails identically, so console-only read as "gates silently stopped working".

## Test quality

Four tests could not fail against broken code. The sharpest: the round-6 open-bound test populated both `annotations` and `propertyValues` — the single state in which the code worked — so it passed against a production path that always took the broken branch. And `describeAgentToolCall`'s "never throws on malformed input" test iterated a hand-maintained list of tool names that never got the two new tools, so the test written for exactly that bug class could not see it; it is now derived from `AGENT_TOOL_NAMES`. If you spot other tests here that assert a mock was called rather than that an effect happened, please say so.

## Verification

- **3,544 frontend / 463 backend** tests pass; `tsc`, `lint:ci` and `flake8` clean. Every new test was confirmed to fail without its fix by restoring the pre-fix behaviour, not by reasoning.
- Live on **52,282** (property axes): two gates resolved server-side in 2.3 s; toggling one left the other's ids intact with the input signature unchanged — the evidence that the old failure was permanent rather than transient.
- Live on **708,983** (stub-only, categorical, no computed properties): the categorical heatmap rendered end-to-end, and the documented jitter-strip gate resolved to **352,994 — byte-for-byte the number recorded before these changes**, which is the real check that the encoder edits preserved TS↔Python parity.

@codex review

arjunrajlab:
## Second self-review round (`d122a195`)

Two more adversarial reviewers, briefed on deliberately different ground than the first pair: one on whether the previous round's fixes were themselves correct, one on the interaction surface nobody had examined (export, select-all, counts, configuration sharing). Every finding was reproduced before being fixed, and every new test confirmed to fail without its fix.

**The two that mattered were both cases of a fix not going far enough.**

**Pure gate invalidation is now per plot, not per request.** Last round I made `dropAnalysisGateIdsFromPlot` preserve other plots' pure ids — and the refresh threw them away anyway. `serverGateInputSignature` covers the whole request, and `resolutionPlots` widens to the *disabled* drawn gates when the palette opens, so merely opening the panel with one disabled gate present dropped every gate's ids before awaiting a re-resolution. For the seconds that takes at 700K the viewer, the Objects tab and the badge all widen to the unfiltered dataset — and a Select All in that window acts on it. A refused re-resolution made it permanent. Each plot now carries its own signature, which also means a definition edit no longer re-scans the whole dataset once per gate.

**Every scalar leaf that lands in the list `$match` is type-checked, not just the containers.** `filters.shape` and `filters.location.*` were `gateMatchClauses`'s siblings in the same validator — I generalized that finding to one key and stopped, which is exactly the "one of two symmetric paths" trap this repo's guidance names. Confirmed live *before* the fix:

| body | before | after |
|---|---|---|
| `{"shape":"polygon"}` | 200, total 52282 | 200, total 52282 |
| `{"shape":{"$ne":"polygon"}}` | **200, total 0** — operator applied | 400 |
| `{"shape":{"$nope":1}}` | **500** | 400 |
| `{"location":{"XY":{"$nope":1}}}` | **500** | 400 |
| `{"tags":{"values":[{"$ne":"x"}]}}` | operator reached `$match` | 400 |

**Also fixed:** `null` is a first-class category identity, so the decoder needed a distinct "not a key" answer — reusing `null` made `"v1:null"` fail `isEncodedAnalysisCategoryKey`, which guards a persisted gate's pinned order, so a gate drawn on an axis containing a document with no `location` was silently discarded on the next page load. CSV export measured every scope decision against `annotations.length`, which is **empty in stub-only mode**: "All annotations (0)", the Filtered radio disabled on exactly the datasets gating exists for, and picking "Selected" sent no id list, so the server exported the whole dataset with nothing on screen saying the selection had been dropped. Hydration replaying identical plots no longer clears resolved ids (the identity is built from plot content, so the signature stayed byte-identical and the watcher never fired — it was rescued only by `contentRevision` incidentally moving). The refusal banner is now cleared by a below-cap resolution and by `resetFilterState`. `currentFilters` reads frame state only when `onlyCurrentFrame` is on.

**One fix was a deletion.** I reverted last round's over-cap badge fallback: the pure count over-states, so every panel open showed a plausible wrong number for a few seconds and then silently changed it. That is worse than the "…" it replaced — a placeholder during a known wait is honest, a quietly wrong number is not.

**Six findings declined rather than implemented**, since retrospective hardening on this branch has been generating the next round's findings rather than paying off:

- duplicate pinned categories evading the 512-category cap — bounded at ~1023 ticks, no user-visible effect;
- `MAX_GATE_CATEGORIES` (10,000) vs `MAX_HISTOGRAM_AXIS_CATEGORIES` (512) disagreeing — limit reconciliation with no demonstrated failure;
- coercing all five categorical fields client-side — the two reviewers contradicted each other here, and `shape`/`channel` are schema-`required`, so the gap is unreachable;
- `reach * 1e3` overflowing to Infinity — needs a measurement above 1.8e305;
- the honesty banner over-reporting hidden-layer visibility — marginal either way;
- the `$nin` complement admitting annotations created inside a millisecond-wide window.

Happy to revisit any of those if you disagree that they're unreachable.

**Verification:** 3,552 frontend / 470 backend tests pass; `tsc`, `lint:ci`, `flake8` clean. Live: the injection payloads above, and — with a disabled gate present — the filtered count held at 34,154 across the entire palette-open window (min == max over 24 samples) where it previously spiked to the full 52,282.

arjunrajlab:
Re-requesting against the current tip — my earlier request pointed at `4ffecd94` and the branch has moved to **`d122a195`** since, with changes that are worth more of your attention than the ones I asked about.

The short version of what is new: I ran a second pair of adversarial reviewers, and **the two most serious things they found were both cases of a previous fix not going far enough.** That is the pattern to weight here.

## 1. A fix that was defeated one layer up

Round 6 made `dropAnalysisGateIdsFromPlot` preserve other plots' PURE gate ids. The refresh then discarded them anyway: `serverGateInputSignature` was one identity over the *whole request*, and `analysisRefreshScope` widens `resolutionPlots` from the enabled gates to every DRAWN gate when the palette opens. So merely opening the panel with one disabled gate present moved that signature and dropped every gate's ids before awaiting a re-resolution — for the seconds a 700K scan takes, the viewer, the Objects tab and the badge all widened to the unfiltered dataset, and a Select All in that window acted on it. A refused re-resolution made it permanent.

Each plot now carries its own signature in `analysisPureGateSignatures` (`["server", datasetId, JSON.stringify(one plot), propertyValuesRevision, contentRevision]`), and only stale plots are sent. **Please check the map cannot drift from the ids it describes** — `setAnalysisGateIds` deliberately resets it so a committer must re-state it, and my first attempt at the merge read the map back *after* that reset and silently lost the preserved entries. A test caught it, but the shape is fragile and worth a second pair of eyes.

## 2. The sibling instance I missed while generalizing the last one

You reviewed the `gateMatchClauses` fix, where an internal key reached `$match` unfiltered. I generalized that to one key and stopped — the "one of two symmetric paths" trap. `filters.shape` and `filters.location.*` sit in the same validator and land in the same `$match`, with only their *container* shape validated. Confirmed live before the fix, on `@access.public` endpoints:

| body | before | after |
|---|---|---|
| `{"shape":{"$ne":"polygon"}}` | **200, total 0** — operator applied as written | 400 |
| `{"shape":{"$nope":1}}` | **500** (OperationFailure) | 400 |
| `{"location":{"XY":{"$nope":1}}}` | **500** | 400 |
| `{"tags":{"values":[{"$ne":"x"}]}}` | operator reached `$match` | 400 |

Now every scalar leaf that lands in `$match` is type-checked to its scalar type. **The generalizable question, again: is there any other place in this plugin where a client value reaches a query without its leaf type being checked?** I have now been wrong about having swept this twice.

## 3. A bug my own previous fix created

Making `null` a category identity (so a document with `"location": {}` encodes identically on both sides) collided with the decoder, which used `null` as its "not a key" answer. So `"v1:null"` failed `isEncodedAnalysisCategoryKey` — the predicate that validates a persisted gate's pinned order — and a gate drawn on such an axis was **silently discarded on the next page load**. Now a distinct symbol sentinel. Worth checking whether any other code treats a decoded `null` as failure.

## 4. Outside the feature, but this branch makes it live

CSV export measures every scope decision against `annotationStore.annotations.length`, which is **empty in stub-only mode** — exactly the >100K regime server-side gating exists for. "All annotations (0)" was displayed, the Filtered radio was disabled, and picking "Selected" sent no id list at all, so the server exported the *whole dataset* with nothing on screen saying the selection had been dropped. Scope is now derived from the chosen scope and `annotationCount`. The preview path in that dialog is still stub-blind; I judged that display-only and left it.

## Also in this commit

Hydration replaying identical plots no longer clears resolved ids (the identity is built from plot content, so the signature stayed byte-identical and the Viewer's watcher never fired — it was rescued only by `contentRevision` incidentally moving around it). The gate-refusal banner is cleared by a below-cap resolution and by `resetFilterState`, not only by an over-cap success. `currentFilters` reads frame state only when `onlyCurrentFrame` is on — the untreated twin of the same fix in `analysisHistogramFilterSpec`. Deleted the unread `analysisPopulationSignature` getter and corrected five `SERVER_GATING.md` claims that had gone stale.

**One change is a revert.** I removed the over-cap badge fallback I added last round: the pure count over-states, so every panel open showed a plausible wrong number for a few seconds and then silently changed it — worse than the "…" it replaced.

## What I deliberately did not do

Retrospective hardening on this branch has been generating the next round's findings rather than paying off, so I declined six findings as unreachable rather than adding guards: duplicate pinned categories evading the 512-category cap (bounded at ~1023 ticks, no visible effect); `MAX_GATE_CATEGORIES` (10,000) disagreeing with `MAX_HISTOGRAM_AXIS_CATEGORIES` (512); coercing all five categorical fields client-side (`shape`/`channel` are schema-`required`, so unreachable); `reach * 1e3` overflowing above 1.8e305; the honesty banner over-reporting hidden-layer visibility; and the `$nin` complement admitting rows created inside a millisecond-wide window. **If you think any of those is actually reachable, say so and I will fix it** — but please weight correctness and reachable security over additional bounds.

## Verification

3,552 frontend / 470 backend tests pass; `tsc`, `lint:ci`, `flake8` clean. Every new test was confirmed to fail without its fix by restoring the pre-fix behaviour. Live: the injection payloads above, and with a disabled gate present the filtered count held at 34,154 across the entire palette-open window (min == max over 24 samples) where it previously spiked to the full 52,282.

@codex review

arjunrajlab:
Round 8 fixed (`a7e417e9`) — one P2, and it is the exact drift I asked you to look for, so thank you for going straight at it.

**Confirmed and fixed.** `dropAnalysisGateIdsFromPlot` is the one writer that assigns `analysisGateIds` directly instead of going through `setAnalysisGateIds`, which resets the signature map wholesale. So dropping an id there left its signature behind: removing a resolved `p2` left ids for `p1` alone but signatures for both, and the no-stale early return compares only the id maps, so nothing repaired the orphan. Both maps now move together, with a separate `signaturesChanged` flag so a signature-only prune still writes.

One correction to the report, for the record rather than as a disagreement: the consequence is a leaked string per removed plot, not a wrong answer. A plot whose ids are missing is stale *regardless* of its signature (`analysisGateIds[plot.id] === undefined` is the second half of the staleness test), so it is still re-resolved; and plot ids are uuidv4, so a later plot cannot collide with an orphan. I fixed it anyway, because `analysisPureGateSignatures` is only meaningful as a description of exactly the ids currently held, and that invariant should hold structurally rather than by argument — which is precisely why I flagged the map as fragile when I asked for this review.

The test asserts the **invariant** across removal, a gate edit and an axis change, rather than reproducing the single reported scenario:

```ts
const sameKeys = () =>
  expect(Object.keys(filters.analysisPureGateSignatures).sort())
    .toEqual(Object.keys(filters.analysisGateIds).sort());
```

Confirmed failing without the fix (`expected [ 'p1', 'p2' ] to deeply equal [ 'p1' ]`). The next writer that forgets will be a different scenario, so pinning the invariant is what actually holds.

**Still open from my last comment**, in case they were crowded out rather than cleared: the two generalizable questions. (1) Is there any other place in this plugin where a client value reaches a query without its *leaf* type checked — I have now been wrong about having swept that twice, first missing `filters.shape` / `filters.location.*` next to `gateMatchClauses`. (2) Does any other code treat a decoded `null` category as failure, now that `null` is a legal category identity? And I listed six findings from my own reviewers that I declined as unreachable; if you think any of those is actually reachable I would rather hear it than not.

3,553 frontend tests pass; `tsc`, `lint:ci`, `flake8` clean. Backend untouched by this commit, so the 470 backend tests from `d122a195` still stand.

@codex review

arjunrajlab:
Round 9 handled (`23330829`). One finding fixed, one declined as out of scope, and I owe you a correction on my own process.

## Fixed — the refusal banner outliving the refusal

Confirmed. Per-plot staleness means a refused batch leaves the *current* gates filtering normally, so once the user deletes or edits away the gate that was refused, every remaining plot is current and `stale.length === 0` is the only path left. It returned without clearing `analysisGateError`, so the banner kept claiming gates were not applying while the retained gates were visibly filtering. Cleared on that path now — "nothing unresolved" and "nothing failing" are the same statement.

Your reachability note was the useful part: I had already fixed this for the below-cap and `resetFilterState` paths, and would have called the remainder a race. The stale-only batching plus the aggregate response cap does produce it without one.

Test: *"clears the refusal banner once the refused gate is gone"* — resolve `p1`, get `p2` refused, remove `p2`, assert no refetch (p1 is still current) and no banner. Confirmed failing without the fix.

## Declined — empty-`annotationIds` CSV export, already covered by #1299

Correct finding, wrong PR. **#1299 "Fix empty-subset CSV exports"** already defines exactly this contract — omitted exports all, non-empty exports those, empty exports none — across `export.py`, `_iterAnnotations`, `ExportAPI.ts` and the CSV dialog, and does it more thoroughly than the patch I had written: it also drops stale selection ids and avoids materializing the id array just to count.

So I have **backed all CSV work out of this branch**, including the `AnnotationCSVDialog` scope rework I had landed in `d122a195`. That file is byte-identical to the base branch again, leaving #1299 a clean merge.

## Two pre-existing fixes I have deliberately KEPT here

Flagging these explicitly since they are not caused by this PR:

- the scalar-leaf type checks in `validateListInputs` — same function as this PR's own `gateMatchClauses` fix, with the test alongside it;
- the conditional frame read in `currentFilters` — this PR moved `buildListFilters` out of that file.

Both would mean merge conflicts if split, and the first would mean leaving a live operator-injection hole open while touching the surrounding lines. Say so if you would rather they were separate.

## Filed rather than fixed

Two findings from my own reviewers are now issues instead of commits, since neither is a wrong-answer bug in this feature:

- **#1305** — `MAX_GATE_CATEGORIES` (10,000) vs `MAX_HISTOGRAM_AXIS_CATEGORIES` (512): a categorical gate with >512 categories keeps filtering correctly above the cap but its heatmap silently never renders, and histogram failures are console-only.
- **#1306** — `showAnnotationsFromHiddenLayers` applies to drawing but not to `filteredAnnotations`, so the viewer and Objects tab can disagree. Which way to fix it is a product question, not a code one.

## A correction

Earlier I reported that your `4ffecd94` review "had no findings". That was wrong — it had two P2 inline comments (the `v1:null` decoder and clearing gate errors on dataset switch) and I had only read the review *body*, which was empty. Both happened to be things my own reviewers found independently and I had already fixed in `d122a195`, so nothing was lost, but I misreported it. I am reading inline comments now.

## Verification

3,554 frontend / 470 backend tests pass; `tsc`, `lint:ci`, `flake8` clean.

## Still open from earlier

The two generalizable questions, if you have anything on them: is there any other place a client value reaches a query without its *leaf* type checked (I have now been wrong about having swept that twice), and does anything else treat a decoded `null` category as failure now that `null` is a legal identity?

@codex review

arjunrajlab:
Round 10 handled (`b558ef8f`). Two fixed, one declined — and both fixes are fallout from my own earlier changes, which is now the reliable pattern on this branch.

## Fixed — reporting a plot the store refused to create

Confirmed. Making `rectangularGate` async in round 7 (so an open bound comes from the server instead of an empty store) opened the window you describe: the cap check at the top is stale by the time the plot is added, and `addAnalysisPlot` no-ops at the cap rather than throwing. The executor then applied axes and a gate to an id that did not exist, waited for it to resolve, and returned a `plotId` it had never created.

I took the second of your two suggestions rather than the first: it **confirms insertion** (`analysisPlots.some(p => p.id === plotId)`) instead of re-reading `canAddAnalysisPlot`, since re-reading is the same check-then-act one tick later. The per-turn budget is now incremented only after the plot really exists, so a failed call does not consume it.

**Your finding also exposed a harness bug that mattered more than the finding.** `addAnalysisPlot` was a bare `vi.fn()` that never appended, so `mockFilters.analysisPlots` stayed empty — meaning no test could observe whether the executor's plot had landed, which is *exactly* the state the race produces. Nine tests were passing against a mock that could not represent the bug. It now mirrors the real action, through a shared `appendAnalysisPlot` helper so the two tests that layer extra side effects on top do not silently drop the append.

## Fixed — the banner misdescribing a partial failure

Also a consequence of round 7. Per-plot resolution means a failed batch leaves already-resolved gates filtering, so "the viewer and the Objects tab show everything the other filters allow" described the pre-per-plot behaviour. Now: *"Some gates could not be resolved: … Those gates are not filtering anything for now; any gate that resolved earlier still is."*

## Declined — threading Stop through the gate refresh

Real, and I am not disputing the mechanism: `await filterStore.refreshAnalysis()` is not abortable, so Stop does not unwind until the server resolution completes.

I am leaving it because it is responsiveness rather than correctness — the user gets control back a few seconds later, with correct state — and doing it properly means plumbing an `AbortSignal` through the store action and the API layer, which is a wider change than this PR should carry at round 10. Racing the await against the signal would unwind the tool while leaving the refresh running, which trades a clear delay for a murkier one. Happy to file it as an issue if you think it is worth more than that.

## Verification

3,555 frontend tests pass; `tsc`, `lint:ci`, `flake8` clean. Backend untouched by this commit, so the 470 from `23330829` still stand. The new test was confirmed to fail without its fix.

@codex review

arjunrajlab:
Round 11 fixed (`814f7be6`). Both findings were real, and both landed on work from the two commits immediately before them — so this is the useful kind of round.

## P2 — the removal twin of the cap race

Confirmed. I fixed the cap race by confirming insertion, and stopped there. But `refreshAnalysis` and `waitForGateResolution` take seconds on a large dataset, and the user can delete the plot in that window; the wait then times out and the tool returned the removed `plotId` with a "still resolving" note — the same stale report, reached from the other end.

It now confirms the plot at **both** ends of the work and reports the removal instead of a phantom plot.

What makes this one worth calling out: it is the symmetric-path shape — *"a rule applied to one of two symmetric points"* — occurring **inside a single function**, between two awaits. My sweeps for that shape have all been cross-file greps by concept, and none of them can see this. I have added that to the note.

## P1 — a checklist item naming no test

You are right, and the miss is a little pointed. I added that invariant to the regression checklist in the *same commit* that reworded the banner, and `AnalysisPanel.test.ts` did not expose `analysisGateError` at all — so no component test could render the banner, let alone assert its wording. The repo's own rule is "an invariant without a test is a wish".

The mock now exposes it, and the test asserts the rendered text: that it says "some gates", that it says the earlier ones still filter, and that it never says "everything the other filters allow". Confirmed failing against the old wording by restoring it.

## The process note, since it is the more interesting finding

Two commits ago I added a "blast-radius sweep" to `CLAUDE.md` and the review skill: when a fix changes an invariant, check five surfaces for code that encoded the old one — including user-facing strings and doc prose, which is where most of this branch's recent findings lived. I then ran it retroactively over the branch and it correctly turned up four stale prose claims.

It did not ask whether the checklist entries *it was producing* had tests. So the sweep found the wrong banner text and fixed it, wrote a checklist line claiming the invariant was held, and left it unheld. Different rule, older rule, still applies — including to entries the sweep itself generates.

## Verification

3,557 frontend tests pass; `tsc`, `lint:ci`, `flake8` clean. Both new tests confirmed to fail without their fixes. Backend untouched since `23330829`, so the 470 there still stand.

Nothing else has changed behaviourally since your `b558ef8f` review apart from these two fixes — the commits in between are comments, docs and skills, including four stale-prose corrections the sweep found.

@codex review

arjunrajlab:
Round 12 fixed (`7abe09f1`), and this PR is now gating-only (`472535e3`).

## The P1 — confirmed, and worse than a stale picture

`histogramsByPlot` was only ever written on success, so after an axis change (or a dataset switch that recurs plot ids) the previous response stayed on screen for the seconds a whole-dataset request takes — and indefinitely if it failed, since a failure deliberately preserves the display. That rule is right for the *same* inputs and wrong for new ones.

The part that makes it a P1 rather than a cosmetic bug is the one you identified: `AnalysisScatterPlot`'s `onShapesRelayout` pins `props.histogram.xCategories` into the gate it creates. So a gate drawn during that window is **saved carrying the previous axis's category order** — a gate whose pinned coordinate space does not match its own axes, which resolves to the wrong objects with nothing on screen indicating it. That is silent wrong membership, not a stale render.

Entries whose signature changed are now cleared *before* the replacement is queued, so the plot shows no histogram rather than the wrong one, and a failed replacement cannot resurrect it. Same-input reopens still refetch nothing. Test asserts both halves and was confirmed failing without the fix.

## Scope: 462 lines removed from this PR

`CLAUDE.md` and three skills were repo-wide review guidance riding on this branch — the same scoping mistake that had CSV export work here until #1299 took it. They are now **#1307**, based on the same branch so both merge independently.

That was padding every re-review with 462 lines of prose you had to scroll past to reach the feature. This PR is now `src/`, `devops/` and `codebaseDocumentation/` only.

The `SERVER_GATING.md` process rules stay here, since they are about this feature.

## Verification

3,558 frontend tests pass; `tsc`, `lint:ci`, `flake8` clean. Backend untouched since `23330829`, so its 470 still stand.

@codex review

arjunrajlab:
Round 13 fixed (`30c88ce6`). One P2, and a good one — it is a user-facing correctness bug, not a wording preference.

Both id-budget messages told the user to **"narrow the filters"**. A gate is a pure predicate resolved over the whole dataset *before* any tag/property/frame filter applies — that is the load-bearing claim of this entire feature — so narrowing those filters cannot change the resolved id count by a single id. Anyone following that advice loops on the same 400 forever.

They now name what actually helps: redraw a gate to cover fewer objects, or disable one. The test asserts the message text, since the message *is* the defect (the 400 itself was already right), and it is cited in the regression checklist as you asked.

## Why this one is different from the last several

Rounds 10–12 were all fallout from my own fixes. **This one is not.** Both messages were wrong from the day they were written. Filter-independence was established early and I never went back to re-read the strings that contradicted it — so the blast-radius sweep would not have caught it either, because no invariant had *changed*; the code and the string simply never agreed.

That is a distinct and, for convergence purposes, more encouraging category: a pre-existing inconsistency rather than a new one I introduced.

## Correction on my last comment

I told you round 12's follow-up looked clean. It was not — I queried `/pulls/N/comments` without `--paginate`, which returns 30 per page, and this thread now has 31. The newest comment was your P2, sitting on page two. Combined with an earlier miss where I read only review *bodies* (which are identical boilerplate whether or not there are findings), that is three rounds I have misjudged as clean.

Both traps are now written into the review skill in #1307: judge a round by the paginated inline-comment count, never by the review body, and count both ways whenever a round looks clean.

## Verification

3,558 frontend / **471** backend tests pass; `tsc`, `lint:ci`, `flake8` clean. The new test was confirmed failing against the old messages.

@codex review

arjunrajlab:
Round 14 fixed (`3fcac20a`). Correct, and it is fallout from round 13's fix — I replaced "narrow the filters" on both messages with "redraw or disable" without checking that **"disable" means different things on the two paths**.

- **List path** (`MAX_GATE_CONSTRAINT_IDS`) is fed `activeAnalysisGateDefinitions`, which filters on `gateEnabled`. Unchecking a gate genuinely removes it from the query, so "disable" is accurate there and stays.
- **`gate_ids`** (`MAX_GATE_RESPONSE_IDS`) receives `analysisRefreshScope`'s `resolutionPlots`, which deliberately keeps disabled drawn gates while the panel is open so their counts can be shown. The checkbox leaves the request byte-identical — exactly the same unactionable-advice loop as round 13, one state deeper.

The response message now says **remove a plot**, which works regardless of panel state.

The test asserts the *difference* between the two messages rather than each string separately, since the defect both times was advice a path cannot honour — a test that checked each in isolation would have passed round 13's version.

**Not taken:** your first option, making `gate_ids` skip disabled gates. They are resolved on purpose — a visible disabled gate still shows its count in the panel — so that trades a wrong sentence for a missing feature.

## Where I think this stands

Findings per round are 2 → 1 → 1 → 1 → 1, severity P1+P2 → P1 → P1 → P2 → P2, and the last three have all been in error-message text rather than logic. Round 13 was a genuinely pre-existing inconsistency; this one is my own fix rebounding, which is the pattern I keep tripping on — this time within a string I had just rewritten.

Both remaining messages are now pinned by a test that encodes *why* they differ, so a future edit that flattens them back together fails.

3,558 frontend tests pass; the 173 gating and server-list backend tests pass; `flake8` clean. Full backend suite was 471 at `30c88ce6` and this commit touches only message text and its test.

@codex review

pchoisel (commented):
I didn't find anything wrong here. The code is really hard to read, but that's mainly due to the "Python yielding the same result as JS" constraint. So there's not really anything to be done.  
LGTM

PR #1333, opened 2026-08-15
Label connection-list tracks with a worker-computed track ID property

### LINKED ISSUE #1330: Connections panel: let track labels use the trackId property computed by the parent/child worker
## User feedback

> The connections panel is SO nice. One comment I'd have about it is that it would be nice for the trackIDs to be the same as they would be when calculating the "connect timelapse" metric so that identifying problematic tracks found during post-processing is easier, but this is a really great feature!

The pain point: a user finds a bad track in post-processing (e.g. in an exported CSV, `trackId = 42`) and then has no way to locate "track 42" in the Connections tab, because the panel invents its own track labels.

## Where the two IDs come from today

**The panel** (frontend, `src/utils/connections.ts` + `ConnectionList.vue`): tracks are connected components computed client-side via union-find over the scoped connection set. A track's id/label is derived from its member annotation ObjectIds — the row shows `Track #<last 6 hex chars of the smallest member id>` (see `trackKey()` and `CONNECTION_LIST.md`, "Track ids are the smallest member annotation id"). These labels never appear in any export.

**The "connect timelapse metric"**: this is actually *two* workers in ImageAnalysisProject:

1. `workers/annotations/connect_timelapse` — creates the connections themselves (tagged `"Time lapse connection"`). It produces **no track IDs at all**.
2. `workers/properties/connections/parent_child_worker` ("Parent-Child Connection IDs") with **"Add track IDs"** checked — this is what produces the `trackId` users see in post-processing. It runs union-find over all connections (same connected-components semantics as the panel) and uploads per-annotation property values: `annotationId`, `parentId`, `childId`, and `trackId` (sequential integers stored as floats), under whatever **custom property name** the user gave the property.

So the two sides compute the *same partition* (connected components of the connection graph) but name the components differently: the worker assigns sequential integers stored as annotation property values; the panel derives a hex fragment from member ObjectIds.

## Why we can't just hardcode the mapping

- The property has a user-chosen name, so the frontend can't "know" which property holds the trackId — hence the proposal below to let the user pick it.
- The worker's integer IDs are assigned by annotation fetch order and are **not stable across runs** (documented in `PARENT_CHILD.md`), so the panel can only mirror the most recently computed values, not re-derive them.
- Property values can be missing or stale (annotations/connections edited after the property last ran). The UI should show that rather than guess.

## Proposal

Add a **"Track ID property" picker** to the Connections tab (By-track view) — a small dialog/select listing the dataset's properties (ideally filtered to ones whose values look per-annotation numeric, or just list all and let the user drill into a sub-path like `Connection IDs → trackId`).

Once a property (path) is chosen:

- Each track row resolves its label by looking up the chosen property value for its member annotations (property values are keyed by annotation id in the properties store; `getPropertyValuesForIds` already exists for the lazy-values mode). Display becomes e.g. `Track 42` (optionally keeping the hex fragment as a secondary identifier).
- Since the panel's components and the worker's components are the same partition when the property is current, all members of a panel track should share one value — so any member's value works as the label.
- **If members disagree or values are missing, badge the track** ("mixed" / "no value"). That's not just an error state — it directly flags tracks whose connections changed since the property was computed, which is exactly the "problematic tracks" the user is hunting.
- Sorting/filtering the track list by the chosen ID would make "go find track 42" a direct lookup.
- Persist the chosen property per dataset (dataset view or the `connectionList` store's persisted state) so it survives reloads.

As the reporter notes, a property-to-track assignment picker is likely useful beyond this case (e.g. labeling tracks by any computed per-track quantity), so the picker shouldn't be hardwired to the name `trackId`.

## Scope notes

- Pure frontend change; property values are already fetched/fetchable client-side. No backend or worker changes required.
- The panel computes components over the *scoped* set, so a narrowed scope can split one worker track into several visible rows — both fragments would then show the same worker trackId, which is the correct reading (they're pieces of the same track). The existing member-count/time-range columns already make truncation visible.
- Related docs: `codebaseDocumentation/CONNECTION_LIST.md` (panel design, regression checklist), `ImageAnalysisProject/workers/properties/connections/parent_child_worker/PARENT_CHILD.md` (worker output).

arjunrajlab:
Closes #1330.

## What

The Connections tab's By-track view titled every track `Track #<hex fragment>`, while post-processing workflows identify tracks by the `trackId` the **Parent-Child Connection IDs** worker stores as property values — so a track flagged in an exported CSV (`trackId = 42`) couldn't be found in the panel.

This adds a **Track ID property** select to the By-track view. Pick a computed property path (e.g. `Connection IDs / trackId`) and each track row is titled with that value — `Track 42` — with the default short id moved into the tooltip.

## Resolution semantics

Per track, over all member annotation ids (`resolveTrackLabelValue` in `src/utils/connections.ts`):

| Members' values | Title | Badge |
|---|---|---|
| all share one value | the value | — |
| one value, some members without | the value | `partial` (warning) |
| differing values | short id | `mixed IDs` (warning) |
| none | short id | `no ID` |

The badges are deliberate staleness signals rather than error states: the worker assigned its ids against the connection graph at compute time, so `partial` (members added since) and `mixed` (tracks joined since) flag exactly the tracks whose connections changed after the property ran — the "problematic tracks" the reporter was hunting. A track id of `0` is a real value (the worker starts numbering there) and is never treated as missing.

## Data access

- **Wholesale mode** reads `propertyStore.propertyValues` directly — no extra requests.
- **Lazy (stub-only) mode** fetches member values with a single batched `getPropertyValuesForIds` call per path/revision, cached in the component with confirmed misses recorded — the store's value cache is viewport-pruned on every pan, so it structurally cannot hold track members. A `propertyValuesRevision` bump (recompute/import) invalidates and refetches.

## Persistence

The chosen path is stored per **configuration** in `annotationBrowserConfig.trackLabelPath` (a property id only means something within one configuration), with the same schedule-on-change / silent-hydration contract as the displayed property columns. `resolveAnnotationBrowserConfig` drops a path whose property left the configuration; a persisted path whose *values* are gone stays listed in the select so it can be seen and cleared.

## Tests & docs

- `resolveTrackLabelValue` / `formatTrackLabelValue` unit tests (including the `0`-is-a-value case).
- Component tests: title/badge rendering for all four statuses, picker items, persisted-path retention, lazy-mode single batched fetch, no fetch in wholesale mode.
- Store tests: save-scheduling vs. silent hydration, reset on dataset switch, build/resolve round trip and validation.
- `CONNECTION_LIST.md`: user-guide section plus a new *Track labels from a property* regression-checklist group, every item naming its test.

`pnpm tsc`, `pnpm lint:ci`, and the full `pnpm test` suite (3650 tests) all pass. Not verified in a live browser from this environment — worth a quick look at the By-track view with a parent/child-computed dataset before merging.

https://claude.ai/code/session_01PKUrbCSPxLMoQLpfbQu8b4

---
_Generated by [Claude Code](https://claude.ai/code/session_01PKUrbCSPxLMoQLpfbQu8b4)_


arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

PR #1340, opened 2026-08-30
Filter connections by track metrics, with opt-in object hiding and dangling cleanup

arjunrajlab:
## Summary

User request (Gia): *"your filtering for measurements and such is great and I'd love to be able to apply that to the connections, ex. Filter by track length/number of connections would be huge for me."*

The Connections tab gains a **Track filters** menu: optional min/max bounds on three **dataset-wide** track metrics — connections in track, objects in track, and duration in timepoints. A connection whose track falls outside any active bound is hidden from the list **and** from both viewer draw paths (normal-mode lines and timelapse tracks), because all three read one store predicate (`connectionListStore.connectionPassesTrackFilters`).

### Track filters
- **Metrics are dataset-wide** (keyed through `trackKeyByAnnotationId`), so a scope-narrowed fragment is judged by its full track — the same rule track colouring and duplicate-ID detection follow.
- With no filter active the predicate is a **stable always-true constant**: zero cost and zero reactive dependencies on the draw paths.
- The count readout becomes **"N of M"** while narrowing, the filter button gets a badge, and the empty state says "No connections match the track filters" instead of pretending the dataset is empty.
- Bulk delete respects the filter **by construction**: `scopedConnections` applies it, so the existing in-scope intersection picks it up; a bound change also clears the selection and resets the page (same rationale as `setScope`).
- Draw and retention apply the same predicate (`drawNewConnections` ↔ `clearOldAnnotations`), and the predicate joins both draw watch lists. A filtered-out timelapse track vanishes entirely — its members are *not* recast as orphan dots.
- Session-only view state, reset per dataset (numeric ranges don't transfer between datasets with different track scales).
- **Unknown duration** (every endpoint pointing at a deleted annotation) is excluded under an active duration bound — pure data rot, and the cleanup below is the remedy. Tracks with some surviving members get their duration from those.

### Opt-in: "Also hide these tracks' objects in the image"
Off-by-default checkbox in the same menu extending the filter to the filtered-out tracks' **objects**. Display lens only (Objects tab, exports, analysis untouched); unconnected objects are never hidden. Applied at `displayableAnnotations` — the single source every display surface derives from — and composed into `updateVisibility`'s id set, so the stub-mode budget, hydration, and HUD viewport counts stay consistent with what is drawn. While narrowing, it registers as an active constraint: the render-coverage HUD reads "1 track filter hiding whole tracks' objects" and clicking it opens the Object Browser.

### Clean up dangling connections
When connections point at deleted annotations (data rot — 2,494 of 12,459 on the verification dataset), the Connections tab shows a row offering a whole-dataset batched delete behind a confirm dialog with the live count. Stubs count as resolvable endpoints, so lazy mode never mistakes an unhydrated live annotation for a deleted one. Participates in undo. Hidden entirely on healthy datasets.

## Testing
- **48 new unit/component/store tests**, each watched failing first (TDD): pure metrics (`computeTrackMetrics`), store predicates and selection safety, viewer draw/retention/watch wiring, component counts/menu/dialog, and the active-constraint plumbing. Full suite: **3,874 green**, `tsc` and `lint:ci` clean.
- **Verified live** on a 12,459-connection / 3,796-object timelapse dataset: min-100-connections shows 2,618 of 12,459 (first metrics computation ~24 ms, cached); timelapse segments 1585 → 453 → restored on Clear; with the opt-in, 1,941 short-track objects hidden, 674 long-track members and all 1,181 unconnected objects kept; zero hidden-track ids leak into `visibleAnnotationIds`; cleanup dialog shows the live count and cancels cleanly; console clean.
- `codebaseDocumentation/CONNECTION_LIST.md` gained a user-guide section and **18 machine-checked regression-checklist invariants**, plus two new symmetric-pair rows.
- A `/branch-review` pass on the branch found 1 Medium (the visibility-refresh twin, fixed test-first) and 2 nits (type rename, raster-exemption doc note), all fixed in the final commit.

https://claude.ai/code/session_01ARvQxkodmoFFzd1Rgvq1SG


arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

PR #1341, opened 2026-08-30
Make timelapse scrubbing fast at high connection counts

arjunrajlab:
## Problem

Gia reported timelapse datasets with many connections being slow to load and interact with. Measured (foreground tab, warm-median via `scripts/nimbus-perf-probe.js`) on the real dataset (**9,965 connections**, 3,796 objects, 145 timepoints) and a synthesized Gia-scale fixture (**51,665 connections**, 45,796 objects — created and fully removed after measurement), the cost is entirely the timelapse layer rebuild:

- The connection fetch is 142 ms for ~10K connections (paginated, not the bottleneck).
- Scrubbing with timelapse OFF blocks **0 ms** at both scales.
- Scrubbing with timelapse ON blocked **157 ms/step** at 10K connections and **~1,060 ms/step** (10 fps) at 51K.

## Fixes (one per measurement)

**1. Coalesce watcher-driven rebuilds (trailing throttle).** The `displayedAnnotations` watcher fires 2-3× per frame change (two-phase visibility update); `drawAnnotations` beside it was throttled but the timelapse rebuild ran directly on every fire, bundling 2-3 full layer reconstructions into one task per scrub step. Mode toggles and selection changes keep their direct path. *10K connections: maxBlock 157 → 88 ms, fps 23.6 → 55.7.*

**2. Cut the rebuild's redundant scans.** `getDisplayedAnnotationIdsAcrossTime` called `store.layerSliceIndexes(layer)` per **annotation** (~67 ms of a ~490 ms rebuild at 45K annotations), and the orphan pass re-ran the whole scan and resolved every displayed id (~86 ms) when only unconnected ids need resolving. *51K connections: rebuild 490 → 307 ms.*

**3. Diff the timelapse layer instead of tearing it down.** Every rebuild ran `removeAllAnnotations` then reconstructed ~10-15K GeoJS features, when a scrub step changes only the mode-window edges plus time-relative styling. Features now carry `tlKey`/`tlGeom`; a pass claims matching features, updates only changed options (skip-if-unchanged, so untouched features cost nothing), and adds/removes only churn. *51K: paced rebuild 307 → ~100 ms (zero-churn pass ~80 ms).*

**4. De-quadratic the desired-set computation.** The window check now precedes the per-member spread-clone (42K clones → ~6K), resolver/getter reads are hoisted out of the hot loops, and `isStart`/`isEnd` comes from two Sets built in one pass instead of two `.some()` scans per member (O(members×connections) — 1.75M comparisons per rebuild on the fixture, quadratic on a single giant track). Within noise on the fixture median; removes the worst case.

## Results

| Scenario | Before | After |
|---|---|---|
| Time scrub @ 9,965 conns (real dataset) | 157 ms blocks, 23.6 fps | **0 ms median blocks, 88.8 fps** |
| Time scrub @ 51,665 conns (Gia-scale) | ~1,060 ms blocks, 10.2 fps | ~400 ms worst-case fast-scrub blocks, ~100 ms paced, 22-29 fps |
| Hover sweep (30 rows) | 0 ms (already restyle-in-place) | 0 ms (unchanged) |
| Timelapse OFF scrub | 0 ms | 0 ms (unchanged) |

## Tests

Each fix has a test confirmed to fail without it (`git stash`): rebuild coalescing, once-per-layer slice indexes, orphan-pass reuse, feature reuse identity, and in-place restyle of kept features. Rebuild passes are observable through the exposed `timelapseRebuildCount` (they were previously counted via `removeAllAnnotations`, which no longer runs in the mode-on path). `CONNECTION_LIST.md`'s Cost checklist gained five rows and a new symmetric pair (a new baked feature option ↔ its entry in the diff materializer).

Verified live at 51,665 connections from a fresh load: unique feature keys, no NaN coordinates, `stroke: true` on every segment, hover widens exactly one segment in the resolved GPU style, selection turns exactly the selected segment cyan and restores on deselect, mode off empties the layer, per-track colours and labels render correctly.

`pnpm tsc` / `pnpm lint:ci` / `pnpm test` (3,890) all green.

https://claude.ai/code/session_01ARvQxkodmoFFzd1Rgvq1SG


arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

arjunrajlab:
@codex review

PR #1342, opened 2026-08-31
Fix [object Object] in image switcher metadata subtitle

arjunrajlab:
## Problem

The image switcher dropdown (`LargeImageDropdown.vue`) shows each image's metadata as a subtitle, but `formatMeta` built the string with plain template interpolation (`` `${key}: ${value}` ``). Worker outputs write nested objects into item metadata — e.g. the Stitch Refinement + Illumination Correction worker writes `illumination`, `parameters`, `refinement`, and `source` as objects — so those rendered as `[object Object]`. Additionally, workers with many parameters produced metadata strings dozens of lines long, stretching the switcher's selected-image display absurdly tall.

## Fix

Added a recursive `formatMetaValue` helper: objects render as `{key: value, ...}` (keys sorted, matching the existing top-level sort), arrays as `[a, b]`, and primitives as before. Both the dropdown list items and the selected-item display go through the same formatting, so one fix covers both.

Follow-up commits:
- **Branch-review fixes:** the special-cased `tool` key (pinned first in the subtitle) also goes through `formatMetaValue`, so an object-valued `tool` can't reintroduce `[object Object]`; and the subtitle is gated on the formatted text being non-empty instead of on the truthy meta object, so images with empty `{}` meta no longer render an empty subtitle. The formatted string is computed once per image in the `formattedLargeImages` computed rather than on every slot render.
- **Length clamp + copy button:** metadata text is clamped to two lines with an ellipsis in both the dropdown items and the selected display, and a small copy button (with transient check-icon feedback) copies the full metadata string to the clipboard for anyone who needs it.

## Tests

Seven regression tests in `LargeImageDropdown.test.ts`: nested objects, arrays and deep nesting, an object-valued `tool` key, empty meta producing an empty string (hiding the subtitle), the precomputed `metaText` mapping, clipboard copy with transient feedback and its timed reset, and clipboard failure logging without feedback. Verified with `pnpm tsc`, `pnpm lint:ci`, and `pnpm vitest run src/components/LargeImageDropdown.test.ts` (15/15 passing).

https://claude.ai/code/session_012ysafiAXksH6i1Ep7BcbDt

### DIFF
--- a/src/components/LargeImageDropdown.test.ts
+++ b/src/components/LargeImageDropdown.test.ts
@@ -1,5 +1,10 @@
-import { describe, it, expect, vi } from "vitest";
+import { describe, it, expect, vi, afterEach } from "vitest";
 import { shallowMount } from "@vue/test-utils";
+
+vi.mock("@/utils/log", () => ({
+  logError: vi.fn(),
+  logWarning: vi.fn(),
+}));
 
 vi.mock("@/store", () => ({
   default: {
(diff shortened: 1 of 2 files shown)


arjunrajlab:
@codex review

